PR #4, opened 2019-05-02
Fixed incorrect optional fields generation.

kak-tus:
Tag "omitempty" was placed at incorrect position and was not used by json encoder.

### DIFF
--- a/pkg/codegen/schema_test.go
+++ b/pkg/codegen/schema_test.go
@@ -1,6 +1,10 @@
 package codegen
 
-import "testing"
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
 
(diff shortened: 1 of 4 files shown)


pkg/codegen/schema_test.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	res := []string{
+		"    1name 1type `json:\"1jname\"`",
+		"    2name 2type `json:\"2jname,omitempty\"`",
+	}
+
+	f := GenFieldsFromSchemaDescriptors(desc)
+
+	if len(f) != 2 {
```
mromaszewicz: Please use assert.EqualValues(t, f, res). You can replace your entire test logic with that one line.

kak-tus: Fixed.

PR #5, opened 2019-05-07
Fix missing ID when call function

minhnguyenmissmp:


pkg/codegen/templates/client.tmpl
```diff
@@ -43,7 +43,7 @@ func New{{$opid}}Request(server string{{genParamArgs .PathParams}}{{if .Requires
         }
         bodyReader = bytes.NewReader(buf)
     {{if not .GetBodyDefinition.Required}}}{{end}}
-        return New{{$opid}}RequestWithBody(server{{if .RequiresParamObject}}, params *{{$opid}}Params{{end}}, "application/json", bodyReader)
+        return New{{$opid}}RequestWithBody(server{{range $paramIdx, $param := .PathParams}}, {{.ParamName}}{{end}}{{if .RequiresParamObject}}, params *{{$opid}}Params{{end}}, "application/json", bodyReader)
```
deepmap-marcinr: Actually, I see the bug. Instead of doing what you've done, please replace your new code with
`{{genParamNames .PathParams}}`

genParamNames will convert the parameters names to Go-compatible names, and it's used elsewhere to accomplish what you did.

minhnguyenmissmp: I updated!

PR #7, opened 2019-05-09
feat(cmd/oapi-codegen): add flag to generate subset of code

goller:
Hey @deepmap-marcinr , I wanted to get your feedback on how I did flags first before having the Generate code use them.

What do you think?

This adds a `-generate` option.  It accepts, `all`, `server`, `client`, and `types`.

By default it uses `all`.

This is the output when an unknown option is passed:

```sh
./oapi-codegen  -generate howdy
invalid value "howdy" for flag -generate: must be "all", "server", "client", or "types"
Usage of ./oapi-codegen:
  -generate value
        The code to generate; valid options: all (default), server, client, types
  -package string
        The package name for generated code

```

### DIFF
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -24,49 +24,61 @@ import (
 	"github.com/deepmap/oapi-codegen/pkg/util"
 )
 
-// Generate is a flag for setting what code to generate.
+// Generate are possible options to the generate flag.
 type Generate string
 
 const (
-	// All generates all code and is the default.
-	All Generate = "all"
 	// Server generates the types and server implementation.
 	Server Generate = "server"
 	// Client generates the types and the client implementation.
 	Client Generate = "client"
-	// Types generates the request, responses, and parameters structs.
-	Types Generate = "types"
 )
 
-func (g Generate) String() string {
-	if g == "" {
-		return string(All)
+// Generators is a flag for setting what code to generate. Acts like
+// a unique set of options.
+type Generators map[Generate]struct{}
+
+func (g Generators) String() string {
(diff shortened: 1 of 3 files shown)


deepmap-marcinr:
Does server or client imply types? For example, if I want only the server code, I would need both the types and the server, for the client, I would need the types and the client code.

Maybe -generate can take a list, like --generate=types,client? This way, you can default it to types,server,client instead of "all"?

goller:
@deepmap-marcinr  Good call... I like the list.  I'll make that change.

Are you ok with client/server and types being split out?  Should server and client imply types?

deepmap-marcinr:
Hmm, that's an interesting question. Certainly, the Client and Server interfaces should be split out, and any helper functions. So, if there's a RequestParams object for an operation, it will be needed in both the client and server. You know, as a first attempt, just to make the code simpler, just generate all types together, independent of the client/server functions. They might be useful in code that operates on those types, so better to err on the side of too much than too little.

goller:
@deepmap-marcinr right on, ok

goller:
@deepmap-marcinr agreed... current code gets my go spidey-sense tingling... too complicated.  I'll switch it to a straight up split for sure.

goller:
@deepmap-marcinr I've updated the CLI handling and did a small refactor of the server/client generation to use the options.

The CLI parameter is now a comma delimited set of choices.  

Here is the error message with an unknown option:

```sh
./oapi-codegen -generate howdy swagger.yml
unknown generate option howdy
  -generate string
        Comma-separated list of code to generate; valid options: "client", "server"  (default client,server) (default "client,server")
  -package string
        The package name for generated code
```

deepmap-marcinr:
Thank you.

PR #13, opened 2019-05-14
feat(pkg/codegen): update for common initialisms

goller:
@deepmap-marcinr I'm not sure you'll want this PR, but, it converts some of the type names into capital letters.

The canonical list of initialisms is here: https://github.com/golang/lint/blob/8f45f776aaf18cebc8d65861cc70c33c60471952/lint.go#L771

### DIFF
--- a/pkg/codegen/operations.go
+++ b/pkg/codegen/operations.go
@@ -118,8 +118,8 @@ func GoName(name string) string {
 	}
 
 	name = ToCamelCase(name)
-	name = strings.ReplaceAll(name, "Url", "URL")
-	name = strings.ReplaceAll(name, "Cpu", "CPU")
+	name = strings.Replace(name, "Url", "URL", -1)
+	name = strings.Replace(name, "Cpu", "CPU", -1)
 	return name
 }
 
--- a/pkg/codegen/operations.go
+++ b/pkg/codegen/operations.go
@@ -111,6 +111,18 @@ func (p ParameterDefinitions) FindByName(name string) *ParameterDefinition {
 	return nil
 }
 
+// GoName converts parameter names to CamelCase and for
+func GoName(name string) string {
+	if name == "id" {
+		return "ID"
+	}
+
+	name = ToCamelCase(name)
+	name = strings.ReplaceAll(name, "Url", "URL")
+	name = strings.ReplaceAll(name, "Cpu", "CPU")
+	return name
+}
+
 // This function walks the given parameters dictionary, and generates the above
 // descriptors into a flat list. This makes it a lot easier to traverse the
 // data in the template engine.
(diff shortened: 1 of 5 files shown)


pkg/codegen/operations.go
```diff
@@ -111,6 +111,18 @@ func (p ParameterDefinitions) FindByName(name string) *ParameterDefinition {
...
 }
 
+// GoName converts parameter names to CamelCase and for
+func GoName(name string) string {
+	if name == "id" {
+		return "ID"
+	}
+
+	name = ToCamelCase(name)
+	name = strings.Replace(name, "Url", "URL", -1)
```
deepmap-marcinr: I don't know if I like this....

Go has many of these capitalizations, why handle only two? We can perhaps do the whole list, but my preference would be to simply not do it at all - and pass through the names in the oapi schema as best we can.

pkg/codegen/operations.go
```diff
@@ -304,6 +316,21 @@ func OperationDefinitions(swagger *openapi3.Swagger) ([]OperationDefinition, err
 				Spec:         op,
 			}
 
+			if opDef.OperationId == "" {
+				opDef.OperationId = strings.Title(strings.ToLower(opDef.Method))
+				for _, p := range strings.Split(opDef.Path, "/") {
```
deepmap-marcinr: I see what this is fixing. We need an operationId. However, I think it's better to error out early in code generation rather than infer what the user meant. It's an executive decision - operationId is mandatory. This lets users name each operation in a meaningful way, which we can't really infer from the path.

pkg/codegen/templates/client.tmpl
```diff
@@ -70,7 +70,7 @@ func New{{$opid}}Request{{if .HasAnyBody}}WithBody{{end}}(server string{{genPara
     }
     {{end}}
 {{end}}
-    queryUrl := fmt.Sprintf("%s{{genParamFmtString .Path}}", server{{range $paramIdx, $param := .PathParams}}, pathParam{{$paramIdx}}{{end}})
+    queryURL := fmt.Sprintf("%s{{genParamFmtString .Path}}", server{{range $paramIdx, $param := .PathParams}}, pathParam{{$paramIdx}}{{end}})
```
deepmap-marcinr: Yeah, these kinds of changes make sense, in the names we generate ourselves, it makes sense to follow Go conventions.

goller:
Ok I’ll narrow the changes to queryURL

winder:
I think this would be a nice change. I'm merging it into my fork of oapi-codegen so that the generated code will pass golint.

wolfeidau:
Anyone working on updating this PR? 

Is there something that needs adjusting to get this to progress?

Happy to help

jamietanna:
We brought this up in #722 but due to it being a breaking change on the API, we've not done it quite yet - one for #440!

PR #15, opened 2019-05-20
remove . from sentence as it triggers matcher for includes

gitu:
if you generate only server code `context.` will match for adding context include which isn't required
additionally i used the -o parameter to generate the templates


deepmap-marcinr:
Nice find, thank you. I need to make that matcher smarter.

Issue #17, opened 2019-05-24
StyleParam on objects with pointer fields not supported


oliver-smspark:
When passing an object to runtime.StyleParam that includes pointer field the following error is returned:
`error formatting 'struct_field': unsupported type *string`

This can be reproduced by adding an optional field in styleparam_test.go:
```
type TestObject struct {
	FirstName string  `json:"firstName"`
	LastName  *string `json:"lastName,omitempty"`
	Role      string  `json:"role"`
}
ln := "Johnson"
object := TestObject{
	FirstName: "Alex",
	LastName:  &ln,
	Role:      "admin",
}
```

I'm guessing the easiest fix would be to add a case in `primitiveToString` for pointers or maybe dereferencing it before going in to the switch.

deepmap-marcinr:
Yup, you're right, the fix is easy, and all it needs is reflect.Indirect() at the top level in StyleParam. I'm traveling right now, will have a look in a few days. I didn't think that things would come in by pointer, but forgot about optionals. Oops.

deepmap-marcinr:
Fixed.

oliver-smspark:
Nice thanks!

Although now it panics when passing a struct with nil pointer.
I've opened a PR with a quick fix for it: #19

PR #19, opened 2019-05-27
Handle empty optional style params

oliver-smspark:
Prevents panic when passing a struct with nil pointers to StyleParam.

Issue: #17


deepmap-marcinr:
I already pushed a fix for this before I saw this PR - I fixed it a little differently, but equivalently. Thank you, though.

PR #22, opened 2019-05-31
Use more go run

gitu:
* added templates as a tool dependency as described here: https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
* changes the travis build to generate the code and then does the git diff


gitu:
for an example on how that would look: https://travis-ci.org/gitu/oapi-codegen/builds/539646634

deepmap-marcinr:
Thank you.

PR #24, opened 2019-06-03
DeepObject support

oliver-smspark:
Adds support for deepObject parameters.
https://swagger.io/docs/specification/serialization/

### DIFF
--- a/pkg/runtime/bindparam.go
+++ b/pkg/runtime/bindparam.go
@@ -388,7 +388,34 @@ func BindQueryParameter(style string, explode bool, required bool, paramName str
 			dv.Set(reflect.ValueOf(output))
 		}
 		return nil
-	case "spaceDelimited", "pipeDelimited", "deepObject":
+	case "deepObject":
+		// Loop through all queryParams and fill the objectMap with all key/value pairs having paramName as key
+		objectMap := map[string]string{}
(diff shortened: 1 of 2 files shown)


pkg/runtime/bindparam.go
```diff
@@ -388,7 +388,34 @@ func BindQueryParameter(style string, explode bool, required bool, paramName str
 			dv.Set(reflect.ValueOf(output))
 		}
 		return nil
-	case "spaceDelimited", "pipeDelimited", "deepObject":
+	case "deepObject":
+		// Loop through all queryParams and fill the objectMap with all key/value pairs having paramName as key
+		objectMap := map[string]string{}
+		for k, v := range queryParams {
+			split := strings.Split(k, "[")
+			if len(split) != 2 {
```
deepmap-marcinr: I think that there is a little bit of a bug here - in that you can potentially split a parameter here that isn't a deepObject, and your error could unexpectedly trigger. We should only validate things known to be deepObjects.

So, I propose that you construct the string, "paramName[", or in your test case, "id[", and search for that as a prefix in the parameter name. Then, once you can safely do validation for the number of splits, etc.

deepmap-marcinr:
Thank you.

PR #28, opened 2019-06-13
Response payload names and comments

chrusty:
* Removed comments from generated types (they weren't always populated, plus they could have unsafe characters)
* Changed the name of the generated response structs (prefixed with "OAPI"), to make them less likely to clash with user-defined model names

### DIFF
--- a/examples/petstore-expanded/api/petstore-client.gen.go
+++ b/examples/petstore-expanded/api/petstore-client.gen.go
@@ -124,8 +124,8 @@ func NewClientWithResponses(server string) *ClientWithResponses {
 	}
 }
 
-// OAPIFindPetsResponse is returned by Client.FindPets()
-type OAPIFindPetsResponse struct {
+// findPetsResponse is returned by Client.FindPets()
+type findPetsResponse struct {
 	Body         []byte
 	HTTPResponse *http.Response
 	JSON200      *[]struct {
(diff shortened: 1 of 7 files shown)


examples/petstore-expanded/api/petstore-client.gen.go
```diff
@@ -124,44 +124,44 @@ func NewClientWithResponses(server string) *ClientWithResponses {
 	}
 }
 
-// FindPetsResponse is returned by Client.FindPets()
-type FindPetsResponse struct {
+// OAPIFindPetsResponse is returned by Client.FindPets()
+type OAPIFindPetsResponse struct {
```
deepmap-marcinr: Things like this is why I was avoiding generating responses :) 

I think that name collisions are a valid issue with these new changes, but I also think that 'OAPIFundPetsResponse' is kinda gross.

The user isn't going to ever be instantiating these types, how about we do away with type names altogether? Declare your functions to return `struct {whatever}`. The struct will be declared in two places - the return type, and at time of creation of the local variable into which to unmarshal.

chrusty: I agree that it is kinda gross, but the anonymous structs are not going to look pretty either. Also how would we deal with the `StatusCode()` and `Status()` methods?

What if we keep the response structs with their previous names, but make them private? That way they'll never clash with the generated schema models.

Does that work?

deepmap-marcinr:
That sounds like a good idea.

chrusty:
Alright, that looks like it works.

PR #36, opened 2019-07-17
Use existing struct for additionalProperties that ref a schema

imjoshholloway:
Right now, when a component is specified with
```
additionalProperties:
  $ref: '#/components/schemas/SomeSchema
```

The generated code produces a map with an anonymous struct: `map[string]struct` which includes all the fields the referenced schema contains. This PR switches it so the generated code uses the named struct: `map[string]SomeSchema`.


deepmap-marcinr:
Thank you for fixing this, that code is tricky.

Issue #43, opened 2019-07-29
Issue with RequestBody types in generated code


rewtraw:
Currently experiencing an issue where the {operation}JSONRequestBody types are not being used within the generated client methods.

e.g.

```
// spec.yaml

paths:
  /:
    post:
      operationId: create
      requestBody:
        content:
          application/json:
            schema:
              type: array
              items:
                $ref: '#/components/schemas/Data'
```

The generated API has the following types

```
// api.gen.go

type Data struct {
   ...
}

type createJSONBody []Data

type CreateJSONRequestBody createJSONBody
```

But, the client methods mistakenly use the `createJSONBody` type when they should use the `CreateJSONRequestBody` type.

The following is generated, and causes the client to error with `cannot refer to unexported name api.createJSONBody`

```
// api.gen.go

Create(ctx context.Context, body createJSONBody) (*http.Response, error)

...

func (c *Client) Create(ctx context.Context, body createJSONBody) (*createResponse, error) {
   ...
}
```

In my testing, manually changing this to `CreateJSONRequestBody` resolves the issue

```
// api.gen.go

type ClientInterface interface {
   Create(ctx context.Context, body CreateJSONRequestBody) (*http.Response, error)
}

...

func (c *Client) Create(ctx context.Context, body CreateJSONRequestBody) (*createResponse, error) {
   ...
}

```

This affects both the `Client` and `ClientWithResponses` methods.

PR #44, opened 2019-07-30
Make the client interface use the exported *RequestBody types

imjoshholloway:
This addresses #43. Note, this is a breaking change because the client interface changes.


deepmap-marcinr:
Sorry for the long time to respond - I was away.

We currently have two sets of client code in flight written by two different people, and there are some incompatibilities between them. I'd like to see this work without the unnecessary intermediate type definitions, so maybe there's a way to fix up the client-with-responses code. I'll have a look in a couple of days, but I'll merge this if I dont see a better way. Thank you.

Issue #25, opened 2019-06-06
Type name hints via extensions


deepmap-marcinr:
Schemas, Operations, support Swagger extensions. Rather than inferring Go type names from those properties, we can also allow specifying them. Adding support for an extension like `x-go-typename` on those would allow for more flexibility in naming the Swagger properties more clearly since they don't have to map to the same Go type.

yudppp:
@deepmap-marcinr @mgurevin
Hi marcinr and mgurevin.
I want to this feature too. 
https://github.com/deepmap/oapi-codegen/pull/212 is implements this issue.
What do you think?

B-Lorentz:
#212 solves a different issue: when you need to inject a type from elsewhere (like something you have hand-written)
What would also be useful if to be able to directly control how the generated types are named.

mromaszewicz:
this is now supported via `x-go-type-name`

PR #45, opened 2019-08-02
Custom type extensions

tamasf97:
I'd like to implement https://github.com/deepmap/oapi-codegen/issues/25. 

This PR draft only contains a small change regarding schemas.

I haven't dug into the code yet, so I don't know what edge cases I need to watch out for, e.g. right now I just assume if x-go-type is defined, then checking for additional props is not necessary.

It works for my use case, which is custom Time within JSON requests, but I've already encountered cases like `type SomeObject CustomType`, which is pointless if the custom type has imported methods defined on it.

So I'm looking for feedback to implement this feature properly, and to extend it to Operations, or wherever needed.

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -1,6 +1,7 @@
 package codegen
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


deepmap-marcinr:
This is looking good so far, thank you. In the case of an unhandled type, though, I wonder whether to use a string or a interface{}, I'm not sure what's better, honestly.

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because `x-go-type` was fully implemented in Aug 2020 via PR #212. The codebase now has comprehensive support for `x-go-type`, `x-go-type-import`, `x-go-type-skip-optional-pointer`, and `x-go-type-name` — all with tests, docs, and examples. This PR is completely superseded.

PR #48, opened 2019-08-12
Add support for external references

weberr13:
Enable external refs with -ri

External references will be encapsulated in a package as described by the
--extrefs argument.

Get can be insecure (eg ignore Cert issues) using the -insecure option

Does not generate the external code (as it would be in a different file anyway)
but allows the code to be referenced.  Automatically adds import paths/package
name to appropriate generated files.


Tobbeman:
Why was this closed?
Just curious, would like to use the feature

vulcanfan:
> Why was this closed?
> Just curious, would like to use the feature

Yes, I'd like to use it as well.  Will it be picked up again?

weberr13:
We have a solution but it is *highly opinionated*.  We can totally package it up with a pull request if people want but we have made a ton of tradeoffs that may not be good for the greater public.  Most of this is not exactly about external refs, those were relatively easy to get working and they are gated by options.  What gets interesting is that we added support for oneOf structures too.  These oneOf structures have limitations (must use discriminator mapping for best experience, not mix and match-able with AnyOf, depends on https://github.com/weberr13/go-decode).  Our source has moved quite a way and we will go to the work to put it in if the code owners want it.

weberr13:
If there is still interest I'm crafting some Pull Requests now.  I will try to tease apart the individual features we have added locally.

vulcanfan:
In fairness, what we here would see in the short term if possible would be just to resolve $ref to other local files on the same filesystem.   I believe the current status is that it can only resolve $ref to within the same file.

If something could be done about support that then that would be great.

Many thanks in advance.

dududko:
+1
It would have help a lot if this feature existed !

goshlanguage:
I am still facing this issue with `oapi-codegen`, is there any documented way of leveraging external references? 

I have something mostly working with a combination of `exclude-schemas` and `import-mapping`, but my client code doesn't import the mapping; I am likely misunderstanding how `import-mapping` is designed.

Issue #49, opened 2019-08-14
feature request: GetSwaggerUi()


pseudo-su:
I'm currently using the following code generation:
```
// go:generate oapi-codegen --package=pkg --generate spec -o ./pkg/spec.gen.go ./openapi.yaml
```

I'm curious if you'd be open to adding the ability to embed the `swagger-ui` html file into the code generated by oapi-codegen.

What would be nice for my use case is if there was a `GetSwaggerUI()` alongside `GetSwagger()`.

deepmap-marcinr:
Generating the Swagger UI requires a bunch of Java or npm tools, and I think it's out of the scope of this project. For my own API docs, for example, I simply serve up the swagger file and use the off the shelf swagger-ui docker container, pointing it at my swagger.json url.

The generated doc html is extremely large - 1 MB+ for even small specs. i'm hesitant to embed so much code into a .go file.

pseudo-su:
Fair enough, for what it's worth I did start experimenting with embedding the inlined `swagger.html` following `oapi-codegen`s code gen examples.
https://github.com/pseudo-su/oapi-ui-codegen/tree/master/tools

The biggest pain point I had was a bug (I think) in `github.com/cyberdelia/templates` when trying to generate the `templates/templates.gen.go` which I presume has something to do with encoding the html properly.

PR #50, opened 2019-08-17
feat: allow generating GetSwaggerUI()

pseudo-su:
I know you said you didn't want to do this (https://github.com/deepmap/oapi-codegen/issues/49)

But I did it anyway so I'm sharing it here in case you change your mind.

## How
* shell script can be used to pull the latest pre-compiled swagger-ui and convert it to a template
  * pulls this https://github.com/swagger-api/swagger-ui/tree/master/dist (with svn because it supports export a sub directory).
  * inlines everything into one file with `npx inliner ...` (you do have to have `npm` installed but that's all npx requires 🤷‍♂)
  * replaces the URL in the inlined html with a `{{"{{"}}.{{"}}"}}` in order to allow templating in the users desired url at runtime
* uses a forked version of the `templates` code library because it failed when  files included backticks https://github.com/cyberdelia/templates/pull/3
* adds the `specui` flag that defaults to `false`

## Where

I have a service template repo where I have an example of my `oapi-codegen` forked repo https://github.com/pseudo-su/golang-service-template


Ehco1996:
@pseudo-su i think this is a good feature for user, could you tell me why close this PR?

Issue #2, opened 2019-05-02
Ideas about new functions: chi support, request/response validation


kak-tus:
I have some ideas to new functions and I am ready to code them.

1. Add command line option to produce chi compatible server interface and handlers.
2. Add support for full request validation.
a) add middleware for chi router with params validation and secure validation;
b) add secure validation to echo middleware.

For security validation there is needed for "AuthenticationFunc". I prefer to add it not as callback function, but as interface function.

3. Add support for full response validation. There is two options.

a) validation may be added as on runtime basis as middleware: get generated response and validate it with openapi3filter. It is simple, but adds time to request processing;

b) static validation. We can generate boilerplate, so interface functions will be returns predefined genereated structs from boilerplate, these struct will be self-validated and produce from them json responses. It is more complicated, but will be more user-friendly to developer and gets better perfomance.

Please, say your opinion about this features.

mromaszewicz:
Yes, I'm planning on response validation in the same way as request validation, it will be very similar. I think it's important for unit tests to validate responses, but in production, we won't.

kak-tus:
About response validation - I mean something like this (in petstore example)
```
type ServerInterface interface {
	//  (GET /pets)
	FindPets(ctx echo.Context, params FindPetsParams) (Pet, error)
...

...
// Wrapper for FindPets
func (w *ServerInterfaceWrapper) FindPets(ctx echo.Context) error {
...
	// Invoke the callback with all the unmarshalled arguments
	res, err = w.Handler.FindPets(ctx, params)
        if err != nil {
	  return err
        }

        return json.NewEncoder(ctx.Response().Writer).Encode(res)
}
```

This will get static compile time validation and needed performance with easy usage.

mromaszewicz:
I'm hesitant to do something like that, because responses may not be JSON. If you look at the petstore example in the code, we do produce the JSON response body, and you just call ctx.JSON() to do what you propose. However, you can have a response body which is a string or a binary payload, and an operation typically has multiple response body types (normal body, error, etc). I don't want to bake the assumption that the response is a JSON response into the handler code, that's up to the application logic.

kak-tus:
Of cause, json encoder in handler code is just an example.
Handler can form response body by content type (media type) of schema.

application/json can produce FindPets(ctx echo.Context, params FindPetsParams) (Pet, error)
text/html can produce FindPets(ctx echo.Context, params FindPetsParams) (string, error)
[...]

PR #51, opened 2019-08-19
Generate chi based http handler

dannymidnight:
This is a first pass at adding [chi](https://github.com/go-chi/chi) support following on from https://github.com/deepmap/oapi-codegen/issues/2 to generate an alternative to the echo based routing. There's still a bunch of missing pieces but I wanted to push it up early for general feels. 

### Why?
The current http handler solution is built around `echo` and requires the calling application to bring along this dependency. `Chi `is "100% compatible with net/http" making it self-contained and simple to integrate with existing go applications whilst also leaning heavily into the core `context` API making it more "go-like".

### Example
Here's an example of how you could go about registering your operations. You'll note that there's no peer dependency on `chi` as `ChiHandler` returns a `http.Handler`.

Implementation is definitely a little verbose having raw request, response and passing around `Context` but I think the http.Handler interface will be more flexible in the long run and should be familiar for most go developers.

```go
type Operations struct {}

func (*Operations) FindPets(w http.ResponseWriter, r *http.Request) {
  params := petstore.ParamsForFindPets(r.Context())
  ...
}

func (*Operations) DeletePet(w http.ResponseWriter, r *http.Request) {
  params := petstore.ParamsForFindPets(r.Context())
  ...
}

func (*Operations) AddPet(w http.ResponseWriter, r *http.Request) {
  ...
}

func (*Operations) FindPetById(w http.ResponseWriter, r *http.Request) {
  ...
}

func Handler() {
  myApi := Operations{}
  petstore.ChiHandler(myApi)
}

```

### Questions
* Is this something you're open to incorporating?! 😁 
* Rename `ChiHandler` to `Handler`? Chi is really an implementation detail that doesn't need to be exposed externally.

### TODO:
* [x] Separate out templates from existing echo based ones.
* [x] Add new flag for generating chi http handler
* [X] Add example implementation
* [x] Add some test coverage!


deepmap-marcinr:
I've been thinking about this, and it's hard to do it cleanly, but your approach looks pretty nice. 

Right now, we have a fair bit of code that's router agonstic (all the type generation), and what's echo specific is the server interface (the client interface is agnostic). So, I would refactor the functionality so that the entirety of Echo bindings are in, say, echo-serverinterface.tmpl and echo-wrappers.tmpl, and you could produce chi-specific templates. You would select the backend via a command line option, which would substitute in chi-whatever.tmpl in place of those two files.

I initially chose Echo because it's quick, and it simplifies a lot of stuff, and it works for us, but I'm not particularly attached it to one way or another.

deepmap-marcinr:
One practical thing that I've found is that it's good to have an interface to implement in your generated code - for mocking, or whatever. So your operations{} struct might want to implement an interface.

I'm totally open to incorporating this, if we can constrain the chi-specific code to a few .tmpl files, just as we've done with Echo.

dannymidnight:
Great. I'll try clean things up a bit and let you know when it's ready for review.

dannymidnight:
> One practical thing that I've found is that it's good to have an interface to implement in your generated code - for mocking, or whatever. So your operations{} struct might want to implement an interface.

Yep makes sense, in my petstore example the `operations` struct is implementing [this interface](https://github.com/deepmap/oapi-codegen/pull/51/files#diff-0965434705f04f2955ae8c6b4b7928dfR32-R41).

dannymidnight:
Hey @deepmap-marcinr I think this is now ready for review. 

Major changes since last time:

* There's now a `chi-server` flag for generating the new implementation and I've kept the echo codepath using `server` for backwards compatibility.
* I've added an example implementation and test runner following a similar approach to the `echo` based tests.
* I've split the petstore examples into the following structure with the client split out from the server implementations:
```
- petstore
-- chi
-- echo
-- petstore-client.gen.go
```

Let me know what you think! Any and all feedback welcome.

examples/petstore-expanded/chi/petstore_test.go
```diff
@@ -0,0 +1,156 @@
...
+	return rr
+}
+
+func TestPetStore(t *testing.T) {
+	var err error
+
+	store := api.NewPetStore()
+	h := api.Handler(store)
+
+	t.Run("Add pet", func(t *testing.T) {
```
deepmap-marcinr: Hmm, that's an interesting way to structure tests. I've not seen this approach before.

dannymidnight:
Cheers @deepmap-marcinr!

ashwinrs:
Thank you for this @dannymidnight

PR #53, opened 2019-08-26
add basic xml support

gitu:
Heya @deepmap-marcinr 

I would like to use the generated json fields to serialize to xml what would be needed to get something like that merged. What is in this PR would be sufficient for my purposes.

Cheers


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because XML unmarshalling for request/response bodies already exists, and the parameterized struct tags system now lets users configure XML tags via config. The original PR against 2019-era code is far too outdated to merge — any XML work would need to be done fresh. There is a tracked issue (#486) noting this gap.

PR #56, opened 2019-09-04
Include path parameters in param types

dannymidnight:
This change generates param types for all operations with request parameters. By doing this it makes for a cleaner calling interface for pulling out path parameters in chi-context based handler functions.

Here's a contrived example but should hopefully help illustrate.

Before
```go
var pets map[int]Pet

func UpdatePet(w http.ResponseWriter, r *http.Request) {
  id := r.Context().Value("id").(int)
  params := ParamsForUpdatePet(r.Context())

  pets[id].Name = params.Name
}
```

Now:

```go
var pets map[int]Pet

func UpdatePet(w http.ResponseWriter, r *http.Request) {
  params := ParamsForUpdatePet(r.Context())

  pets[params.Id].Name = params.Name
}
```

### DIFF
--- a/examples/petstore-expanded/chi/api/petstore.gen.go
+++ b/examples/petstore-expanded/chi/api/petstore.gen.go
@@ -79,6 +79,9 @@ func FindPetsCtx(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 
+		// Parameter object where we will unmarshal all parameters from the context
+		var params FindPetsParams
+
 		var err error
 
 		// ------------- Optional query parameter "tags" -------------
@@ -123,6 +126,9 @@ func DeletePetCtx(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 
+		// Parameter object where we will unmarshal all parameters from the context
+		var params DeletePetParams
+
 		// ------------- Path parameter "id" -------------
 		var pathErr error
 		var id int64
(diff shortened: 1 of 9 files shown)


deepmap-marcinr:
I see what this is doing, but we're now hitting an asymmetry between Echo and Chi.

These extra params objects are completely unused in the Echo case, since path parameters are included in the function signature in the server interface. I don't think we should create unused code for Echo.

Since you have the AllParams list available in your templates for Chi, I think you should generate the Params objects for path parameters only in the Chi code.

dannymidnight:
Hey @deepmap-marcinr apologies for the slow reply – I've been out on leave for a little while!

> These extra params objects are completely unused in the Echo case, since path parameters are included in the function signature in the server interface. I don't think we should create unused code for Echo.

I can see the argument for not introducing code that's unused by Echo but I'd argue when generating parameter types via the `types` cli flag I found the omission of OpenAPI path parameters was unexpected.

I'm definitely coming at this change to improve the router agnostic (non-echo) server interface but I believe this change cleans up the default generated types to allowing for potentially hand-rolled clients based on these param structs.

> Since you have the AllParams list available in your templates for Chi, I think you should generate the Params objects for path parameters only in the Chi code.

Yeah I could do this but I think going down this path would lead to confusion having **very** similar parameters generated from both the `types` and `chi-server` code generation flags. 

Thoughts on reconsidering?

deepmap-marcinr:
How about doing API symmetry the other way - crate some adapters in the Chi case which pass in mandatory path arguments as function arguments, and only create the params structure if there are optional query arguments?

dannymidnight:
> How about doing API symmetry the other way - crate some adapters in the Chi case which pass in mandatory path arguments as function arguments, and only create the params structure if there are optional query arguments?

Yeah cool that's not a bad idea. I might close this PR for the time being and come back around on it.

I'm going to put up another PR shortly that has some bug fixes which this change kind of resolved on the side 😁

PR #57, opened 2019-09-09
Tidy template helper methods

chrusty:
@deepmap-marcinr Here is what I've got so far. There is still some more stuff I'd like to improve, and some tests I'd like to run, but it should give you an idea of where I'm going with it.

### DIFF
--- a/pkg/codegen/operations.go
+++ b/pkg/codegen/operations.go
@@ -294,6 +294,7 @@ func (o *OperationDefinition) GetResponseTypeDefinitions() ([]TypeDefinition, er
 
 					td := TypeDefinition{
 						TypeName: typeName,
+						JsonName: responseName,
 						Schema:   responseSchema,
 					}
 					if contentType.Schema.Ref != "" {
--- a/pkg/codegen/operations.go
+++ b/pkg/codegen/operations.go
@@ -293,9 +293,9 @@ func (o *OperationDefinition) GetResponseTypeDefinitions() ([]TypeDefinition, er
 					}
 
 					td := TypeDefinition{
-						TypeName: typeName,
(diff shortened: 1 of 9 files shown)


pkg/codegen/operations.go
```diff
@@ -294,6 +294,7 @@ func (o *OperationDefinition) GetResponseTypeDefinitions() ([]TypeDefinition, er
 
 					td := TypeDefinition{
 						TypeName: typeName,
+						JsonName: responseName,
```
chrusty: Using this to convey the responseName. Not sure if it is better to add a new attribute to pollute `TypeDefinition`.

pkg/codegen/template_helpers.go
```diff
@@ -130,26 +136,9 @@ func genResponseUnmarshal(operationID string, responses openapi3.Responses) stri
 		// If we made it this far then we need to handle unmarshaling for each content-type:
 		sortedContentKeys := SortedContentKeys(responseRef.Value.Content)
 		for _, contentTypeName := range sortedContentKeys {
-			contentType, ok := responseRef.Value.Content[contentTypeName]
```
chrusty: Managed to remove quite a bit of cruft

pkg/codegen/operations.go
deepmap-marcinr: I wouldn't overload the meaning of a field. Add a ResponseName and say what its purpose is. I've found that it's better to be overly verbose in this code, it's damn complex, and this kind of field separation helps inevitable future refactoring.

chrusty: Agreed. I've added a new attribute.

PR #65, opened 2019-10-03
Bug fixes for chi handler

dannymidnight:
This fixes a couple bugs in the generated `chi` http handler middleware. 

* Multiple path parameters were duplicating error variables.
* Having both query and path parameters defined caused the path parameter to get dropped.
* Multi-line resource summaries were breaking generated comments.


pkg/codegen/template_helpers.go
```diff
@@ -241,6 +241,10 @@ func toStringArray(sarr []string) string {
 	return `[]string{"` + strings.Join(sarr, `","`) + `"}`
 }
 
+func toSingleLine(s string) string {
+	return strings.ReplaceAll(s, "\n", "")
```
ashwinrs: The build is failing because go is not able to find the `strings` package

ashwinrs: Probably because travis is using [go 1.11 ](https://travis-ci.com/deepmap/oapi-codegen/builds/130332379#L195) but replaceAll was introduced in [go 1.12](https://github.com/Codewars/codewars.com/issues/1861#issuecomment-507417558)

dannymidnight:
Hey @deepmap-marcinr any feedback or issues with this PR?

pkg/codegen/template_helpers.go
dannymidnight: Yeah good spot. I've switched over to using the go 1.11 compatible `strings.Replacer`.

Issue #54, opened 2019-08-27
Missing Enum Values and validation code


mayurpate:
I am trying to generate Golang code for following openapi 3 spec. I have doubts regarding Go code generation for enum related types. 

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  description: A sample API that uses a petstore as an example to demonstrate features in the OpenAPI 3.0 specification
  termsOfService: http://swagger.io/terms/
  contact:
    name: Swagger API Team
    email: apiteam@swagger.io
    url: http://swagger.io
  license:
    name: Apache 2.0
    url: https://www.apache.org/licenses/LICENSE-2.0.html
servers:
  - url: http://petstore.swagger.io/api
paths:
  /pets:
    post:
      summary: Post a PetType Information.
      description: Post a PetType Information.
      operationId: postPets
      requestBody:
        content:
          application/json:
            schema:
              oneOf:
                - $ref: '#/components/schemas/Cat'
                - $ref: '#/components/schemas/Dog'
      responses:
        '200':
          description: Updated
    get:
      summary: Returns a pet info by given petType.
      description: Get Pet Infrmation based on pet type.
      operationId: getPets
      parameters:
        - $ref: '#/components/parameters/PetType'
        - in: query
          name: petColor
          schema:
            type: string
      responses:
        '200':
          description: Get Pet Type 
          content:
            application/json: 
              schema:
                oneOf:
                  - $ref: '#/components/schemas/Cat'
                  - $ref: '#/components/schemas/Dog'
components:
  schemas:
    Pet:
      type: object
      properties:
        color:
          $ref: '#/components/schemas/Color'
        name:
          type: string
          nullable: true
        owners:
          type: array
          nullable: false
          items:
            type: object
            properties:
              ownerName:
                type: string
              ownerAddress:
                type: string
                nullable: true
    Dog:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          required:
            - bark
          properties:
            bark:
              type: boolean
            breed:  
              type: string
              enum: [Dingo, Husky, Retriever, Shepherd]
              nullable: true
    Cat:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          properties:
            hunts:
              type: boolean
              nullable: true
            age:
              type: number
              format: float
              nullable: true
    Color:
      type: string
      nullable: true
      enum: 
        - black
        - white
        - red
        - green
        - blue
[...]

PR #68, opened 2019-10-11
generate const if schema type is string and integer with enum

nakatamixi:
Hi.

Thanks for your awesome package.
I want to generate const if I use enums in `#/components/schemas`
So, I modify schema.go to have `openapi3.ExtensionProps` in codegen.Schema,
and have EnumValues method for generate.

But I wonder my implamentation according codegen design policy.
Please Review this PR.

Thank you.

ref: https://github.com/deepmap/oapi-codegen/issues/54

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -1,7 +1,9 @@
 package codegen
 
 import (
+	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/getkin/kin-openapi/openapi3"
@@ -19,6 +21,14 @@ type Schema struct {
 	AdditionalTypes          []TypeDefinition // We may need to generate auxiliary helper types, stored here
 
 	SkipOptionalPointer bool // Some types don't need a * in front when they're optional
+	Extensions          Extensions
+}
+
+type Extensions map[string]json.RawMessage
+
+type Enum struct {
+	Name  string
+	Value string
 }
 
 func (s Schema) IsRef() bool {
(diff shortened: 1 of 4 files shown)


pkg/codegen/schema.go
```diff
@@ -52,6 +62,52 @@ func (s Schema) GetAdditionalTypeDefs() []TypeDefinition {
...
 }
 
+func (s Schema) EnumValues() []Enum {
+	if s.GoType != "string" && !strings.HasPrefix(s.GoType, "int") {
+		return nil
+	}
+	if s.Extensions == nil {
+		return nil
+	}
+	if _, ok := s.Extensions["enum"]; !ok {
```
deepmap-marcinr: all extensions must be of the form `x-something`, or openapi validation of the spec will fail.

pkg/codegen/schema.go
```diff
@@ -52,6 +62,52 @@ func (s Schema) GetAdditionalTypeDefs() []TypeDefinition {
 	return result
 }
 
+func (s Schema) EnumValues() []Enum {
+	if s.GoType != "string" && !strings.HasPrefix(s.GoType, "int") {
```
deepmap-marcinr: We should allow enums for all types. IMO. No sense limiting to strings and ints.

nakatamixi:
> My general philosophy in this package is to either support something well and fully, or not at all :) If enums work in one place, but not in another, I think people will be confused and will file bugs.

Thanks for your kind review. 
I understand.

So, I close this Pull Request until I or someone will change to cover all enums behavior.

PR #78, opened 2019-10-18
Fix #77 - Remove byte-order-marks from Go-Code

### LINKED ISSUE #77: Remove hidden byte-order-marks (BOM) in generated code
Given some weird schemas with weird descriptions (i.e. a description, which gets autogenerated and filled via a CMS), one might end up with BOM (byte-order-marks) inside the `[]byte` buffer.

The Go-Compiler is not happy with BOM. Therefore the generated Go-Code has to be cleaned up.

This could be done using awk:

```bash
 awk '{ gsub(/\xef\xbb\xbf/,""); print }' inputFile > outputFile
```

A better way, would be to add this cleanup right before one writes the generated Go-Code to disk.

weitzj:
This removes any byte-order-marks from the generated
Go-code. The BOMs might have gotting in there from
a schema and are preserved. This is fine as long
as you work with the schema.

Since we generate Go-Code from parts of the schema,
we have to take BOMs into consideration.

### DIFF
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -229,7 +229,8 @@ func Generate(swagger *openapi3.Swagger, packageName string, opts Options) (stri
 		return "", errors.Wrap(err, "error flushing output buffer")
 	}
 
-	goCode := buf.String()
+	// remove any byte-order-marks which break Go-Code
+	goCode := SanitizeCode(buf.String())
 
 	// The generation code produces unindented horrors. Use the Go formatter
 	// to make it all pretty.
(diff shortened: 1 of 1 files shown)


pkg/codegen/codegen.go
```diff
@@ -228,7 +228,8 @@ func Generate(swagger *openapi3.Swagger, packageName string, opts Options) (stri
 		return "", errors.Wrap(err, "error flushing output buffer")
 	}
 
-	goCode := buf.String()
+	// remove any byte-order-marks which break Go-Code
+	goCode := strings.Replace(buf.String(), "\uFEFF", "", -1)
```
deepmap-marcinr: huh, weird. Could you make a function, something like "sanitizeCode" or something which will be the future place to put additional such filters? Thanks.

weitzj:
@deepmap-marcinr Could you please review again. I have rebased the branch against the current master and have introduced the sanitizeCode function. I have made it public.

PR #80, opened 2019-10-20
Fix #79 - Distinguish disable/inherit of securityScheme on path

### LINKED ISSUE #79: Bug: Cannot distinguish between SecurityDefinitions override per path vs. inheriting  global SecurityDefinitions
# Bug

`Type OperationDefinition` assumes a slice `SecurityDefinitions []SecurityDefinition` as part of the operations, which are defined on a `path`.

See:
https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/operations.go#L205

# Explanation

This is problematic according to the Specification: https://swagger.io/docs/specification/authentication/

> ...
> Step 2. Applying security
After you have defined the security schemes in the securitySchemes section, you can apply them to the whole API or individual operations by adding the security section on the root level or operation level, respectively. When used on the root level, security applies the specified security schemes globally to all API operations, unless overridden on the operation level. In the following example, the API calls can be authenticated using either an API key or OAuth 2. The ApiKeyAuth and OAuth2 names refer to the schemes previously defined in securitySchemes. 

# How to fix
Therfore to distinguish, whether a `path` wants to disable a globally defined securityScheme vs. inheriting the globally defined securityScheme, one has to use a pointer to a slice:

```
// use a pointer * to distinguish disable vs. inherit
`SecurityDefinitions *[]SecurityDefinition`
```

weitzj:
This fix ensures it is possible to distinguish whether
a path disables a globally defined securityScheme vs.
that it should actually inherit the globally defined
securityScheme.

See Step 2: Applying Security of the Openapi3 Spec:

https://swagger.io/docs/specification/authentication/

See:

- #79

### DIFF
--- a/pkg/codegen/operations.go
+++ b/pkg/codegen/operations.go
@@ -435,8 +435,18 @@ func OperationDefinitions(swagger *openapi3.Swagger) ([]OperationDefinition, err
 				TypeDefinitions: typeDefinitions,
 			}
 
+			// check for overrides of SecurityDefinitions.
+			// See: "Step 2. Applying security:" from the spec:
+			// https://swagger.io/docs/specification/authentication/
 			if op.Security != nil {
 				opDef.SecurityDefinitions = DescribeSecurityDefinition(*op.Security)
+			} else {
+				// use global securityDefinitions
+				// globalSecurityDefinitions contains the top-level securityDefinitions.
+				// They are the default securityPermissions which are injected into each
+				// path, except for the case where a path explicitly overrides them.
+				opDef.SecurityDefinitions = DescribeSecurityDefinition(swagger.Security)
+
 			}
 
 			if op.RequestBody != nil {


pkg/codegen/templates/wrappers.tmpl
```diff
@@ -25,9 +25,11 @@ func (w *ServerInterfaceWrapper) {{.OperationId}} (ctx echo.Context) error {
 {{end}}
 {{end}}
 
+{{if .SecurityDefinitions}}
```
deepmap-marcinr: I don't see how this changes anything.

Previously, if security definitions was nil or empty, we would range over nothing, producing no code. This seems like a no-op

deepmap-marcinr: Oh, I see. I read the related issue.

Could we be much more explicit about this? Rater than inferring global vs local security schemes, can we have a flag instead? This way, you can switch on the value of that flag in the template code - it's easier to read and clearer.

weitzj: I guess it would be clearer if we decide beforehand in the Go-Code calling the template, which `SecurityDefinitions` "win", since we have everything in our hands. This way the template can stay dumb enough and either attached its own definitions or gets a copy of the global definitions.

deepmap-marcinr: That sounds good.

weitzj:
@deepmap-marcinr  Could you please review again the second commit. It ensures that we either inherit the global security definitions or override with out own.

pkg/codegen/operations.go
```diff
@@ -197,18 +197,23 @@ func DescribeSecurityDefinition(securityRequirements openapi3.SecurityRequiremen
...
+	PathParams      []ParameterDefinition // Parameters in the path, eg, /path/:param
+	HeaderParams    []ParameterDefinition // Parameters in HTTP headers
+	QueryParams     []ParameterDefinition // Parameters in the query, /path?param
+	CookieParams    []ParameterDefinition // Parameters in cookies
+	TypeDefinitions []TypeDefinition      // These are all the types we need to define for this operation
+
+	// These are the operation-specific SecurityDefinitions.
+	// A non-nil slice implies an explicit override of any global SecurityDefinitions.
+	// See: "Step 2. Applying security:" from the spec: https://swagger.io/docs/specification/authentication/
+	SecurityDefinitions *[]SecurityDefinition
```
deepmap-marcinr: Do we need this to be a pointer to a slice at all? It doesn't seem like it to me. You either make it local, global and nil and it should work without the change in the template.

weitzj: True. This can be reverted

weitzj:
@deepmap-marcinr  Please review again.

PR #82, opened 2019-10-20
Feature/81 client security scheme runtime

### LINKED ISSUE #81: Feature - Make it easier for the client to use securitySchemes
OpenAPI3 provides various means on howto to interact with an API, i.e. BasicAuth, Bearer, ApiKey, Oauth2. https://swagger.io/docs/specification/authentication/

It would be nice, if the `Client` can be easily configured. Currently this is possible by injection a customized `HTTPClient` and/or using the `RequestEditorFn`.

It would be nice if the runtime provided some basic guidance on how to do this.

weitzj:
Fix #81 Add NewClient() constructor with Interceptors

This adds a new client constructor and adds some
fields to the client to extend its usage to have
multiple request-interceptors, which can be used
to help use the client for some securitySchemes.

Please look at the separate commits.

## What this does

It adds some default securityproviders, which are optional to use. It also adds `Interceptors` to a client

## Using SecurityProviders

If you generate client-code, you can use some default-provided security providers
which help you to use the various OpenAPI 3 Authentication mechanism.

```
	// Example BasicAuth
	// See: https://swagger.io/docs/specification/authentication/basic-authentication/
	basicAuthProvider, basicAuthProviderErr := runtime.NewSecurityProviderBasicAuth("MY_USER", "MY_PASS")
	if basicAuthProviderErr != nil {
		panic(basicAuthProviderErr)
	}

	// Example BearerToken
	// See: https://swagger.io/docs/specification/authentication/bearer-authentication/
	bearerTokenProvider, bearerTokenProviderErr := runtime.NewSecurityProviderBearerToken("MY_TOKEN")
	if bearerTokenProviderErr != nil {
		panic(bearerTokenProviderErr)
	}

	// Example ApiKey provider
	// See: https://swagger.io/docs/specification/authentication/api-keys/
	apiKeyProvider, apiKeyProviderErr := runtime.NewSecurityProviderApiKey("query", "myApiKeyParam", "MY_API_KEY")
	if apiKeyProviderErr != nil {
		panic(apiKeyProviderErr)
	}

	// Example providing your own provider using an anonymous function wrapping in the
    // InterceptoFn adapter. The behaviour between the InterceptorFn and the Interceptor interface
    // are the same as http.HandlerFunc and http.Handler.
	customProvider := InterceptorFn(func(req *http.Request, ctx context.Context) error {
		// Just log the request header, nothing else.
		log.Println(req.Header)
		return nil
	})

	// Exhaustive list of some defaults you can use to initialize a Client.
	// If you need to override the underlying httpClient, you can use the option
	//
	// WithHTTPClient(httpClient *http.Client)
	//
	client, clientErr := NewClient(context.Background(), []Option{
		WithBaseURL("https://api.deepmap.com"),
		WithUserAgent("MY_USER_AGENT"),
		WithMaxIdleConnections(10),
		WithIdleTimeout(10 * time.Second),
		WithRequestTimeout(1 * time.Second),
		// You can register multiple providers.
		// They will be called in insertion order.
		// If a provider fails, the call-chain stops
		// an does not continue the request.
		WithInterceptors([]Interceptor{
			basicAuthProvider,
			bearerTokenProvider,
			apiKeyProvider,
			customProvider,
		}...,
		),
	}...,
	)
 ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,68 @@ There are some caveats to using this code.
  for anything other than trivial objects, they can marshal to arbitrary JSON
  structures. When you send them as cookie (`in: cookie`) arguments, we will
  URL encode them, since JSON delimiters aren't allowed in cookies.
- 
+
+## Using SecurityProviders
+
+If you generate client-code, you can use some default-provided security providers
+which help you to use the various OpenAPI 3 Authentication mechanism.
+
+
+```
+    import (
+        "github.com/deepmap/oapi-codegen/pkg/securityprovider"
+    )
+
+    func CreateSampleProviders() error {
+        // Example BasicAuth
+        // See: https://swagger.io/docs/specification/authentication/basic-authentication/
+        basicAuthProvider, basicAuthProviderErr := securityprovider.NewSecurityProviderBasicAuth("MY_USER", "MY_PASS")
+        if basicAuthProviderErr != nil {
+            panic(basicAuthProviderErr)
+        }
+
+        // Example BearerToken
+        // See: https://swagger.io/docs/specification/authentication/bearer-authentication/
+        bearerTokenProvider, bearerTokenProviderErr := securityprovider.NewSecurityProviderBearerToken("MY_TOKEN")
+        if bearerTokenProviderErr != nil {
+            panic(bearerTokenProviderErr)
+        }
+
+        // Example ApiKey provider
+        // See: https://swagger.io/docs/specification/authentication/api-keys/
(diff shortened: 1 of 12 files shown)


weitzj:
Makes sense to me. Would you still put the package in this repo here? (But not in runtime but somewhere else?)

Can we keep the fluent builder?

weitzj:
@deepmap-marcinr  Could you please review again. What I have done:

- Move securityproviders to a securityproviders package
- Only kepy the `NewClient` fluent builder
- Remove the `interceptor`
- Updated the README.

Thanks, 
Jan

PR #85, opened 2019-10-24
improvement: Use interface for the HTTP Client

sebnow:
Depend on an interface, rather than a concrete implementation, so as to
allow for alternate implementations to be used, such as decorated
clients.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,3 +14,5 @@ require (
 	golang.org/x/sys v0.0.0-20190712062909-fae7ac547cb7 // indirect
 	golang.org/x/tools v0.0.0-20190724185037-8aa4eac1a7c1 // indirect
 )
+
+go 1.13
--- a/pkg/codegen/templates/client.tmpl
+++ b/pkg/codegen/templates/client.tmpl
@@ -1,6 +1,11 @@
 // RequestEditorFn  is the function signature for the RequestEditor callback function
 type RequestEditorFn func(req *http.Request, ctx context.Context) error
 
+// Doer performs HTTP requests
+type Doer interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
 // Client which conforms to the OpenAPI3 specification for this service.
 type Client struct {
 	// The endpoint of the server conforming to this interface, with scheme,
@@ -8,7 +13,7 @@ type Client struct {
 	Server string
 
 	// HTTP client with any customized settings, such as certificate chains.
-	Client *http.Client
+	Client Doer
 
 	// A callback for modifying requests which are generated before sending over
 	// the network.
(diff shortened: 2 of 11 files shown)


pkg/codegen/templates/client.tmpl
```diff
@@ -1,14 +1,19 @@
...
+
 // Client which conforms to the OpenAPI3 specification for this service.
 type Client struct {
 	// The endpoint of the server conforming to this interface, with scheme,
 	// https://api.deepmap.com for example.
 	Server string
 
 	// HTTP client with any customized settings, such as certificate chains.
-	Client *http.Client
+	Client Doer
```
weitzj: Could you please fix the code-comment to match the Doer interface.

pkg/codegen/templates/client.tmpl
```diff
@@ -1,14 +1,19 @@
 // RequestEditorFn  is the function signature for the RequestEditor callback function
 type RequestEditorFn func(req *http.Request, ctx context.Context) error
 
+// Doer performs HTTP requests
```
weitzj: I guess it would be nice to have an additional example comment and possibly an example where one mentions that `http.Client` implements this interface

weitzj:
For this to work, you would need to remove https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/templates/client-with-responses.tmpl#L49-L89

This would be fine with me. I would suggest to at least add a ClientOption like so:

```go
// WithDoer allows overriding the default httpClient, which is
// automatically created. The Doer interface is typically implemented
// by http.Client. 
func WithDoer(doer Doer) ClientOption {
	return func(c *Client) error {
		c.Client = doer
		return nil
	}
}
```

@deepmap-marcinr wdyt?

sebnow:
> For this to work, you would need to remove https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/templates/client-with-responses.tmpl#L49-L89

Oh, right. I was working off an earlier version (current release) and didn't notice that was added. It appears that the options would still be compatible, albeit maybe a bit confusing, considering the options are only used when a client is not provided.

weitzj:
Sorry for the late reply. The other options would fail on you either way. Therefore it would be better to remove them.

weitzj:
Looks good to me 👍

I don’t have permissions to merge. @deepmap-marcinr  can help

go.mod
```diff
@@ -14,3 +14,5 @@ require (
 	golang.org/x/sys v0.0.0-20190712062909-fae7ac547cb7 // indirect
 	golang.org/x/tools v0.0.0-20190724185037-8aa4eac1a7c1 // indirect
 )
+
+go 1.13
```
deepmap-marcinr: did you use anything here which requires go 1.13? I'd like to avoid shutting out users of earlier versions of Go - since there's a compile time dependency on our runtime package.

sebnow: Oh my bad. I must have committed it by mistake - `go mod` is quite intrusive. No reason to add a requirement on 1.13.

PR #86, opened 2019-10-29
Switch to HTTP client interface to allow wrapped clients

nitishm:
The standard library doesn't support features like Retry of error and different backoff strategies for the same. The recommended approach is to wrap the standard library HTTP client package with any custom handling logic. (See ianlancetaylor's reponse in https://github.com/golang/go/issues/16047)
I propose adding an HTTPClient interface with the same method signatures as the standard library HTTP Client 
```
    type Client
        func (c *Client) CloseIdleConnections()
        func (c *Client) Do(req *Request) (*Response, error)
        func (c *Client) Get(url string) (resp *Response, err error)
        func (c *Client) Head(url string) (resp *Response, err error)
        func (c *Client) Post(url, contentType string, body io.Reader) (resp *Response, err error)
        func (c *Client) PostForm(url string, data url.Values) (resp *Response, err error)
```

### DIFF
--- a/examples/petstore-expanded/petstore-client.gen.go
+++ b/examples/petstore-expanded/petstore-client.gen.go
@@ -58,14 +58,12 @@ type AddPetJSONRequestBody addPetJSONBody
 // The interface matches all methods attached to the standard library HTTP Client
 // https://golang.org/pkg/net/http/#Client
 // type Client
-//     func (c *Client) CloseIdleConnections()
 //     func (c *Client) Do(req *Request) (*Response, error)
 //     func (c *Client) Get(url string) (resp *Response, err error)
 //     func (c *Client) Head(url string) (resp *Response, err error)
 //     func (c *Client) Post(url, contentType string, body io.Reader) (resp *Response, err error)
 //     func (c *Client) PostForm(url string, data url.Values) (resp *Response, err error)
 type HTTPClient interface {
-	CloseIdleConnections()
 	Do(req *http.Request) (*http.Response, error)
 	Get(url string) (*http.Response, error)
 	Head(url string) (*http.Response, error)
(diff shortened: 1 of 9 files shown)


weitzj:
@nitishm Just asking: since the generated client here accepts an HTTPClient, the retry mechanisms or any other advanced stuff like oauth2 should already be possible. What would make sense is provide maybe an extension to the Readme how one can create some cool HTTPclients and inject them in the generates client here. But maybe I am misunderstanding something.

nitishm:
> @nitishm Just asking: since the generated client here accepts an HTTPClient, the retry mechanisms or any other advanced stuff like oauth2 should already be possible. What would make sense is provide maybe an extension to the Readme how one can create some cool HTTP clientsΩ and inject them in the generates client here. But maybe I am misunderstanding something.

The standard HTTP client is pretty good but there are cases where you might want to wrap it with some extra functionality (think decorator pattern). 
Let's say we want the client to automatically log before and after a POST command. The wrapper struct would allow us to add the functionality by implementing the interface, but under the hood continue to use the standard go HTTP library
For instance, the wrapping struct would look like below,
```
type Wrapper struct {
    logger logrus.Logrus
    client *http.Client
}
```
In my case, I am looking to add some retry options to the HTTP client using different backoff strategies. Unfortunately, the standard HTTP client library does not provide retry capabilities, and due to the generated code using oapi-codegen client, I cannot modify it to include retry capability.

@weitzj Hope that makes sense. Let me know if you think there is a better alternative.

StevenACoffman:
Hey, this looks handy for adding retries with exponential backoff. How does this stand?

tooolbox:
Seems like this was already done in #85 and can be closed.

Issue #90, opened 2019-11-11
Support "date" string format during marshaling


jmcnevin:
Version: 1.3.0

I have noticed that if I have a spec that includes a date field:

```yaml
birthday:
  type: string
  format: date
```

My generated clients will have a *time.Time field created in the parameters struct.  This works fine during unmarshaling, but a marshaled time.Time will not be in the date format `2017-07-21`, but the datetime format `2017-07-21T00:00:00Z`.  This may throw off servers with strict input validation.

jmcnevin:
Fixed by #92

PR #92, opened 2019-11-12
Added custom types for un/marshaling

jmcnevin:
This adds special handling for "date" and "byte" string formats in OpenAPI 3.  When marshaled, "date" strings should have their time truncated, and "byte" strings should be Base64 encoded.

This is to address https://github.com/deepmap/oapi-codegen/issues/90.

Needs a bit more testing, but I'd appreciate any comments.


jmcnevin:
After some research, it appears that []byte is JSON marshaled as base64 by default, so that custom type may not even be necessary.

pkg/codegen/codegen.go
```diff
@@ -39,6 +39,21 @@ type Options struct {
 	SkipFmt            bool // Whether to skip go fmt on the generated code
 }
 
+// imports contains list of imported libraries as alias/fullname tuples, and implements sort.Interface interface
```
deepmap-marcinr: Would you mind making an `Import` struct, with Alias and Name strings instead of [2]string? It could work exactly the same, but IMO, named objects are much nicer than positional tuples.

jmcnevin:
All set.  This backed out some of my other changes.

pkg/codegen/codegen.go
```diff
@@ -121,79 +148,18 @@ func Generate(swagger *openapi3.Swagger, packageName string, opts Options) (stri
...
-		}
-		if strings.Contains(str, "errors.") {
-			imports = append(imports, [2]string{"", "github.com/pkg/errors"})
-		}
-		if strings.Contains(str, "chi.") {
-			imports = append(imports, [2]string{"", "github.com/go-chi/chi"})
-		}
-		if strings.Contains(str, "openapi_types.") {
-			imports = append(imports, [2]string{"openapi_types", "github.com/deepmap/oapi-codegen/pkg/types"})
+		for _, goImport := range allGoImports {
```
deepmap-marcinr: nice improvement, thanks!

Issue #91, opened 2019-11-11
Clean up double-slashes in URLs


jmcnevin:
Version: 1.3.0

If I have a document that defines absolute path URLs like `/api/pets` and then use the `WithBaseURL` option with my generated client, like so: 

```
_, _ = lib.NewClient(context.TODO(), lib.WithBaseURL("http://localhost"))
```

The URLs that will be hit for my client requests will contain a double-slash, like `http://localhost//api/pets`.  Not the end of the world, but some servers may fail to understand this request.

PR #93, opened 2019-11-12
Safer URL/Query generation

jmcnevin:
In attempting to fix https://github.com/deepmap/oapi-codegen/issues/91, I have modified the client templates to lean on the "net/url" and "path" packages to create URLs and query strings that are as close to correct as possible.

Could use more robust testing.


deepmap-marcinr:
Thank you.

PR #96, opened 2019-11-30
Add basePath support, and implementation for chi-server

danf0rth:


deepmap-marcinr:
Thanks for this, however, it's not quite complete. I've been too busy to implement this, because there are several cases that must be handled, and that's why it's still missing.

There is both a global server list, and a per-operation server list. Each may have more than one server. Each server definition may contain a partial path (https://deepmap.com/v2). An operation for path "op" needs to be registered as "/v2/op" for this server.

It's also not clear how to handle multiple servers properly, we can't just ignore it and pick the first. Perhaps we register all possible handlers in the server? In the client, we would have to choose which server to use, so this is a pretty invasive change.

If you make a chi server from a spec file with two servers using your CL, then swap their order, you'll break server behavior - we can't have that.

danf0rth:
Thanks. I agree that this PR not completed. Hope the basePath feature will be added soon. Thanks for your library and attention.

PR #103, opened 2019-12-11
Add support for proto in testutil.CompletedRequest, for testing proto…

yangzh-deepmap:
Add support for proto in testutil.CompletedRequest, for testing proto-related message request/resp.


go.mod
```diff
@@ -4,6 +4,7 @@ require (
 	github.com/cyberdelia/templates v0.0.0-20141128023046-ca7fffd4298c
 	github.com/getkin/kin-openapi v0.2.0
 	github.com/go-chi/chi v4.0.2+incompatible
+	github.com/gogo/protobuf v1.3.1
```
deepmap-marcinr: Could you use github.com/golang/protobuf instead of gogo?

yangzh-deepmap: talked offline, and will use a registry that allows custom response handler.

yangzh-deepmap:
@marcin, I was confused about this Travis CI failure, as the log didn't show anything obvious to me, can you take a look? thanks.

pkg/testutil/request_helpers.go
```diff
@@ -167,13 +167,13 @@ func (c *CompletedRequest) UnmarshalBodyToObject(obj interface{}) error {
...
-		return json.Unmarshal(c.Recorder.Body.Bytes(), obj)
-	default:
-		return fmt.Errorf("no Content-Type on response")
+	content := strings.TrimSpace(contentParts[0])
+	handler := getHandler(content)
+	if handler == nil {
+		return fmt.Errorf("unhandled content: %s", content)
 	}
+
+	return handler(c.Recorder.Body.Bytes(), obj)
```
deepmap-marcinr: Could you just pass the io.ReadCloser directly, instead of bytes() to the handler? It allows it to do different things inside.

pkg/testutil/response_handlers.go
```diff
@@ -0,0 +1,39 @@
...
+
+	RegisterResponseHandler("application/json", jsonHandler)
+}
+
+var (
+	knownHandlersMu sync.Mutex
+	knownHandlers   map[string]ResponseHandler
+)
+
+type ResponseHandler func(raw []byte, obj interface{}) error
```
deepmap-marcinr: It would be useful to also pass the content type into the handler, as it may have useful information.

yangzh-deepmap: thanks, done.

PR #106, opened 2019-12-17
#105 remove special character for go names

daolis:

Issue #105, opened 2019-12-17
Failing generation (formatting) if there are special characters in names


daolis:
Example:
```
  parameters:
    Count:
      name: $count
      in: query
      schema:
        type: boolean
```
Generator uses $count as go variable name and this is not compiling...

StevenACoffman:
Pretty sure this is resolved now! Please take a look!

mromaszewicz:
This has been fixed. Parameter names with special characters like $count are now properly sanitized via SchemaNameToTypeName(), which applies ToCamelCase() (stripping $ as a word separator) and typeNamePrefix(). So $count becomes Count, $top becomes Top, etc. The original parameter name is preserved in the struct tags for correct marshaling.

This was fixed in commit 0751b9d (Remove special characters for go names (#107), Jan 2020).

Closing as completed.

PR #107, opened 2019-12-19
Remove special character for go names

StevenACoffman:
This is a refinement of the work in #106 to address #105 

The OpenID spec I was given had operationIds such as: `POST_student-unenrolled-in-school` which resulted in invalid Go code. It now will more correctly remove special characters and camel case to be `NewPOSTStudentUnenrolledInSchoolRequestWithBody`. 

To  ensure that identifier names remain private (i.e. not exported because they start with a capital letter) they may still appear to be oddly cased as in `pOSTStudentUnenrolledInSchool`.  Correcting that would require prefixing as in `_POSTStudentUnenrolledInSchoolRequest` or `xPOSTStudentUnenrolledInSchoolRequest`.  

Also, use of "anyOf" or "oneOf" resulted in invalid Go code, since it would attempt to write `response.JSON200 = &interface{}{}` . This special cases `typeDefinition.Schema.TypeDecl()` of `interface{}` to generate valid go code, but does not otherwise extend the `anyOf` or `oneOf` support at this point.

I adjusted the travis Go version to 1.13 since the go.mod file specified 1.13, which I think is why the test in #106 appeared to fail in travis, but not locally. This appears to have happened since #84 was merged.


StevenACoffman:
This PR will pass all the tests, but when you run `go generate ./...` you will get capitalization differences:
```
-// addPetJSONBody defines parameters for AddPet.
-type addPetJSONBody NewPet
+// AddPetJSONBody defines parameters for AddPet.
+type AddPetJSONBody NewPet

 // AddPetRequestBody defines body for AddPet for application/json ContentType.
-type AddPetJSONRequestBody addPetJSONBody
+type AddPetJSONRequestBody AddPetJSONBody
```
From the comment, it appears that line three of the `param-types.tmpl` has the `.TypeName` first character uppercased.
```
// {{.TypeName}} defines parameters for {{$opid}}.
```

StevenACoffman:
Ok, I have the tests passing. However, in the existing generated files, I see oddities like in `petstore-client.gen.go`:
```
// ParsefindPetsResponse parses an HTTP response from a FindPetsWithResponse call
```
I would think this should be `ParseFindPetsResponse`, but the genResponseType method is used in too many places where it breaks lots of other things if this returns the correct type. Perhaps something to clean up in another PR if this is accepted.

StevenACoffman:
@daolis can you see if this also meets your needs?

StevenACoffman:
I think I corrected all the situations like `ParsegetTestByNameResponse` to be camel cased properly. This should be good to merge.

pkg/codegen/utils.go
```diff
@@ -57,25 +56,33 @@ func LowercaseFirstCharacter(str string) string {
...
-		for _, inStr := range in {
-			parts := strings.Split(inStr, sep)
-			out = append(out, parts...)
+	separators := "-#@!$&=.+:;_~ (){}[]"
+	s := strings.Trim(str, " ")
+
+	n := ""
+	capNext := true
+	for _, v := range s {
+		if v >= 'A' && v <= 'Z' {
```
deepmap-marcinr: This change using ASCII character ranges makes one big difference - it no longer works properly with unicode characters. If you do it this way, it might be better to use the unicode package and test for IsUpper, isLower, IsNumber, etc. The previous code carefully avoided that by breaking up what we had into words, and capitalizing each word

StevenACoffman: Done! Thanks for taking a look.

PR #110, opened 2019-12-20
chi-server: Add HandlerFromMux

m110:
Hello! 👋 

This is a proposal of adding `HandlerFromMux` function to the generated chi server. It behaves like `Handler`, but allows you to provide your own `Mux`. Thanks to this, you can pass a router with  middlewares enabled.

Example:

```go
r := chi.NewRouter()

r.Use(middleware.Recoverer)
r.Use(middleware.Logger)

handler := HandlerFromMux(server, r)
```

### DIFF
--- a/examples/petstore-expanded/chi/api/petstore.gen.go
+++ b/examples/petstore-expanded/chi/api/petstore.gen.go
@@ -161,8 +161,11 @@ func FindPetByIdCtx(next http.Handler) http.Handler {
 
 // Handler creates http.Handler with routing matching OpenAPI spec.
 func Handler(si ServerInterface) http.Handler {
-	r := chi.NewRouter()
+	return HandlerFromMux(si, chi.NewRouter())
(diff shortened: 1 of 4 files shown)


dannymidnight:
Hello @m110 , looks like I'm _way_ late to the party on this one and wondering if this change was actually required.

Even though there's a Chi server under the hoot it felt nice that we had encapsulated that detail from the calling code whereas this change has opens that up.

Aren't you able to achieve the same thing using `Mount()`?

e.g.

```golang
r := chi.NewRouter()

r.Use(middleware.Recoverer)
r.Use(middleware.Logger)
r.Mount("/", Handler(server, r))
```

m110:
@dannymidnight Hey. I guess you're right, I missed it. :(

dannymidnight:
All good! I think it points out there's some missing docs on how to use this server.

PR #111, opened 2019-12-31
keep path suffix "/" if exists

jbking:
Path pattern can have suffix "/", like `/pets/` . In this case, the suffix "/" have to keep for requesting.
However its dropped by `path.Join()`

```
package main

import (
	"fmt"
	"path"
)

func main() {
	queryPath := "/prefix"
	fmt.Println(path.Join(queryPath, fmt.Sprintf("/pets")))
	fmt.Println(path.Join(queryPath, fmt.Sprintf("/pets/"))) // same output above
	fmt.Println(path.Join(queryPath, fmt.Sprintf("/pets/")) + "/") // keep suffix "/"
}
```

```
$ go run main.go
/prefix/pets
/prefix/pets
/prefix/pets/
```

Dropping the suffix "/" can cause a redirect response from a service for appending suffix "/". And it makes creating resource (POST) failed.

### DIFF
--- a/pkg/codegen/templates/client.tmpl
+++ b/pkg/codegen/templates/client.tmpl
@@ -165,6 +165,9 @@ func New{{$opid}}Request{{if .HasBody}}WithBody{{end}}(server string{{genParamAr
         return nil, err
     }
     queryUrl.Path = path.Join(queryUrl.Path, fmt.Sprintf("{{genParamFmtString .Path}}"{{range $paramIdx, $param := .PathParams}}, pathParam{{$paramIdx}}{{end}}))
+    if strings.HasSuffix("{{genParamFmtString .Path}}", "/") && ! strings.HasSuffix(queryUrl.Path, "/") {
+        queryUrl.Path += "/"
+    }
 {{if .QueryParams}}
     queryValues := queryUrl.Query()
 {{range $paramIdx, $param := .QueryParams}}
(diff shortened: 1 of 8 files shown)


pkg/codegen/templates/client.tmpl
```diff
@@ -165,6 +165,9 @@ func New{{$opid}}Request{{if .HasBody}}WithBody{{end}}(server string{{genParamAr
         return nil, err
     }
     queryUrl.Path = path.Join(queryUrl.Path, fmt.Sprintf("{{genParamFmtString .Path}}"{{range $paramIdx, $param := .PathParams}}, pathParam{{$paramIdx}}{{end}}))
+    if strings.HasSuffix("{{genParamFmtString .Path}}", "/") && ! strings.HasSuffix(queryUrl.Path, "/") {
```
deepmap-marcinr: Thank you for doing this, but this is a bit hacky and does a bunch of unnecessary string compares.

You already have queryUrl in the context, we can use that to concatenate paths properly while preserving trailing slashes - I should have thought of it when writing the code in the first place.

Try queryUrl.Parse(fmt.Sprintf("{{genParamFmtString .Path}}"{{range $paramIdx, $param := .PathParams}}, pathParam{{$paramIdx}}{{end}})

Basically, queryUrl.Parse can parse a relative url in the context of the server string.

deepmap-marcinr:
I've committed an alternate fix myself.

pkg/codegen/templates/client.tmpl
jbking: Thx!

Issue #104, opened 2019-12-13
incorrect fill params with date type


apopov-app:
Hello, i found bug again :)

If i use url parameter with "format: date" it generate model code:
```
Date *openapi_types.Date `json:"date,omitempty"`
```

this code in middleware:
```
err = runtime.BindQueryParameter("form", true, false, "date", r.URL.Query(), &params.Date)
```
	
I think the reflex inside runtime.BindQueryParameter determines that this is a structure and tries to fill the field "as a structure". 

As a result, to fill out this type, you must complete the request:
localhost/pets?Time=2006-01-02T15:04:05.999999999Z&limit=111 
(*openapi_types.Date contains inside time.Time)

instead
localhost/pets?date=2019-10-01&limit=111

Swagger example:
```
     parameters:
        - name: date
          in: query
          description: test date
          required: false
          schema:
            type: string
            format: date
```

Demo project: https://github.com/ruelephant/petstore

mromaszewicz:
This has been fixed. The Date type in github.com/oapi-codegen/runtime/types now implements UnmarshalText (added in commit d38f1c9, July 2022), which allows runtime.BindQueryParameter to correctly parse date strings like 2019-10-01 instead of trying to fill struct fields via reflection.

Closing as completed.

apopov-app:
opened on Dec 13, 2019

<img width="224" height="180" alt="Image" src="https://github.com/user-attachments/assets/026fdc87-7ccb-46a8-bc3e-18da9c5f93ef" />

PR #112, opened 2020-01-02
Fixing query parsing of date parameters

jmcnevin:
Looking at [this issue](https://github.com/deepmap/oapi-codegen/issues/104), I neglected to handle query parsing on the server side when implementing the custom Date type.

This attempts to rectify that, with the caveat that this could probably stand a more elegant solution if additional custom types are added.

### DIFF
--- a/pkg/runtime/bindparam.go
+++ b/pkg/runtime/bindparam.go
@@ -21,6 +21,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/deepmap/oapi-codegen/pkg/types"
 	"github.com/labstack/echo/v4"
 )
 
@@ -299,6 +300,7 @@ func BindQueryParameter(style string, explode bool, required bool, paramName str
 			// different things based on destination type.
 			values, found := queryParams[paramName]
 			var err error
+
 			switch k {
 			case reflect.Slice:
 				// In the slice case, we simply use the arguments provided by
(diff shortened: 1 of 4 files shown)


ticruz38:
Still buggy for format date
```
// openapi.yml
day:
      type: object
      description: The day is either workday, weekend or holiday
      properties:
        date:
          type: string
          format: date
```

It looks like the format: date, in opeanapi.yml get generated as time.Time and not dates.Type, resulting in the wrong casting of specific types in bindParamsToExplodedObject method
```
// special handling for custom types
	switch dest.(type) {
	case *types.Date:
		return BindStringToObject(values.Get(paramName), dest)
```
we never fall in *types.Date case, so the date parameter is never formatted, we end up with an empty time.Time struct.
I don't know what would be the proper way to fix this. I'm fine with the time.Time being generated, we could try formatting both for RFC3339 (for date-time) or 2006-01-02 (for date), if none works, return an error
```
case reflect.Struct:
		switch dstType := dst.(type) {
		case *time.Time:
			// Time is a special case of a struct that we handle
			parsedTime, err := time.Parse(time.RFC3339Nano, src)
			if err != nil {
                                parsedTime, err := time.Parse("2006-01-02", src)
                                if err != nil {
				          return fmt.Errorf("error parsing '%s' as RFC3339 or 2006-01-02, time: %s", src, err)
			}
			*dstType = parsedTime
			return nil
```

PR #113, opened 2020-01-02
Use regex to figure out optional imports

nadir-promo:
When using "Audio" (or any other word that ends with "io") as name of a schema, the generator was adding the "io" package to imports. It caused "imported and not used" error.
Using this regex should fix the issue.

### DIFF
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"fmt"
 	"go/format"
+	"regexp"
 	"sort"
 	"strings"
 	"text/template"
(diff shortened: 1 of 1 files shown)


pkg/codegen/codegen.go
```diff
@@ -56,28 +57,28 @@ type goImports []goImport
...
-		{lookFor: "openapi_types.", alias: "openapi_types", packageName: "github.com/deepmap/oapi-codegen/pkg/types"},
-		{lookFor: "path.", packageName: "path"},
-		{lookFor: "runtime.", packageName: "github.com/deepmap/oapi-codegen/pkg/runtime"},
-		{lookFor: "strings.", packageName: "strings"},
-		{lookFor: "time.Duration", packageName: "time"},
-		{lookFor: "time.Time", packageName: "time"},
-		{lookFor: "url.", packageName: "net/url"},
-		{lookFor: "xml.", packageName: "encoding/xml"},
-		{lookFor: "yaml.", packageName: "gopkg.in/yaml.v2"},
+		{lookFor: "base64\\.", packageName: "encoding/base64"},
```
deepmap-marcinr: ha! thanks for this fix. I was wondering when it would bite us.

PR #119, opened 2020-01-17
Pass validator's errors inside of echo.HTTPError

derfenix:
Wondering why validator's errors silenced, when they can be very useful within custom echo.HTTPErrorHandler. For example for building custom 400 error responses with error field name, or for more verbose logging on SecurityRequirementsError.

So just add original error into echo.HTTPError{Internal}, make it available at the application's level.


deepmap-marcinr:
The "why" is because this code was rushed :)

PR #121, opened 2020-01-20
Generate Comments for Struct Fields

srueg:
The first commit introduces a test (which fails) and the second commit the implementation (to fix the test).


srueg:
Failing build: https://travis-ci.com/deepmap/oapi-codegen/builds/145175906

PR #130, opened 2020-02-06
formatting both date and date-time resolve #112

ticruz38:


ticruz38:
#112

pkg/runtime/bindparam.go
```diff
@@ -436,6 +437,9 @@ func bindParamsToExplodedObject(paramName string, values url.Values, dest interf
 	switch dest.(type) {
 	case *types.Date:
 		return BindStringToObject(values.Get(paramName), dest)
```
deepmap-marcinr: Would you be willing to change this a bit - 

In here, we know if something is a time or a date, but that context is lost in BindStringToObject.

I propose two more functions - BindStringToTime and BindStringToDate, each called from the correct place in this case statement.

If you do that, you don't have to do the fall back of failing to parse a time, and then trying a date.

ticruz38: Maybe the problems belong to the generator then.
I assume date should be generated as types.Date, and date-time as time.Time yes?
Currently both oapi format (date and date-time) are generated as time.Time.
For me it's fine, but as you say, there is no way to differenciate them at formatting with time.Time.
For now, unless we fix the generator to differenciate both oapi format with different types (as it is supposed to do), we have to try catch both string at formatting. I just don't see a more elegant way.

ticruz38: to be more specific, currently, the types.Date case is never used.

deepmap-marcinr:
You're right, I missed that. Good enough for now!

PR #131, opened 2020-02-07
Another attempted fix for date param parsing

jmcnevin:
This is related to https://github.com/deepmap/oapi-codegen/pull/130, so perhaps @ticruz38 would like to collaborate here.  I've attempted to bring in his changes `pkg/runtime/bindstring.go`.

We have continued to find problems with query string date parameters not being parsed properly, and this was related to the fact that the generator is creating new types for the Params structs, like so:

```go
type At openapi_types.Date
```

This makes it difficult to reason about how to handle these types, since a simple type switch won't work.  I've attempted to create a unique signature for our custom type that will help us reflect on what these new types are and how they should be parsed.

This strategy seems to be working in our manual tests, but I will admit to being a little uneasy about using reflect in this way.  Any input is appreciated.

### DIFF
--- a/examples/petstore-expanded/chi/petstore_test.go
+++ b/examples/petstore-expanded/chi/petstore_test.go
@@ -71,8 +71,8 @@ func TestPetStore(t *testing.T) {
 
 	t.Run("List all pets", func(t *testing.T) {
 		store.Pets = map[int64]api.Pet{
-			1: api.Pet{},
-			2: api.Pet{},
+			1: {},
+			2: {},
 		}
 
 		// Now, list all pets, we should have two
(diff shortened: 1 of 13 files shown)


jmcnevin:
Closing this until we've vetted these changes a bit more.

ticruz38:
For any date related parameter, I suggest that the generator always emit a time.Time type, we should then figure out how to format these, either RFC3339 (for date-time) or 2006-01-02 (for date).
This is currently the case, the generator only emit time.Time for date parameters, it works for date-time, which is generated as opeapi.Date, and has a special switch case, but does'nt for date params that are left to the reflect strategy. My pull request add a switch case for time.Time and handle the parsing for both case.

Issue #126, opened 2020-02-04
Generated API Client not respecting basePath when passing baseURL


pseudo-su:
Possibly related: https://github.com/deepmap/oapi-codegen/issues/55 (but more specifically issues with the generated client)

Using this: https://swagger.io/docs/specification/api-host-and-base-path/
<img width="740" alt="Screen Shot 2020-02-04 at 1 40 55 pm" src="https://user-images.githubusercontent.com/4339582/73708998-03838f80-4754-11ea-9dfd-b448716b9a3a.png">

## Problem

Given a swagger spec that looks like this:
```yaml
openapi: 3.0.0
servers:
  - url: https://my-server.com/v1/some/base/path
paths:
  /things:
    get:
      operationId: getThings
      responses:
        "200":
          content:
            application/json:
              schema:
                type: object
```

And an api client initialized like this:
```go
client := NewClientWithResponses("https://my-server.com/v1/some/base/path")

resp, err := client.GetThingsWithResponse(context.TODO())
```

The HTTP request sent by the generated client isn't what I expected it to be

Expected request: `GET https://my-server.com/v1/some/base/path/things`
Actual request: `GET https://my-server.com/things`

## Investigation

Looking through the generated code, this seems to be the cause of my problem (comments added for explaination):

```golang
// when server == "https://my-server.com/v1/some/base/path"
queryUrl, err := url.Parse(server)
if err != nil {
	return nil, err
}
// passing the absolute path "/things" overwrites the existing basePath "/v1/some/base/path"
queryUrl, err = queryUrl.Parse("/things")
if err != nil {
	return nil, err
}
```
I think according to the docs [here](https://golang.org/pkg/net/url/#URL.Parse) it should be `queryUrl.Parse("things")` instead so it doesn't overwrite the basePath in the baseURL. But we would also have to make sure to append a "/" to the end of the baseURL if it isn't present

<img width="851" alt="Screen Shot 2020-02-04 at 5 03 39 pm" src="https://user-images.githubusercontent.com/4339582/73718345-54a17c80-4770-11ea-91dd-a8319f4c0db3.png">

repl.it link: https://repl.it/repls/ColorlessNumbArchitects
<img width="1092" alt="Screen Shot 2020-02-04 at 5 22 28 pm" src="https://user-images.githubusercontent.com/4339582/73719252-ef9b5600-4772-11ea-8fdc-0bc940f41f93.png">

pseudo-su:
[...]

PR #134, opened 2020-02-11
Fix for: Generated API Client not respecting basePath when passing baseURL #126 (updated client.tmpl)

apechenev:
Respecting already set basePath in base URL when resolving ref URL in New*Request function


srueg:
Just hit that bug in my generated client. :+1+

PR #136, opened 2020-02-18
Add support for user-provided templates

falzm:
This change introduces a new `-templates` command flag to enable users
to override code generation templates at runtime instead of having to
recompile the binary after having edited the built-in templates.

### DIFF
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -46,7 +46,7 @@ func main() {
 	flag.StringVar(&outputFile, "o", "", "Where to output generated code, stdout is default")
 	flag.StringVar(&includeTags, "include-tags", "", "Only include operations with the given tags. Comma-separated list of tags.")
 	flag.StringVar(&excludeTags, "exclude-tags", "", "Exclude operations that are tagged with the given tags. Comma-separated list of tags.")
-	flag.StringVar(&templatesDir, "templates", "", "Path to directory containing template overrides")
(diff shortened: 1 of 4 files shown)


falzm:
@deepmap-marcinr let me know if this is an acceptable change to be merged upstream: we'd like to use your software at [Exoscale](https://www.exoscale.com/) and we'd greatly appreciate not having to fork it 😉

cmd/oapi-codegen/oapi-codegen.go
```diff
@@ -125,3 +134,46 @@ func splitCSVArg(input string) []string {
 	}
 	return args
 }
+
+func loadTemplateOverrides(templatesDir string) (map[string]string, error) {
+	var (
+		knownTemplates = map[string]interface{}{
```
deepmap-marcinr: I don't like having a hard coded file list - it's another place to forget to update.

We also don't really need to load all templates either, since the Chi and Echo stuff is mutually exclusive, for example.

It's a bummer that that the templates compiler I'm using doesn't export the file map, then this would be much nicer!

Can we simply enumerate the files in the templates directory? - if the filename matches, good, it'll override the original, if it doesn't match, you get no behavior change.

falzm: I've just pushed a follow-up commit (a9a410b) that implements your suggestion, WDYT?

deepmap-marcinr:
Thanks for doing this.

falzm:
No problem, thank you for merging!

PR #142, opened 2020-02-27
Respect an http error returned by the authentication function

tschaub:
Currently, the `OapiRequestValidator` middleware always responds with a 403 if all authentication functions return an error.  The authentication function should be able to distinguish a 401 (missing credentials) from a 403 (invalid credentials).  The change in this branch makes it so the middleware respects an `echo.HTTPError` if one is returned from an authentication function.  The tests demonstrate how the middleware can be configured to generate a 401 instead of a 403.

### DIFF
--- a/pkg/middleware/oapi_validate_test.go
+++ b/pkg/middleware/oapi_validate_test.go
@@ -87,6 +87,15 @@ paths:
       responses:
         '204':
           description: no content
+  /protected_resource_401:
+    get:
+      operationId: getProtectedResource
+      security:
+        - BearerAuth:
+          - unauthorized
+      responses:
+        '401':
+          description: no content
 components:
   securitySchemes:
     BearerAuth:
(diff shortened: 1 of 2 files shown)


pkg/middleware/oapi_validate_test.go
```diff
@@ -87,6 +87,15 @@ paths:
       responses:
         '204':
           description: no content
+  /protected_resource_401:
+    get:
+      operationId: getProtectedResource
+      security:
+        - BearerAuth:
+          - unauthorized
```
deepmap-marcinr: It's kind of odd to have an authorization scope named "unauthorized" which returns a 401, because you still have to have that scope present in your credentials. I see how it fits into the test, though, but this isn't the common way you'd do unauthorized errors.

tschaub: I just named it this way for the tests.

I hope the point of the functionality was clear.  Clients should get a 401 (e.g. with a www-authenticate header) if credentials are absent so they can retry with appropriate credentials.  The previous behavior of returning a 403 meant re-authenticating would make no difference.

PR #146, opened 2020-03-06
Property customization support

qystishere:
I use a static schema for JSON in the project, so it was inconvenient with constant omitempty in the json tag.

I update examples & added support for:
1. Changing property nullability.
2. Ability to specify whether to use `omitempty` option for json tag. 
image
3. Specifying a custom tag for the property.
image
4. CLI-flag `-no-omitempty` to avoid `omitempty` option for json tags by default.


qystishere:
@deepmap-marcinr can you review it?

pkg/codegen/schema.go
```diff
@@ -285,6 +297,48 @@ type FieldDescriptor struct {
...
+		Schema:             s,
+	}
+	if sRef == nil || sRef.Value == nil {
+		return prop, nil
+	}
+	if description == "" {
+		prop.Description = sRef.Value.Description
+	}
+	for name, value := range sRef.Value.Extensions {
+		if len(name) < 6 || !strings.HasPrefix(name, "x-go-") {
```
deepmap-marcinr: You don't need the length check. HasPrefix will work just fine without.

pkg/codegen/schema.go
```diff
@@ -53,10 +60,24 @@ func (s Schema) GetAdditionalTypeDefs() []TypeDefinition {
 }
 
 type Property struct {
-	Description   string
-	JsonFieldName string
-	Schema        Schema
-	Required      bool
+	Description        string
+	Required           bool
+	FieldNullable      bool
```
deepmap-marcinr: This code is a little bit of an ill fit with the rest of the rather complicated code in here.

Required and FieldNullable are kind of related. FieldTag, JsonFieldName, and JsonFieldOmitEmpty are also a bit special case relative to the very general usage of this Property.

What would you think about parsing all x-go-XXX extensions into a map, which you then store on this object, so that it's easier to add future extensions. Then, you create getter functions on Property which look into that map.

For the case of not nullable fields, you could rename Required to required and add a function called Required() which would then be used by the templates, and it will achieve what you want, by using the nullable extension to actually make things required.

pkg/codegen/schema.go
```diff
@@ -65,7 +86,7 @@ func (p Property) GoFieldName() string {
 
 func (p Property) GoTypeDef() string {
 	typeDef := p.Schema.TypeDecl()
-	if !p.Schema.SkipOptionalPointer && !p.Required {
```
deepmap-marcinr: This will cause regressions. What if something is required but FieldNullable is at its default zerovalue of false? See my note about having a smarter Required() function. I don't like having two roughly synonymous flags.

qystishere:
@deepmap-marcinr I made use of extensions from one map, but did not understand why we need the Required() function.

If we need a field, but we want to get a pointer to it, and not an empty string, for example. Or the field is optional, but we want an empty string instead, not a null value. Will this cause any problems?

Check out the current implementation, please.

PR #148, opened 2020-03-10
Follow golang standards when passing in context

ashwinrs:
Change the order of arguments for RequestEditorFn to have `context` be the first argument.

Golang standards specify that context should be the first parameter: https://golang.org/pkg/context/ . If not the lint will complain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ type Client struct {
 
     // A callback for modifying requests which are generated before sending over
     // the network.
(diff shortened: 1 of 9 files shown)


deepmap-marcinr:
I should have caught that in code review earlier, thanks. It started out with req only, then context got added later.

ashwinrs:
@deepmap-marcinr  Thank you for making this library, its been super helpful.

PR #152, opened 2020-03-15
Fix time serialization panic (reflect.Value.Interface: cannot return value obtained from unexported field or method)

### LINKED ISSUE #153: OAS 'date-time' type in query parameter causes panic in client
Using time (date-time) parameter in client causes the following panic when the parameter is set:

`panic: reflect.Value.Interface: cannot return value obtained from unexported field or method`

Example OAS spec:
```
paths:
  /test:
    description: Test API
    get:
      tags:
        - test
      summary: Get Test
      operationId: GetTest
      parameters:
        - in: query
          description: Timestamp
          name: ts
          schema:
            type: string
            format: 'date-time'
            description: Timestamp
```

Suggested fix in PR
#152

joonashaapsaari:
Fixes #153 

- Fix time serialization
  * For query parameter, time is stored as a direct reference
    (non-pointer) which was previously not supported and
    caused the panic:
    panic: reflect.Value.Interface: cannot return value obtained from unexported field or method
  * It should not be possible to serialize time alone as a
    deepObject, as it should be handled as primitive.
- Add tests for time serialization

### DIFF
--- a/pkg/runtime/styleparam.go
+++ b/pkg/runtime/styleparam.go
@@ -132,18 +132,31 @@ func sortedKeys(strMap map[string]string) []string {
 	return keys
 }
 
+
+// This is a special case. The struct may be a time, in which case, marshal
+// it in RFC3339 format.
+func marshalTimeValue(value interface{}) (string, bool) {
+	if timeVal, ok := value.(*time.Time); ok {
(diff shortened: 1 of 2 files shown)


StevenACoffman:
@deepmap-marcinr I tested this and verified this fixes another of our issues.

Issue #151, opened 2020-03-15
Feature proposal: extend include-tags and exclude-tags to affect generating types


pseudo-su:
I'd like to propose extending the `--exclude-tags` and `--include-tags` functionality to affect types generation.

Currently when I use `exclude-tags` or `include-tags` to generate based on a subset of my spec, the leftover types that aren't used in the subset i've selected still end up in my generated models.

Example:
https://github.com/pseudo-su/golang-service-template/pull/1

I believe that it should be possible to walk the openapi spec after the `tags` filter has been applied and prune any models that are identified as being unused in any operations.

I'd be happy to investigate implementing this if it's something that would likely be accepted

PR #154, opened 2020-03-15
feat: prune unused components by default

pseudo-su:
## Changes

- Added [`.editorconfig`](https://editorconfig.org/) file (can remove this if you want but IMO it's handy to have and helped fix my whitespace formatting).
- Added a `skip-prune` flag that is similar to `skip-fmt` in that it can be used to prevent running the pruning logic when generating code.
- Rather than making the test code generation use the `skip-prune` flag (which would be a quick way to make the tests pass without updating the specs) I thought it was better to update them and make sure than any components that would have been pruned are referred to in at least one operation (so they don't get pruned). This means that `test/components/components.yaml`, `test/schemas/schemas.yaml` and `test/test-schema.yaml` have a number of additions to the generated code but they're not **missing** anything apart from this https://github.com/deepmap/oapi-codegen/pull/154/files#r392671169
- I moved the filtering logic of the `openapi3.Swagger` from `codegen.go` into `filter.go` (can move these back if it's preferred).
- I have two main types of test in `prune_test.go`
  - `pruneComprehensiveTestFixture` has a fixture where with every type of component and none of them are referred to by an operation. This is to test that it's correctly removing all types of components if they're not referenced.
  - `pruneSpecTestFixture` has a `GET /cat` operation tagged with `cat` and a `GET /dog` operation tagged with `dog`. Both operations share a reference to `#/components/schemas/Error`. The test ensures that when you filter by one tag you don't get models that are only used in the other.

## Notes

Fixes https://github.com/deepmap/oapi-codegen/issues/151

### DIFF
--- a/internal/test/components/components.yaml
+++ b/internal/test/components/components.yaml
@@ -5,6 +5,38 @@ info:
   license:
     name: MIT
 paths:
+  /ensure-everything-is-referenced:
+    get:
+      operationId: ensureEverythingIsReferenced
+      description: |
+        This endpoint exists so that components can be created in this
+        spec and not be pruned
+    # TODO: figure out why uncommenting this causes failures
+    #   parameters:
+    #     - $ref: "#/components/parameters/ParameterObject"
+      requestBody:
+        $ref: "#/components/requestBodies/RequestBody"
+      responses:
+        200:
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  one:
+                    $ref: "#/components/schemas/AdditionalPropertiesObject1"
+                  two:
+                    $ref: "#/components/schemas/AdditionalPropertiesObject2"
+                  three:
+                    $ref: "#/components/schemas/AdditionalPropertiesObject3"
+                  four:
+                    $ref: "#/components/schemas/AdditionalPropertiesObject4"
+                  five:
+                    $ref: "#/components/schemas/AdditionalPropertiesObject5"
+                  jsonField:
+                    $ref: "#/components/schemas/ObjectWithJsonField"
+        default:
+          $ref: "#/components/responses/ResponseObject"
   /params_with_add_props:
     get:
       operationId: ParamsWithAddProps
--- a/internal/test/components/components.yaml
+++ b/internal/test/components/components.yaml
@@ -173,6 +173,7 @@ components:
             type: string
   parameters:
     ParameterObject:
+      name: ParameterObject
       description: a parameter
       in: query
       content:
--- a/pkg/codegen/prune.go
+++ b/pkg/codegen/prune.go
@@ -1,0 +1,487 @@
+package codegen
+
+import (
+	"fmt"
+
+	"github.com/getkin/kin-openapi/openapi3"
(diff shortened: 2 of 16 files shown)


internal/test/components/components.yaml
```diff
@@ -5,6 +5,38 @@ info:
   license:
     name: MIT
 paths:
+  /ensure-everything-is-referenced:
+    get:
+      operationId: ensureEverythingIsReferenced
+      description: |
+        This endpoint exists so that components can be created in this
+        spec and not be pruned
+    # TODO: figure out why uncommenting this causes failures
```
pseudo-su: Uncommenting this causes the following error

```sh
error generating code: error formatting Go code: 990:17: expected selector or type assertion, found '!=' (and 10 more errors)
```

I'm unsure exactly what it is but it seems like it might be an pre-existing issue with the code generation.

pkg/codegen/prune.go
```diff
@@ -0,0 +1,487 @@
...
+		return err
+	}
+	if !shouldContinue {
+		return nil
+	}
+	if ref.Value == nil {
+		return nil
+	}
+
+	// TODO: for some reason linkRef.Value.Headers is a `map[string]*Schema` instead of a `map[string]*SchemaRef`
```
pseudo-su: I think there's something weird going on with the type for `linkRef.Value.Headers`

I've started a conversation here https://github.com/getkin/kin-openapi/issues/188

internal/test/components/components.yaml
pseudo-su: I made some progress with this, the first problem was that the `ParmeterObject` was missing a `name`.

After fixing that, the new problem is that there are instances in both the server and client generated code where the generated code assumes that the parameter will be of type `string` but it is actually the `ParameterObject` struct.

StevenACoffman:
This needs a rebase!

StevenACoffman:
@deepmap-marcinr Ok, this is clear for merge now!

deepmap-marcinr:
Could we invert the flag, call it prune-unused? Have it maintain current behavior without the flag, and you can pass it in if you want the pruning?

pseudo-su:
@deepmap-marcinr I can do that, but before I do I just wanted to mention why I decided on prune by default when I was thinking about it:

1. Keeps the flag name consistent alongside `skip-fmt`
2. It doesn't **seem** like a breaking change on upgrade because if schemas are never referenced elsewhere in the spec then generated code from this library never would've needed it. The only situation where types would dissapear when consumers of this library upgrade is when users have "bugs" in their OpenAPI specs. On top of that, in order to manifest as an actual issue they would have to be relying on the generated types in code they've manually written.
3. The pruning is actually pretty useful in detecting errors/missing-references in OpenAPI specs.

deepmap-marcinr:
Ok, lets try it like it is.

sakethrampotturi:
Hi! When can I expect a release that has this feature?

deepmap-marcinr:
Right now :) Sorry, I've been slammed and neglecting this a bit.

PR #155, opened 2020-03-16
fix: operations not appending to server url

pseudo-su:
If you provide a server of `https://my-url.com/path/v1` and you have a `GET /things` operation the correct full path for the operation is `https://my-url.com/path/v1/things`.

Current behavior is that if you don't append a trailing slash to your server string before you pass it to `NewClient(server string, opts ...ClientOption)` then for the above case you will end up with `https://my-url.com/path/things` because that's how go's `url.Parse` treats relative URLs.

## Notes

Fixes https://github.com/deepmap/oapi-codegen/issues/126

## Other

Somewhat related but the more I look into this the more I think that the `WithBaseURL` option doesn't actually make much sense and should probably be discouraged.

I have two main problems with it:

- Ultimately all it's actually doing is setting the  `client.Server` string, but that's already handled by passing it in as a string to `NewClient` so I don't understand why it's even needed.
- The name `WithBaseURL` doesn't align to the OpenAPI spec terminology, if an option like this had to exist I think it'd make more sense for it to be called `WithServer(serverURL string)`.
- I think [this](https://github.com/deepmap/oapi-codegen/pull/155/files#diff-04c6e90faac2675aa89e2176d2eec7d8) fix to the `README.md` docs shows how it's a bit clunky

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,7 +369,7 @@ which help you to use the various OpenAPI 3 Authentication mechanism.
         //
         // WithHTTPClient(httpClient *http.Client)
         //
-        client, clientErr := NewClient(context.Background(), []ClientOption{
+        client, clientErr := NewClient("https://api.deepmap.com", []ClientOption{
             WithBaseURL("https://api.deepmap.com"),
             WithRequestEditorFn(apiKeyProvider.Edit),
         }...,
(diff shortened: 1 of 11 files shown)


StevenACoffman:
@deepmap-marcinr FYI, ready to merge!

Issue #125, opened 2020-01-29
remove echo dependancy for chi server


aca:
Chi server shares same functions in runtime package which uses echo.
Removing unnecessary dependency against echo would make chi server more simple

qmuntal:
`echo` is used for two purposes in `pkg/runtime`:

1. In `EchoRouter` interface methods
2. In `BindStyledParameter` and `BindQueryParameter` to return `echo.HTTPError` instead of a plain error.

The `EchoRouter` interface could be moved from the runtime to the echo specific templates, as it is not used in the runtime.

The two other methods returning echo.HTTPError could return instead a plain `fmt.Errorf`, as the auto generate code calling them are already adding the necessary context to the error:

***echo server***
```go
err = runtime.BindStyledParameter("simple", false, "id", ctx.Param("id"), &id)
if err != nil {
    return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter id: %s", err))
}
```

***chi server***
```go
err = runtime.BindStyledParameter("simple", false, "id", chi.URLParam(r, "id"), &id)
if err != nil {
    http.Error(w, fmt.Sprintf("Invalid format for parameter id: %s", err), http.StatusBadRequest)
    return
}
```

PR #164, opened 2020-04-01
Remove echo dependency from bindparam.go

qmuntal:
Updates #125

### DIFF
--- a/pkg/runtime/bindparam.go
+++ b/pkg/runtime/bindparam.go
@@ -50,11 +50,7 @@ func BindStyledParameter(style string, explode bool, paramName string,
 			return err
 		}
 
-		err = bindSplitPartsToDestinationStruct(paramName, parts, explode, dest)
-		if err != nil {
-			return err
-		}
-		return nil
+		return bindSplitPartsToDestinationStruct(paramName, parts, explode, dest)
 	}
 
 	if t.Kind() == reflect.Slice {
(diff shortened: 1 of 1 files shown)


pkg/runtime/bindparam.go
```diff
@@ -35,7 +33,7 @@ func BindStyledParameter(style string, explode bool, paramName string,
 	value string, dest interface{}) error {
 
 	if value == "" {
-		return echo.NewHTTPError(http.StatusBadRequest, "parameter '%s' is empty, can't bind its value", paramName)
+		return fmt.Errorf("parameter '%s' is empty, can't bind its value", paramName)
```
deepmap-marcinr, line 36: This change has one unfortunate side effect. Now, when invalid parameters are specified, what were formerly HTTP/400 errors will now be HTTP/500 errors. So, we need to either return some error wrapper with a status code which the Chi or Echo layer can turn into the relevant error, or all the call sites in codegen need to make this an HTTP/400.

qmuntal: Looking at the generate code and the templates, it seems that in both codegen this error is already wrapped into a HTTP/400 error. See [wrappers.tmpl#L39](https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/templates/wrappers.tmpl#L39) and [chi-middleware.tmpl#L75](https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/templates/chi-middleware.tmpl#L75).

deepmap-marcinr: You are correct. I didn't notice.

PR #166, opened 2020-04-02
Wrap Chi server and pass parameters as function params

qmuntal:
Wrap the Chi server in the same way it is being done with Echo: 
- Pass in mandatory path arguments as function arguments
- Pass in query, header and cookie params struct only if there is at least on parameter of these types.

This way the resulting code is more symmetric with Echo and path parameters can be obtained with consulting opaque context values.

The `http.ResponseWriter` and `http.Request` are still passed to the server interface because the server is still responsible of writing the response and not all entities are yet managed by the auto generated code, such as decoding the `request.Body`.


dannymidnight:
This  is something that's come up in the past https://github.com/deepmap/oapi-codegen/pull/56#issuecomment-536011721 and looks like a decent change :+1:

mikekonan:
@deepmap-marcinr Is there is any chance these changes will be merged?

pgolm:
Awesome improvement 👍 . But unfortunately this is a breaking change. I'm not sure if this project uses semver, but it would be nice if you could release this as major change instead as patch the next time.

deepmap-marcinr:
This project is in an interesting place. It's a code generator, with a
small runtime. I've been strict about semver with anything under
pkg/runtime, since a simple "go get -u" should not break existing generated
code. However, many changes to code generation are breaking changes. If I
treated this as a runtime for semver purposes, I'd probably be on
oapi-codegen/v50 right now, and not quite sure how bug fixes would work
across so many versions.

Anyhow, I'm open to suggestions how to manage it better, but it's been
challenging. I've been thinking of creating a V2 in the near future which
is severely incompatible, because some improvements would create very
difficult refactorings for users, and I've been avoiding those. In
retrospect, I should have insisted that Chi handlers looked like Echo
handlers from the beginning.

ashwinrs:
It broke our pipeline as well. We should try and avoid breaking changes to be introduced without updating the major version. Or find another alternative.

PR #167, opened 2020-04-03
feat: make responses public and add interface

pseudo-su:
* The `ClientWithResponses` struct was missing a corresponding `ClientWithResponsesInterface`
* The values returned by the `ClientWithResponses` client were lower-case/private which makes it impossible for consumers to create interfaces for the client.


StevenACoffman:
Nice improvement. Thanks. @deepmap-marcinr This worked for me.

StevenACoffman:
@deepmap-marcinr This is useful and works for our organization. If you get a chance, please consider merging. #169 is a duplicate of this PR.

PR #168, opened 2020-04-03
Feature/add flag to generate private API client

pseudo-su:
This includes the fixes in https://github.com/deepmap/oapi-codegen/pull/167 with the additional feature of allowing users to pass a `--make-private` CLI flag that allows users to toggle visibility of the client (switch it to being defined as lowercase).

USAGE:
```
--make-private=client,client-with-responses
```

EDIT: removing the `ClientWithResponses` turns this into just
```
--make-private=client
```

## The problem

It can be somewhat confusing to package consumers when there are multiple interfaces/clients available in the same package (EG: `ClientInterface` and `ClientInterfaceWithResponses`), it's not necessarily obvious which one they should use. Additionally if you want to create a manually written client wrapping the generated one you might end up with *another* named client (EG `Client`, `ClientWithResponses`, `ClientWithGoStyleErrors`). Giving the ability to generate the clients you don't expect people to use as private helps reduce that problem.

## Use cases

I have two main use cases where I believe this is useful:

### First

Generate an API client package alongside a service (EG: `github.com/me/my-service/pkg`) and only expose the `ClientWithResponses` for consumers of the package. This has the effect of encouraging/forcing people into always using the `ClientWithResponses` client/interface.

### Second

Generate both the `client` **and** the `clientWithResponses` as private. This is helpful when it's desirable to capture things beyond the scope of what openapi+golang can reasonably support. This makes manually writing wrappers for the generated client easier/clearer. Accomodating this is somewhat annoying at the moment because the generated code "uses up" the names that you would want to use for the client (`pkg.Client`, `pkg.ClientInterface`).

If I was able to generate this client (as lowercase `generatedClient`, `generatedClientInterface` and `newGeneratedClient`):
```golang
type generatedClient struct {
  /* etc */
}

var _ generatedClientInterface = &generatedClient{}

/* generated methods on client etc ... */

// newGeneratedClient ...
func newGeneratedClient(server string, opts ...ClientOption) (*generatedClient, error) {
	/* etc */
}

type generatedClientInterface {
  PostJson(ctx context.Context, body PostJsonJSONRequestBody) (*http.Response error)
  PostJsonWithResponse(ctx context.Context, body PostJsonJSONRequestBody) (*PostJsonResponse, error)
}

// SchemaObject defines model for SchemaObject.
type SchemaObject struct {
	FirstName string `json:"firstName"`
	Role      string `json:"role"`
}

type Error struct {
	HttpStatusCode string `json:"httpStatusCode"`
	ErrorCode string `json:"errorCode"`
	Message string `json:"message"`
}

type PostJsonResponse struct {
	Body         []byte
	HTTPResponse *http.Response
	JSON200      *SchemaObject
	JSONDefault  *Error
}
```
I would then be able to create a lightweight wrapper for the generated client to act as the public `Client` and `ClientInterface` for the package that encapsulates how to adapt the (sementically correct) generated client into one that has go-style error return characteristics EG:
```golang
type Client struct {
  generatedClientInterface
}

var _ ClientInterface = &Client{}

type ClientInterface interface {
  PostJson(ctx context.Context, body PostJsonJSONRequestBody) (*SchemaObject, error)
}

// NewClient ...
func NewClient(server string, opts ...ClientOption) (*Client error) {
	genClient, err := newGeneratedClient(server, opts...)
	if err != nil {
		return nil, err
	}
	return &Client{genClient}, nil
}

// manually implement golang's `Error() string` interface so it can be passed as an `error`
func (e *Error) Error() string {
 return fmt.Sprintf("Error httpStatusCode: '%s', errorCode: '%s', message: '%s'", e.HttpStatusCode, e.ErrorCode, e.Message)
}

func (c *Client) PostJson(ctx context.Context, body PostJsonJSONRequestBody) (*SchemaObject, error) {
  resp, err := c.gen.PostJsonWithResponse(ctx, body)
  if err != nil {
    return nil, err
  }

  switch {
  case resp.JSON200 != nil:
    return resp.JSON200, nil
  case resp.JSONDefault != nil:
    return nil, resp.JSONDefault
  default:
    return nil, errors.New("blah")
  }
}
```
I understand why the second example/wrapper wouldn't be achievable through code generation from the spec because it requires some extra assumptions to be made about the API that can't be articulated with openapi easily, but I think supporting generating the client as private is enough to support the use case well


pseudo-su:
I'm keeping this as a draft PR for the moment, but I'd be interested in your thoughts @StevenACoffman

pseudo-su:
After sleeping on it I think it might be a good idea to combinee `Client` and `ClientWithResponses`. It seems like the only reason they're seperate might be that they were contributed separately

pseudo-su:
The `--make-private=client` CLI flag accepting an array of options doesn't make as much sense now i've removed the secondary client. keeping it as-is could still allow for future configuration of more things being private 🤷‍♂ ... open to suggestions of what it should be tho

internal/test/issues/issue-52/issue.gen.go
```diff
@@ -116,10 +116,12 @@ type Client struct {
 	RequestEditor RequestEditorFn
 }
 
+var _ ClientInterface = &Client{}
```
pseudo-su, line 119: This will make sure that the `Client` struct will always implement the `ClientInterface` otherwise it will be a build failure

pseudo-su:
@StevenACoffman I've added a PR branch here for me to keep track of how the different changes I'm proposing affect my `golang-service-template` repo in case you're interested. I still haven't added the example using the middleware for circuit-breakers/retry but I if I have time I might try and add that over the next couple of days.

https://github.com/pseudo-su/golang-service-template/pull/2

deepmap-marcinr:
ClientWithResponses and Client are indeed separate due to separate contributions, but they work at different layers of abstraction.

We use Client a lot, and skip the ClientWithResponses parsing, since we do custom parsing in our code (which I would like to merge back into this library, since it's swagger aware). 

I'd vote against merging them, to allow opting out of pre-canned parsing. Very generally, I've tried very hard to allow replacing our code at most any level, since it works in such a generic space that it's hard to come up with generic answers to people's needs.

pseudo-su:
@deepmap-marcinr 
> it's hard to come up with generic answers to people's needs
I totally agree and it's one of the reasons why I really enjoy using this library.

I'll try to make the case for why I don't think that combining the `ClientInterface` with the `ClientWithResponsesInterface` goes against that goal. (this has gotten a bit long, please don't interpret this as me trying to be pushy & complain about things, just trying to put forward my 2c 😄)

> I'd vote against merging them, to allow opting out of pre-canned parsing

I don't think that merging the two structs into one makes it any more difficult to choose to use the method that skips the pre-canned parsing.

Maybe for the benefit of providing an example, broadly speaking the change would be to go from this:

```golang
type Client struct{}

func (c *Client) MyOperation() {

}

type ClientWithResponses struct{
  baseClient *Client
}

func (c *ClientWithResponses) MyOperationWithResponse() {
  c.baseClient.MyOperation();
}
```

to this:
```golang
type Client struct {}

func (c *Client) MyOperation() {

}

func (c *Client) MyOperationWithResponse() {
  /* does some stuff and then delegates the actual http req/res to MyOperation() */
  c.MyOperation()
}
```

In the second (combined) example there's nothing stopping a consumer of the client from creating an instance of `Client` and then exclusively using `client.MyOperation()` (to skip pre-canned parsing) or `client.MyOperationWithResponse()` (to utilise the pre-canned parsing).

I would also assume that there's a good chance that if a consumer wanted to use pre-canned response parsing for some operations they might not want to use them for all endpoints, it seems strange to require users that might start by using the normal `Client` to create a whole new/different instance of the generated API client just to opt-in to the pre-canned response parsing.

I kind of view this as a little bit similar to what the aws sdk does when it has things like `dynamoDbClient.CreateBackup(input)` and `dynamoDbClient.CreateBackupWithContext(ctx, input)`. It would be somewhat unusual to require users to create a different instance of the `dynamoDbClient` in order to have `*WithContext` versions of all the actions of the client.
https://docs.aws.amazon.com/sdk-for-go/api/service/dynamodb/#pkg-types

At the end of the day, if a consumer wanted to explicitly prevent anyone from using any of the `*WithResponse` APIs they could also just use an interface that only exposes the subset of functionality they want to allow.

eg:
```golang
type MyClientInterface {
  MyOperation(/* etc */) /* etc */
}

var client MyClientInterface
client, err := pkg.NewClient(
  "https://myapi.com/v1",
  pkg.WithHTTPClient(httpClient),
)
client.MyOperation()
```
OR

```golang
type MyClientInterface {
  MyOperationWithResponse(/* etc */) /* etc */
}

var client MyClientInterface
client, err := pkg.NewClient(
  "https://myapi.com/v1",
  pkg.WithHTTPClient(httpClient),
)
client.MyOperationWithResponse()
```

deepmap-marcinr:
If you generate your API under a directory named `internal`, then Go will not allow others to import the API client, and you can wrap it yourself, without affecting the code generation. All these special cases add up and make things more complex. Would this fix the problem for you?

clarkmcc:
@deepmap-marcinr probably not, unfortunately. I'm not concerned about third-parties using the client, I'm concerned about other packages within the same project accessing the generated code. Everything in the internal directory is still accessible to other packages in the same project.

deepmap-marcinr:
There is already a way to change the Client name in the configuration under output options

https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/configuration.go#L84

It sounds like it needs to be extended to support ClientWithResponses

zak905:
@pseudo-su can you please clarify why the idea was abandoned ? It seems like useful option to have.

PR #169, opened 2020-04-07
Add ClientWithResponsesInterface a la ClientInterface, export response types

symbiont-james-sedgwick:
We want to use gomock/mockgen to mock out our generated client, but this required two steps
1. Create `ClientWithResponsesInterface`, an interface that `ClientWithResponses` (and a generated mock) conform to
2. Export the response types (which, since they're returned from exported functions, should be exported anyway)

Test plan: `make generate && make test`, successfully used w/ gomock in our internal code.


pseudo-su:
FYI I already have a PR for this [here](https://github.com/deepmap/oapi-codegen/pull/167)

deepmap-marcinr:
I'll happily merge this, but could you rebase and regenerate the .gen.go files? Thanks for doing this.

pseudo-su:
@deepmap-marcinr I think this PR can be closed because the same same/equivalent changes were made in https://github.com/deepmap/oapi-codegen/pull/167

PR #179, opened 2020-04-30
Fix Sprintf

Jleagle:


mitjaziv:
@Jleagle You should run `make generate` after the fix so `template.gen.go` is regenerated and add that to PR also.

Jleagle:
> @Jleagle You should run `make generate` after the fix so `template.gen.go` is regenerated and add that to PR also.

Done, thanks.

deepmap-marcinr:
Thanks.

PR #180, opened 2020-05-01
Adds support for openapi tags as OperationGroups

mitjaziv:
With this change we can group paths in different interfaces based on swagger tags, paths without tags are still rendered in `ServiceInterface`, Handlers are also grouped accordingly.

This change is back compatible.

Example with changed templates for `go-chi`:

```
{{ range $group := .Groups }}
{{ if gt ( len $group.Definitions ) 0 -}}

{{ if eq $group.Name "" -}}
{{ else }}
type {{ $group.Name }}Interface interface {
{{- end -}}

{{ range $group.Definitions -}}
// {{ .Summary | stripNewLines }} ({{ .Method }} {{ .Path }})
{{ .OperationId }}(w http.ResponseWriter, r *http.Request)
{{ end }}{{/* $group.Definitions */}}

}
{{ end }}{{/* if gt ( len $group.Definitions ) 0 */}}
{{ end }}{{/* range $group := .Groups */}}
```

Generated code (swagger comment where added manually):
```
// Swagger tag: - countries
type CountriesInterface interface { // Retrieves a list of countries (GET /countries)
	GetCountries(w http.ResponseWriter, r *http.Request)
	// Download list of countries in a pdf file (GET /countries/pdf)
	GetCountriesPdf(w http.ResponseWriter, r *http.Request)
	// Retrieves a single country (GET /countries/{code})
	GetCountriesCode(w http.ResponseWriter, r *http.Request)
}

// Without swagger tag.
	// Download list of countries in a pdf file (GET /countries/pdf)
	GetCities(w http.ResponseWriter, r *http.Request)
	// Retrieves a single country (GET /countries/{code})
	GetPostCodes(w http.ResponseWriter, r *http.Request)
}
```

I can also add go-chi templates with grouping, which are back compatible also.


pkg/codegen/templates/server-interface.tmpl
```diff
@@ -1,6 +1,6 @@
 // ServerInterface represents all server handlers.
 type ServerInterface interface {
```
deepmap-marcinr, line 2: How can this ever produce an interface named something other than ServerInterface?

deepmap-marcinr:
This is an interesting change I've not considered - as I thought of the tags as purely a documentation issue.

Your change breaks any complex, tagged existing servers (like mine). One way to remain compatible is to have ServerInterface embed all the other interfaces so it still contains everything.

But, it leads to a design question - how much should the structure of server code reflect the structure of the swagger doc?

PR #182, opened 2020-05-07
Update docs with more examples of registering http handlers

dannymidnight:
Based on https://github.com/deepmap/oapi-codegen/pull/110, https://github.com/deepmap/oapi-codegen/pull/124 it looks like there's some missing docs on how to setup the chi based server.

Open to feedback if docs aren't quite inline with style!

cc @m110 @aca @deepmap-marcinr


StevenACoffman:
This looks good to me!

OneCricketeer:
@dannymidnight Would you mind clarifying why net/http example code was added? Seems confusing based on the text above the section mentioning server generation when net/http generation doesn't exist #229

dannymidnight:
Yeah perhaps this could be made a little clearer in the docs for that section! ‘go-chi’ is 100% net/http compatible as opposed to the echo server.

The chi-server when first implemented had nothing specific to chi in the server interface, only used internally, making it more obvious consumers didn’t need to use chi if they didn’t want to.

OneCricketeer:
We started with `net/http` as a POC, and still want to use OpenAPI, and this repo popped-up on a random Google search, thus the question. 

We'd since migrated to Gin (although not too far down that route), so unfortunately Echo nor Chi are an option for us 😞

deepmap-marcinr:
Man, I'm so busy these days, but my intent is to decouple router from the
models themselves in a V2 API, but I'm not sure when I'll get to it. I want
to formalize the generation of code a little bit.

PR #189, opened 2020-05-28
fixed variable shadowing in the client.tmpl

dududko:
Fixes the variable showing in client.tmpl, which was detected with go vet.


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the templates have been completely rewritten multiple times since 2020. This PR's diff is against code that no longer exists. Any shadowing fix would need to be done fresh against the current codebase.

PR #192, opened 2020-06-09
add bson to descriptions

jamesraby:


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the `x-oapi-codegen-extra-tags` extension now provides a general-purpose mechanism for adding any arbitrary struct tags, including `bson`. It's fully documented with examples. Users can add `x-oapi-codegen-extra-tags: {bson: "field_name"}` to their specs.

PR #196, opened 2020-06-21
feat(middleware) Enable localisation of validation to routes

wolfeidau:
This change narrows the scope of the request validation middleware to
the routes which are registered, rather than the entire API.

This will enable you to have more than one generated api bould to a
listener, and localise authentication to these services.


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because route-scoped validation middleware is now fully supported. Each framework (Chi, Echo, Gorilla, Fiber, etc.) registers routes individually via `HandlerWithOptions()` which accepts per-API `ServerOptions` with a `Middlewares` field. Multiple generated APIs can coexist on a single listener with completely independent middleware chains.

PR #198, opened 2020-06-26
support BindQueryParameter handling map[string]string case

ponerine:


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the runtime package has been extracted to a separate module (`github.com/oapi-codegen/runtime`). Any fix for `BindQueryParameter` with deepObject-style query params would need to be done in that separate repo.

PR #200, opened 2020-07-01
fixed duplicate type names

mgurevin:
Generated code can't compile if the spec contains multiple definitions with the same name even they defined in different sections.

It may better idea to separate the generated code into different packages. 

But I need a solution to generate code from someone else's spec and it works for me.

E.g.:
```yaml
openapi: 3.0.1

info:
  title: ...
  version: 0.0.0

paths:
  /foo:
    post:
      parameters: 
        - $ref: '#/components/parameters/Bar'
      requestBody:
        $ref: '#/components/requestBodies/Bar'
      responses:
        200:
          $ref: '#/components/responses/Bar'

components:
  schemas:
    Bar:
      type: object
      properties:
        value:
          type: string
    Bar2:
      type: object
      properties:
        value:
          type: number
    BarParam:
      type: array
      items:
        type: integer
    BarParam2:
      type: array
      items:
        type: integer
 
  headers:
    Bar:
      schema:
        type: boolean

  parameters:
    Bar:
      name: Bar
      in: query
      schema:
        type: string

  requestBodies:
    Bar:
      content:
        application/json:
          schema:
            type: object
            properties:
              value:
                type: integer

  responses:
    Bar:
      description: ...
      headers:
        X-Bar:
          $ref: '#/components/headers/Bar'
      content:
        application/json:
          schema:
            type: object
            properties:
              value1:
                $ref: '#/components/schemas/Bar'
              value2:
                $ref: '#/components/schemas/Bar2'
              value3:
                $ref: '#/components/schemas/BarParam'
              value4:
                $ref: '#/components/schemas/BarParam2'

```


tooolbox:
On the surface, seems like this might fix #254 and #256 

Any thoughts @deepmap-marcinr ?

mgurevin:
> On the surface, seems like this might fix #254 and #256
> 
> Any thoughts @deepmap-marcinr ?

I agree with you, I had a similar issue with #254 and #256 before. So this PR was solved my problem. If this PR be merged, I'll get rid of maintaining my fork.

tooolbox:
I tried out your PR on a spec I have, and while it solved a name conflict between a Component and a RequestBody, there were other conflicts relating to Responses that prevented the generating code from compiling.

I think that there was some kind of object named FooResponse, and then `oapi-codegen` took a `Foo` object and added `Response` as a suffix in the templates, resulting in a conflict.  So, different issue I guess, but just as annoying.

deepmap-marcinr:
This is a really tricky problem, honestly, because these type conflicts
need to be propagated through `json` annotations, and the code makes other
assumptions about swagger name to go name mappings.

I was going to approach this in a very different way, but it's breaking.

Parameters, Responses, RequestBodies, etc, give you intent, but I'm not
sure if it's necessary to even generate types for them at all, unless the
$red into something in #/components/schemas.

It looks like your code basically no-ops when there are no conflicts, right?

mgurevin:
> It looks like your code basically no-ops when there are no conflicts, right?

correct.

guiguan:
maybe we separate them into different packages?

mromaszewicz:
I'm revisiting this ancient PR. It's a good idea, and it's implemented nicely. I've tested it and put it behind a configuration option to maintain old behavior.

PR #201, opened 2020-07-02
:pencil: Added missing feature to README

nilskuhn:
First of all, great work, I really appreciate people working together and sharing open source solutions!

Correct me if I am wrong, but:
Seems like defining parameters in global `#/components/parameters` section leads to a struct assumed for named global param, but struct itself (DocTypePathParam in example below) is not generated. Generated code doesn't compile.

openapi-yaml:
```
paths
  /document/{documentType}/{docTxid}:
      parameters:
        - $ref: "#/components/parameters/docTypePathParam"

<...>

components:
  parameters:
    docTypePathParam:
      in: path
      name: documentType
      <...>  
```
Generated go code:
```
	RetrieveDocuments(ctx echo.Context, documentType DocTypePathParam) error
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,6 +480,8 @@ need it. We've not yet implemented several things:
  knows how to parse them, but they're not part of OpenAPI 3.0, so we've left
  them out, as support is very complicated.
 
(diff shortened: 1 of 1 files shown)


deepmap-marcinr:
This was a bug introduced in an earlier PR. I've fixed it, so this works fine.

PR #203, opened 2020-07-02
unnecessary imports due to the comments

mgurevin:
Generated code contains unused imports when the spec contains that package names with a dot char in the type/endpoint descriptions.

Because we find the package names with a regex which one used in the generated code including comments.

This PR fixes that. Maybe can be found a better solution because this one only considers line comments. Currently, codegen only yields line comments, no block comments.

E.g.;

```yaml
openapi: 3.0.2

info:
  title: ...
  version: 0.0.0

paths:
  /foo:
    get:
      parameters:
        - name: Foo
          in: header
          description: base64. bytes. chi. context. echo. errors. fmt. gzip. http. io. ioutil. json. openapi3.
          schema:
            type: string
          required: false
        - name: Bar
          in: header
          description: openapi_types. path. runtime. strings. time.Duration time.Time url. xml. yaml.
          schema:
            type: string
          required: false
      description: ...
      responses:
        200:
          description: ...
          content:
            application/json:
              schema:
                type: string

```


deepmap-marcinr:
I'm happy to merge this, thanks for doing it. Could you resolve conflicts please?

mgurevin:
> I'm happy to merge this, thanks for doing it. Could you resolve conflicts please?

sure, it's done.

Issue #195, opened 2020-06-19
use $ref to point to pre-existing types and being able to import them


dududko:
I need to be able to reference in openapi to types, that pre-exist in some packages and I want to be able to import them in the generated code.

It is a very handy feature which I really need in my project. Without it I have to merge multiple openapi specs and write manually (or generate) type converters.

This feature is implemented in vanilla openapi-generator. They claim that it is posible to import existing types using option `--import-mappings`.
https://openapi-generator.tech/docs/usage/#target-external-models
Unfortunately this option seem to work proper only in java, I can confirm that it does not work for go and ts, and I found evidence that it also does not work for python and Kotlin. Also this feature lacks the examples. Nevertheless, I managed to implement this feature for `ts` generator by doing post-processing of generated files. It is important to point that I could achieve this only because vanilla openapi has option `--skip-validate-spec`.

I tried to find a solution for `oapi-codegen` to be able to introduce this feature. Unfortunately I almost instantly failed, because `oapi-codegen` always validates the spec. To be more precise, github.com/getkin/kin-openapi validates the spec when it parses the file. It means that to implement this feature, I will have to somehow generate a valid spec with valid imports which I later could modify by knowing which structs are external. 

I would like to ask the community if anyone else need this feature ? Do you think that it should be implemented for deepmap/oapi ?

dududko:
Here is a small example
```
➜ cat oapi.yaml
components:
  schemas:
    Pet:
      type: string
    PetOwner:
      $ref: "PetOwner"
```

`oapi-codegen -o api.go --package petstore -generate=types,skip-prune` --import-mappings=PetOwner=github.com/dududko/pet-project/types

```
➜ cat api.go
package petstore

import "github.com/dududko/pet-project/types"

type (
    Pet string

    PetOwner = types.PetOwner
)
```

dududko:
Or the same with a more distinct format:
```
components:
  schemas:
    Pet:
      type: string
    PetOwner:
      $ref: "github.com/dududko/pet-project/types:PetOwner"
```

Issue #42, opened 2019-07-29
Adding support for external refs


dannymidnight:
Hey thanks for writing this library, I know it's still early days for this library but I was wondering if there were any plans to add support for external refs?

e.g. 

```
`$ref: "../resource.yml/#components/schemas/Hello"`
```

I had a play around enabling `SwaggerLoader.IsExternalRefsAllowed` but ran into [this](https://github.com/deepmap/oapi-codegen/blob/2177b2dc5487b76933051ccab327922d6d5ab204/pkg/codegen/utils.go#L199)

> Only local document components are supported

Happy to contribute a patch but before doing so was wondering if there was any reason in particular for not supporting this feature.

deepmap-marcinr:
We use kin-openapi to parse and resolve swagger specs, and it doesn't seem to support that.

dannymidnight:
> We use kin-openapi to parse and resolve swagger specs, and it doesn't seem to support that.

Riiight, I was under the impression `SwaggerLoader.IsExternalRefsAllowed` flattened out all the refs but looking through the source that's not the case :D

weberr13:
I have a proposed solution to at least some of the external refs issues here:  

https://github.com/deepmap/oapi-codegen/pull/48

The primary issue is that external refs should reside in a different package.  The codegen is designed to write a single file so this cannot be done with one pass.  This proposal allows the user to specify the package path for any external refs and use a previous generate stage to generate that package.  This is written more for url based refs but this is an approach that may get the ball rolling.

j--wong:
Hi @dannymidnight @deepmap-marcinr @weberr13 I believe external refs are still not supported? 
As I am getting below error during code gen...

```
error loading swagger spec: Encountered non-allowed external reference
```

I can see both this issue and linked PR are in `Closed` status.  Has the project decided not to add this feature?

We are working with amazon eventbridge and dealing with quite a bit schema files.  The external refs would allow us to not have to repeat event structure in every file.  External refs support would be super handy for us and for many others.

Is there any plans for this to be added or any ideas on workarounds for projects with many schema files that need to reference one another?  thanks!

deepmap-marcinr:
[...]

PR #204, opened 2020-07-06
Added support for external references through the --import-mapping option

dududko:
In this PR I added a support for external references. The support is added with the use of `--import-mapping` option - a json that specifies the mapping from external reference to the golang package path. This will allow users to specify two openapi specs `A` and `B`, where `A` references `B`; generate golang package for spec `B` and then generate a valid golang package for spec `A`, which contains imports of a package `B` and uses it's types. 

You can see the example here `internal/test/externalref/`. 

Since the external packages that are imported can have the same package name, I sort imported package paths and them give them a uniques alias `externalRef%d`, where `%d` is a serial number.

I chose to call the option `--import-mapping` due to the similar option in vanilla openapi generator. 
https://openapi-generator.tech/docs/customization/#bringing-your-own-models

Feel free to review and comment this PR.

P.S. I also tried to implement the embedding of external references instead of importing them, but it introduces too many changes... It will require the major refactoring to have this feature, this is why I did not complete it.


dududko:
Implements #42, #195 

The solution is similar to https://github.com/deepmap/oapi-codegen/pull/48

lukekhamilton:
Hey @dududko thanks for the PR, code looks good to me. I will alsov test the feature out over this week if I can find the space. 

@deepmap-marcinr be great to get some feedback from you on this one? This is a much needed feature for many.

deepmap-marcinr:
This is a very interesting way of doing it!

I like the notion of import mapping.

I'm going to merge this and change it around a little bit, for example, import maping being JSON is kinda tedious to type in on command lines, so I might just make it key1:value1,key2:value2 with some provision for escaping commas.

dududko:
I am glad to hear that you accepted this new feature. 
I think it is important to tell you that recently I found out that if there are external references in openapi specification, then the generated `swaggerSpec` variable is not valid. As a result, is is not possible to use the `middleware.OapiRequestValidator`.

So of we will add the following lines of code to the test:
```
    swagger, err := api.GetSwagger()
    if err != nil {
        panic(err)
    }

    data, err := swagger.MarshalJSON()
    if err != nil {
        panic(err)
    }

    fmt.Println(string(data))

```
`GetSwagger` function will work, but `MarshalJSON` will panic
```
panic: error loading Swagger: Encountered non-allowed external reference: 'bitbucket.org/projectplato/rules_openapi/examples/external_ref/packageB/spec.yaml#/components/schemas/ObjectB'
```

I think is should be fixed in `github.com/getkin/kin-openapi`, otherwise the embedding logic need to be implemented in this repo.

deepmap-marcinr:
Thanks. That stinks about the validator. I might have to write a function
to fetch external dependencies and merge them together. I already have some
rather tricky code written which can merge two specs and resolve type
collisions, might as well use it.

dududko:
> That stinks about the validator
To be honest I hate the fact, that there is no single struct in go that represents the openapi container as is with all optional/required attributes. Instead we are working with internal representation that `getkin` is using. I want to just read the file, unmarshal it's content and then decide if it should be validated, if all dependent specs should be parsed and etc.

Hope one day someone will refactor all this code.

PR #206, opened 2020-07-07
Generate plain slice members for structs instead slice references

viktorasm:
Breaking change: generate struct members of slice type as `m []Type` instead of `m *[]Type`. Having pointer-to-slice is near-pointless when slice is a member of a struct.

This suggested behaviour is consistent with how e.g. GRPC generator behaves.

### DIFF
--- a/pkg/codegen/codegen_test.go
+++ b/pkg/codegen/codegen_test.go
@@ -169,6 +169,9 @@ type GetTestByNameResponse struct {
 	problems, err := linter.Lint("test.gen.go", []byte(code))
 	assert.NoError(t, err)
 	assert.Len(t, problems, 0)
+
+	// Make sure that slices are generated without references
+	assert.Contains(t, code, "Cases []TestCase")
 }
 
 const testOpenAPIDefinition = `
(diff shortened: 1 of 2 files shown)


ulfurinn:
I agree, I don't see the benefit of adding a pointer to a type that is already intrinsically nullable and works naturally with `omitempty`.

deepmap-marcinr:
This indirect was never intentional, it's just a special case which wasn't flattened.

Doing this right now is going to break existing code when people regenerate.

I've got a v2 branch going for such breaking changes. I think we should skip the indirect for slices, maps, and anything else which has distinct nil and empty states. It will make the code nicer to work with.

vnarek:
Hello, do you think we could create a compatibility option for this or is it going to be hard to implement? I can probably look into this If people would like this.

mromaszewicz:
This PR is ancient, and sorry to let it dangle for so long.

since it was filed, two options are available in the configuration (`prefer-skip-optional-pointer` and `prefer-skip-optional-pointer-on-container-types`) which optionally enable just this behavior.

We can't default it on until a major version update.

PR #208, opened 2020-07-15
feature: --exclude-schemas option for excluding schemas from type generation

dududko:
### New feature:

introduce `--exclude-schemas` option to specify the list of schemas that need to be excluded from type generation. 

This will allow to have both generated and hand-written types or interfaces in the same package, and have valid references in both source code, generated code and the openapi spec.


dududko:
@deepmap-marcinr could you please have a look on this PR ?

PR #214, opened 2020-08-03
unsigned int fields

theromis:
Using uint64 as my postgresql messages IDs, can't find other option as just add it to this great generator.

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -231,8 +231,12 @@ func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
 			// We default to int if format doesn't ask for something else.
 			if f == "int64" {
 				outSchema.GoType = "int64"
+			} else if f == "uint64" {
+				outSchema.GoType = "uint64"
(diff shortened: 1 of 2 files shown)


pkg/codegen/schema.go
```diff
@@ -231,8 +231,12 @@ func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
 			// We default to int if format doesn't ask for something else.
 			if f == "int64" {
 				outSchema.GoType = "int64"
+			} else if f == "uint64" {
```
deepmap-marcinr, line 234: This is not valid swagger. uint64 and uint32 are not supported, see "Data Types" here:
https://swagger.io/specification/

I wonder if we could do this with x-go-type

theromis: @deepmap-marcinr it will be great if we can do it any how, because my current `go-pg` schema `id` are `uint64`

deepmap-marcinr: Well, it makes us more permissive, we still work with vanilla swagger, so ok.

PR #224, opened 2020-08-25
Fix generic status code responses

natsukagami:
Before, generic responses (like `4XX`) generate un-compilable code.

This fixes the generated code so that it handles generic responses more gracefully.

---
This change is [Reviewable](https://reviewable.io/reviews/deepmap/oapi-codegen/224)


dahu33:
Ref: https://swagger.io/specification/#responsesObject

deepmap-marcinr:
thank you for fixing this.

PR #226, opened 2020-09-03
Replace gofmt by goimports and move the imports the template

dahu33:
This is great project with a lot of potential! In this PR I've tried to add my little contribution:

1) Replaced `gofmt` by `goimports`: `goimports` does everything that `gofmt` do but it also groups and corrects imported packages. This will make the generated code nicer to read and reduce the probability of having a missing/incorrect import statement.

2) Removed the imports list (and the related code) from the source code and moved them to the template. `goimports` will automatically "clean" the import section. This will make the templates more easily customizable (devs can add imports) without the need of modifying the import list in the source code.


deepmap-marcinr:
Nice! Thanks.

PR #228, opened 2020-09-09
Allow multiple RequestEditorFns in Client and Request Methods

natsukagami:
I usually find myself having to combine multiple editors into one, or change the client in order to customize a request with e.g. special headers.

This changes that with 
- Allowing multiple `RequestEditorFn`s in the Client, with each `WithRequestEditorFn` call appending one to the list.
- Allowing `RequestEditorFn`s as a (spread) parameter in every Client method. They are run after the default ones.

---
This change is [Reviewable](https://reviewable.io/reviews/deepmap/oapi-codegen/228)

### DIFF
--- a/pkg/codegen/templates/client.tmpl
+++ b/pkg/codegen/templates/client.tmpl
@@ -18,9 +18,9 @@ type Client struct {
 	// customized settings, such as certificate chains.
 	Client HttpRequestDoer
 
-	// A callback for modifying requests which are generated before sending over
+	// A list of callbacks for modifying requests which are generated before sending over
 	// the network.
-	RequestEditor RequestEditorFn
+	RequestEditors []RequestEditorFn
 }
 
 // ClientOption allows setting custom parameters during construction
@@ -62,7 +62,7 @@ func WithHTTPClient(doer HttpRequestDoer) ClientOption {
 // called right before sending the request. This can be used to mutate the request.
 func WithRequestEditorFn(fn RequestEditorFn) ClientOption {
 	return func(c *Client) error {
-		c.RequestEditor = fn
+		c.RequestEditors = append(c.RequestEditors, fn)
 		return nil
 	}
 }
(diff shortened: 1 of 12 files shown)


mitar:
This would be super useful. I just now have to complicate my code because I want to use different access tokens for some requests. With this PR I could just pass `RequestEditorFn` to the call itself. Super useful.

dahu33:
@deepmap-marcinr any chance we could get this in? It's a non-breaking change and it's been proven useful.

pkg/codegen/templates/client.tmpl
```diff
@@ -18,9 +18,9 @@ type Client struct {
 	// customized settings, such as certificate chains.
 	Client HttpRequestDoer
 
-	// A callback for modifying requests which are generated before sending over
+	// A list of callbacks for modifying requests which are generated before sending over
 	// the network.
-	RequestEditor RequestEditorFn
+	RequestEditors []RequestEditorFn
```
deepmap-marcinr, line 23: This is a breaking change. Anyone accessing the previous RequestEditor will get a compile error now.

We don't need a list of RequestEditors on the client, could you leave it as a single RequestEditor?

Your per-handler request editors supercede this global per-client editor.

deepmap-marcinr (changes requested):

deepmap-marcinr:
I've made one small comment in there to avoid breaking backward
compatibility, and I'll happily merge and cut a new release once you submit
this.

pkg/codegen/templates/client.tmpl
mitar: > We don't need a list of RequestEditors on the client, could you leave it as a single RequestEditor?

But how would then one register multiple RequestEditors with a client?

I mean, modifying `RequestEditors` directly should be a bad pattern anyway, no? You have to modify it through constructor options.

But if this is really the problem, we could have `RequestEditor` and `RequestEditors`, where first the `RequestEditor` is run an then new `RequestEditors` in order.

deepmap-marcinr: I'm guessing the breakage from this will be minimal, I will merge as-is.

The reason that I originally had a single request editor is that you yourself can call a bunch from within it, so it seemed redundant.

Your code never uses that editor array on the client, as you use the list specified on the handler, and only of present, do you append. So, the change from 1 to many doesn't really gain much.

PR #232, opened 2020-09-21
Type aliasing

natsukagami:
Some types are unnecessary wrapped, for example 
```go
type AddPetJSONRequestBody Pet
```
which lengthens the code needed to create and handle such types.

I added an option, called `-alias-types`, that changes all types satisfying:
```go
type A struct { /* ... */ } // unchanged
type B string // native type wrappers are unchanged, they are useful for e.g. enums
type C = A  // Direct wrappers of another defined type is aliased
type D = []A // Array of referenced types are aliased
type E []int // Array of non-referenced types are not aliased
```

---
This change is [Reviewable](https://reviewable.io/reviews/deepmap/oapi-codegen/232)

### DIFF
--- a/pkg/codegen/templates/param-types.tmpl
+++ b/pkg/codegen/templates/param-types.tmpl
@@ -1,6 +1,6 @@
 {{range .}}{{$opid := .OperationId}}
 {{range .TypeDefinitions}}
 // {{.TypeName}} defines parameters for {{$opid}}.
-type {{.TypeName}} {{.Schema.TypeDecl}}
+type {{.TypeName}} {{if and (opts.AliasTypes) (.CanAlias)}}={{end}} {{.Schema.TypeDecl}}
 {{end}}
 {{end}}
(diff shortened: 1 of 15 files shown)


pkg/codegen/templates/param-types.tmpl
```diff
@@ -1,6 +1,6 @@
 {{range .}}{{$opid := .OperationId}}
 {{range .TypeDefinitions}}
 // {{.TypeName}} defines parameters for {{$opid}}.
-type {{.TypeName}} {{.Schema.TypeDecl}}
+type {{.TypeName}} {{if and (opts.AliasTypes) (.CanAlias)}}={{end}} {{.Schema.TypeDecl}}
```
deepmap-marcinr, line 4: if opts.AliasTypes=faalse or .CanAlias=false, you will generate an empty type alias, no?

`type TypeName`

natsukagami: If one of them is false, the type will be 
`type TypeName .Schema.TypeDecl`
else it will be 
`type TypeName = .Schema.TypeDecl`

PR #234, opened 2020-09-25
Add Buffalo generator and Swagger validator

egonz:
- Added Buffalo templates based on the Echo implementation.
- Added a new Swagger validator for use with Buffalo.
- Updated README with a buffalo example.

Tested using the petstore-expanded.yaml Swagger definition. Also incorporated the generated code and validator into a Buffalo HelloWorld project. Handlers were called properly, and confirmed the validator was properly validating JSON input.


egonz:
Tests should be fixed now. Some of the gen files were missing the types.

jamietanna:
Closing as we didn't get to this in time, and it may not still be necessary - feel free to reopen if it's still of use!

PR #237, opened 2020-10-12
Validation tags

dududko:
### New Feature: validation tags

This PR adds generation of validation tags, which can be used by [validator](https://github.com/go-playground/validator) - a framework for validation. This approach is [recommended](https://echo.labstack.com/guide/request) by `echo` web framework.

It is the alternative solution to request validation by `github.com/getkin/kin-openapi`, which uses the source openapi spec to create Swagger container and use it to validate incoming requests.

### Example
```
# input spec
components:
  schemas:
    StructA:
      required:
        - requiredString
        - color
        - rangeInt
      properties:
        requiredString:
          type: string
        rangeInt:
          type: integer
          format: int64
          minimum: 2
          maximum: 42
        color:
          type: string
          enum:
            - black
            - white
```

```
// StructA defines model for StructA.
type StructA struct {
	RangeInt       int64  `json:"rangeInt,omitempty" validate:"required,gte=2,lte=42"`
	RequiredString string `json:"requiredString" validate:"required"`
	Color          Color  `json:"color" validate:"required,oneof=black white"`
}
```

```
a := types.StructA{
	RangeInt: 99,
	Color: "orange",
}

// stdout: map[StructA.Color:Color must be one of [black white] StructA.RangeInt:RangeInt must be 42 or less StructA.RequiredString:RequiredString is a required field]
fmt.Println(validate.Struct(a).(validator.ValidationErrors).Translate(translator))
```
you can find more examples in this file `internal/test/validator/validator_test.go`

Credits: @madesst, @dududko


dududko:
There is a related PR which uses different validation framework - `ozzo-validation` https://github.com/deepmap/oapi-codegen/pull/225.

chemidy:
Maybe you can add an option `validator` to -generate like chi-server, so we can have another kind of validator like `ozzo-validation` !

dududko:
FYI: I decided to change this solution in favor of vanilla approach for request/response validation. 
Was not able to use it because the external references were not fully supported in the embedded spec. 
I fixed this in #320

dududko:
Closed in favor of #327

PR #238, opened 2020-10-13
Feature/response helper

dududko:
### New Feature: response helper for echo server

This PR introduces the <OperationId>Context struct, which helps to send specified responses. 

### Example

```
paths:
  /issues/127:
    get:
      operationId: Issue127
      responses:
        200:
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/GenericObject"
            text/markdown:
              schema:
                $ref: "#/components/schemas/GenericObject"
            text/yaml:
              schema:
                $ref: "#/components/schemas/GenericObject"
            application/xml:
              schema:
                $ref: "#/components/schemas/GenericObject"
        default:
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/GenericObject"
            text/markdown:
              schema:
                $ref: "#/components/schemas/GenericObject"
```

```
	// (GET /issues/127)
	Issue127(ctx Issue127Context) error
}

type Issue127Context struct {
	echo.Context
}

func (c *Issue127Context) JSON200(resp GenericObject) error {
	err = c.Validate(resp)
	if err != nil {
		return err
	}
	return c.JSON(200, resp)
}

func (c *Issue127Context) XML200(resp GenericObject) error {
	err = c.Validate(resp)
	if err != nil {
		return err
	}
	return c.XML(200, resp)
}

func (c *Issue127Context) YAML200(resp GenericObject) error {
	err = c.Validate(resp)
	if err != nil {
		return err
	}
	var out []byte
	out, err = yaml.Marshal(resp)
	if err != nil {
		return err
	}
	return c.Blob(200, "text/yaml", out)
}
```


klowdo:
@dududko Love this PR! but i think we need to make this an opt in somw how

dududko:
@klowdo what do you mean by opt ? This feature does not break the backward compatibility since it adds a new interface for server so it is actually optional to use.

klowdo:
@dududko that once this PR is merged the contract changes. Yes it's is backward compatible but you'll have to change all your implementation due to breaking change. So i'm thinking an option that you can pass in to generate like `--with-reponse-helper` or on the path x-go-reponse-helper to make incremental changes

klowdo:
@dududko any update on this? I could try to help you?

dududko:
@klowdo another option it to add a new interface for server - the same way it is done for go client: `ClientInterface` and `ClientWithResponses`.

Unfortunately in the upcoming months I will not have time to work on this PR so feel free to finish it yourself.
The only thing I can help you with is to actualize this PR. A month ago I did clean up the code in the public [fork](https://github.com/Project-Plato/oapi-codegen/commits/v1.9.0-plato.1) that my company maintains and rebased it on the v1.9.0 release. [This](https://github.com/Project-Plato/oapi-codegen/commit/3e2d0293403004b2a35b7731202ac8c79b7fb0c9) the commit that you need to cherry-pick.

klowdo:
@dududko  thanks! i might play around with anotherinterface .. but that means that we need another template file that need to be maintained... anyway added this #520 that you could have a look at :)

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the strict-server pattern, introduced in June 2022, completely supersedes this PR. It provides typed request/response objects with a visitor pattern, automatic Content-Type headers, status codes, and multi-framework support.

PR #242, opened 2020-10-21
fix: fixing time styling

vlastahajek:
This fixes the serialization of time values. 

If a  param with type _time_ has a value with positive timezone, e.g.  `2020-01-01T22:00:00+02:00`, it is unescaped by `url.ParseQuery` in the generated code into the `2020-01-01T22:00:00 02:00` string (missing the + char) and then wrongly escaped with %20 at that place later by `queryValues.Encode() `.

### DIFF
--- a/pkg/runtime/styleparam.go
+++ b/pkg/runtime/styleparam.go
@@ -16,11 +16,14 @@ package runtime
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"reflect"
 	"sort"
 	"strconv"
 	"strings"
 	"time"
+
+	errorsw "github.com/pkg/errors"
 )
 
 // Given an input value, such as a primitive type, array or object, turn it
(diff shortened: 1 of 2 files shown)


pkg/runtime/styleparam.go
```diff
@@ -149,7 +151,11 @@ func marshalTimeValue(value interface{}) (string, bool) {
 
 func styleStruct(style string, explode bool, paramName string, value interface{}) (string, error) {
 	if timeVal, ok := marshalTimeValue(value); ok {
-		return stylePrimitive(style, explode, paramName, timeVal)
+		styledVal, err := stylePrimitive(style, explode, paramName, url.QueryEscape(timeVal))
+		if err != nil {
+			return "", errorsw.Wrap(err, "failed to style time")
```
pablocrivella, line 156: I think here you cold just use `fmt.Errorf("failed to style time: %w", err)` instead and avoid the dependency on the errors pkg.

deepmap-marcinr: We've already got the pkg/errors dependency in other places, it's a great package. Instead of importing it and errors, please just import pkg/errors which is compatible with errors, and using errors.Wrap() is a great idea. I'm going to clean up and use that e verywhere.

deepmap-marcinr (changes requested):

PR #250, opened 2020-11-05
make task for release assets

myles-mcdonnell:
Hi

Great work, I got up and running this AM in minutes, thank you!

A pattern I like to follow is to add binaries to GitHub releases (e.g. https://github.com/myles-mcdonnell/oapi-codegen/releases/tag/v1.4.0), I can then pin my toolchain (docker builds etc) to specific versions that I pull from the GH release.  You probably don't want to do this or you would have already but wanted to share what I did this morning on the off chance that you find it useful.

Thanks
Myles

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea
 .DS_Store
 .vscode/
+.bin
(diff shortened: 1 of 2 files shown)


jamietanna:
Thanks for this (and sorry it took a while to get back to you), this isn't something we're looking for right now.

PR #258, opened 2020-11-30
Feature: Expose constants for the context keys that pass expected security scopes

maclav3:
## What changed

The context keys used when adding scopes of a security scheme are exported as constants, one per each security scheme.
This should enable depedent code to read the scopes more reliably via:

```go
scopes := ctx.Value(types.BearerTokenScopes)
```

instead of the hardcoded:

```go
scopes := ctx.Value("bearerToken.Scopes")
```

## Rationale

An effective way to check permissions for API operations is to write a middleware that takes the expected security scopes for the operation, and confronts them with the scopes that are present for the request.

Exposing the context key as a `const` increases safety by enabling compile-time checking, prevents typos and unexpected errors.

## Notes

I made an attempt to expose more constants, like operation IDs (which could be passed as context values) and paths for each operation: https://github.com/deepmap/oapi-codegen/compare/v1.4.1...maclav3:expose-constants

However, I decided that it was too opinionated, and I don't really have good enough use cases for that.
This adds only the context key for security scopes, but I retained the constants templating mechanism, because it is extensible in case there's a need to expose more constants in the future.

I thought to make the context key a private type, like recommended in https://golang.org/pkg/context/#WithValue:
```go
type contextKey string
```
> The provided key must be comparable and should not be of type string or any other built-in type to avoid collisions between packages using context. Users of WithValue should define their own types for keys

but this doesn't work with `echo.Context.Set`, which expects a `string`.
Made no sense to generate different types for `server` and `chi-server`, so I retained the `string`.


pkg/codegen/templates/constants.tmpl
```diff
@@ -0,0 +1,7 @@
+{{- if gt (len .SecuritySchemeProviderNames) 0 }}
+const (
+{{range $ProviderName := .SecuritySchemeProviderNames}}
+    {{- $ProviderName | ucFirst}}Scopes = "{{$ProviderName}}.Scopes"
```
maclav3, line 4: This is useful when generating a server, less so with a client
I didn't want to overcomplicate the PR by generating this only when `server|chi-server` is chosen, but maybe I should?

pkg/codegen/codegen.go
```diff
@@ -298,6 +309,38 @@ func GenerateTypeDefinitions(t *template.Template, swagger *openapi3.Swagger, op
 	return typeDefinitions, nil
 }
 
+// Generates operation ids, context keys, paths, etc. to be exported as constants
```
deepmap-marcinr, line 312: This comment is out of date with the code.

pkg/codegen/codegen.go
```diff
@@ -298,6 +309,38 @@ func GenerateTypeDefinitions(t *template.Template, swagger *openapi3.Swagger, op
...
+func GenerateConstants(t *template.Template, ops []OperationDefinition) (string, error) {
+	var buf bytes.Buffer
+	w := bufio.NewWriter(&buf)
+
+	constants := Constants{
+		SecuritySchemeProviderNames: []string{},
+	}
+
+	providerNames := map[string]struct{}{}
```
deepmap-marcinr, line 321: We've been using map[string]bool for a map of key presence. It allows for `if foo[x]`. It's nice to stick to one style.

deepmap-marcinr (changes requested):
Please run "make generate" since your change requires regenerating all generated code. This is why the precommit failed.

mitar:
This broke our generated code. It ends up generating:

```
const (
Access-tokenScopes = "access-token.Scopes"

)
```

Relevant part of our spec (in YAML):

```
  securitySchemes:
    access-token:
      type: http
      scheme: bearer
      bearerFormat: |
        JWT-format access token.

security:
  - access-token: []
```

mitar:
It seems there is no way to disable generating those constants.

deepmap-marcinr:
Thanks for reporting this. I'll get it fixed asap. These things are so fragile!

deepmap-marcinr:
Try v1.5.1, it should fix your issue while preserving the enums.

I really need to refactor this code to sanitize names better...

mitar:
Thanks! It works (compiles). The output is pretty ugly though:

```
Access_tokenScopes = "access_token.Scopes"
```

deepmap-marcinr:
Yes, I agree, but at least it works now. Step 2 is to make it less ugly
later.

PR #261, opened 2020-12-08
Mark embedded structs for inlining in yaml

mitar:
Configures yaml to inline the embedded struct, see: https://github.com/go-yaml/yaml/blob/7649d4548cb53a614db133b2a8ac1f31859dda8c/yaml.go#L183-L186

### DIFF
--- a/examples/petstore-expanded/chi/api/petstore.gen.go
+++ b/examples/petstore-expanded/chi/api/petstore.gen.go
@@ -39,7 +39,7 @@ type NewPet struct {
 // Pet defines model for Pet.
 type Pet struct {
 	// Embedded struct due to allOf(#/components/schemas/NewPet)
-	NewPet
+	NewPet `yaml:",inline"`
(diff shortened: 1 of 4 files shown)


deepmap-marcinr:
Interesting. Since the structs in this case are embedded already, the YAML output effectively looked the same already. Are you seeing a behavior difference due to this change?

mitar:
> the YAML output effectively looked the same already.

Not to me? Without this I get YAML where embedded struct is nested, with this change, the embedded struct is inlined (all fields put at the top-level).

The problem is that I am using YAML to have some fixtures for the API. So if I want to load YAML and have it "flat" this is necessary.

deepmap-marcinr:
Oh, yeah, you're right. I was thinking of the JSON behavior.

mitar:
Thanks!

mitar:
Any plans to tag a release?

deepmap-marcinr:
I will. I'm waiting on one more change to go in.

PR #262, opened 2020-12-12
New Bind method to allow more custom basic types

derekbassett:
This introduces the Bind Method to allow new custom types to be defined for query parameters without introducing a new dependency between runtime and types.

### DIFF
--- a/pkg/runtime/bindstring.go
+++ b/pkg/runtime/bindstring.go
@@ -76,6 +76,11 @@ func BindStringToObject(src string, dst interface{}) error {
 			v.SetBool(val)
 		}
 	case reflect.Struct:
+		// if this is not of type Time or of type Date look to see if this is of type Binder.
+		if dstType, ok := dst.(Binder); ok {
+			return dstType.Bind(src)
+		}
+
 		if t.ConvertibleTo(reflect.TypeOf(time.Time{})) {
 			// Don't fail on empty string.
 			if src == "" {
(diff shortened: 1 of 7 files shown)


pkg/runtime/bindstring.go
```diff
@@ -101,27 +99,9 @@ func BindStringToObject(src string, dst interface{}) error {
 			v.Set(reflect.ValueOf(parsedTime))
 			return nil
 		}
-
-		if t.ConvertibleTo(reflect.TypeOf(types.Date{})) {
```
deepmap-marcinr, line 105: This code currently works on anything that's convertible to time or date, while your new code, which is a great idea, by the way, only works on things which implement Binder, so we're potentially breaking existing code here.

Could you leave in the old code as well as adding your new path? Check for Binder first, but fall back to the convertible test.

derekbassett:
Yes absolutely.

pkg/runtime/bindstring.go
derekbassett: Okay, I have updated the PR with this. Please re-review and let me know if it meets your expectations.

pkg/runtime/deepobject.go
```diff
@@ -3,6 +3,7 @@ package runtime
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/deepmap/oapi-codegen/pkg/types"
```
mitar, line 6: This looks like a wrong place for the import?

PR #265, opened 2020-12-16
feat: Allow response unmarshalling switch customization

vlastahajek:
This PR adds switch wrapper into the template allowing adding a custom switch clause to the custom template. 
Example:
```
 switch {
    {{genResponseUnmarshal .}}
    // Fallback for unexpected error
    default:
        if rsp.StatusCode > 299 {
            return nil, &types.Error{StatusCode: rsp.StatusCode, Message: rsp.Status}
        }
    }
```


vlastahajek:
Maybe, it could better to generate default error when no HTTP error case matches by default.
Comments?

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the user-templates system now allows complete template overrides via configuration. Users can replace any built-in template (including `client-with-responses.tmpl`) with a custom one. This PR's goal is achievable through the existing template override mechanism.

PR #268, opened 2021-01-03
Error Message for Non-String Enum

tooolbox:
I just ran across a spec that had an enum containing the values `[Y, N]` which were meant as strings but interpreted as boolean, causing a panic during generation with the message `interface conversion: interface {} is bool, not string` at `schema.go:272`.

This tweak adds a helpful error message that lets you locate which part of the spec broke the generation and fix it, such as changing to `['Y', 'N']`.

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -268,8 +268,12 @@ func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
 			outSchema.GoType = "bool"
 		case "string":
 			enumValues := make([]string, len(schema.Enum))
+			var ok bool
 			for i, enumValue := range schema.Enum {
-				enumValues[i] = enumValue.(string)
+				enumValues[i], ok = enumValue.(string)
+				if !ok {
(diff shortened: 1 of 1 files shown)


tooolbox:
Hm, 32 open PRs, and I might need to open more shortly.  Anyone reviewing contributions?  @deepmap-marcinr ?

deepmap-marcinr:
Thanks for your PR. I'm quite busy and don't always get to PR's
immediately, however, we have a whole bunch of backlogged ones that I don't
know what to do about. They'll break existing users if I submit them. I'm
thinking of making a V2 package which incorporates a lot of breaking
changes but also allows for more future compatibility.

tooolbox:
Hey @deepmap-marcinr thanks for the merge!  And thanks for clarifying, I understand now why the PRs are backlogged.

The V2 package sounds like a good idea.  In the interim, you could also make a "breaking" label and go through and mark all the breaking-change PRs with that tag&mdash;just a suggestion, so you can perceive the makeup of the backlog more clearly and visitors will know why they're not merged.

PR #270, opened 2021-01-05
Fix for case when only cookie params are exists for operation

derfenix:
In this case the `err` variable, declared above, is unused and causes compilation error


deepmap-marcinr:
thanks.

Issue #279, opened 2021-01-13
Query parameter that has reserved char will result to invalid query string


wangyoucao577:
### Sample OpenAPI:        
```
openapi: 3.0.3

info:
  title: api
  description: sample api
  version: "0.0.0"

paths:
  /endpoint:
    get:
      operationId: getOperation
      parameters: 
        - in: query
          name: indexes
          required: false
          description: '{index};{index};{index}...'
          schema:
            type: string
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: string
```

### generate go code 
```bash
$ oapi-codegen  -generate types,client -package api ./api.yaml > ./api/api.gen.go
```

### main
(`require github.com/deepmap/oapi-codegen v1.4.2`)

```go
func main() {

	indexes := "1;2"
	params := api.GetOperationParams{Indexes: &indexes}

	r, _ := api.NewGetOperationRequest("test_server", &params)
	fmt.Println(r.URL)
}

// expect output: '/endpoint?&indexes=1%3B2' (';' escaped to '%3B')
// but got :      '/endpoint?2=&indexes=1'
```

PR #280, opened 2021-01-14
Fix invalid query string due to reserved char in query value issue 

wangyoucao577:
### Changes
- Escape primitive string when style param 
- Add unit tests 

Closes https://github.com/deepmap/oapi-codegen/issues/279


deepmap-marcinr:
I'm happy to merge this, thanks for the fix, could you resolve the conflicts?

PR #285, opened 2021-01-21
Implement marshalling of Date and aliased time parameters

karloygard:
Currently, marshalling of date parameters fails with an error similar to `error formatting 'date-from': unsupported type time.Time`.

To resolve this, I added marshalling of types.Date to `marshalTimeValue`, similar to what was already done for time.Time. Furthermore, I added support for aliasing of both types, which required some reflection magic.


karloygard:
Thank you for your effort!

PR #292, opened 2021-02-05
Parameter generated structures has params postfix to avoid names conflicting

deansotnikov:
Hi!
When 'parameter' has the same naming as 'response' they are conflicting.


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because parameter types already have a hardcoded "Params" suffix (e.g., `FindPetsParams`), and the configurable `response-type-suffix` addresses the remaining naming conflict scenarios that this PR was trying to solve.

PR #296, opened 2021-02-08
Support baseURL on server RegisterHandler

xoanmi:
Support the usage of a `baseURL` on the RegisterHandlers on the echo server


davidovich:
@xoanmi I am not sure why you would not just use `RegisterHandlersWithBaseURL` in your client code, as it is available to you. Changing the `RegisterHandlers` convenience method will break compatibility, and brings no benefit. Perhaps you meant to improve documentation for a `RegisterHandlersWithBaseURL` usage example ?

xoanmi:
@davidovich you're totally right! Thanks for the advice

Issue #312, opened 2021-03-05
Issue with generated client when colon appears in API path


bvwells:
There seems to be a bug is the way generated clients construct request paths from the server address and the operation path. The issue is shown when a colon appears in the API path, e.g. http://petstore.swagger.io/api/pets:validate.

The following spec illustrates the issue:

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  description: A sample API that uses a petstore as an example to demonstrate features in the OpenAPI 3.0 specification
  termsOfService: http://swagger.io/terms/
  contact:
    name: Swagger API Team
    email: apiteam@swagger.io
    url: http://swagger.io
  license:
    name: Apache 2.0
    url: https://www.apache.org/licenses/LICENSE-2.0.html
servers:
  - url: http://petstore.swagger.io/api
paths:
  /pets:validate:
    post:
      summary: Validate pets
      description: Validate pets
      operationId: validatePets
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/PetNames'
      responses:
        '200':
          description: valid pets
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Pet'

        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Error'
components:
  schemas:
    Pet:
      type: object
      required:
        - name
      properties:
        name:
          type: string
          description: The name of the pet.

    PetNames:
      type: object
      required:
        - names
      properties:
        names:
          type: array
          description: The names of the pets.
          items:
            type: string

    Error:
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
          description: Error code
        message:
          type: string
          description: Error message
```

The generated client for this spec contains the following code:

```Go
func NewValidatePetsRequestWithBody(server string, contentType string, body io.Reader) (*http.Request, error) {
	var err error

	queryUrl, err := url.Parse(server)
	if err != nil {
		return nil, err
	}

	basePath := fmt.Sprintf("/pets:validate")
	if basePath[0] == '/' {
		basePath = basePath[1:]
	}

	queryUrl, err = queryUrl.Parse(basePath)
	if err != nil {
		return nil, err
	}

	req, err := http.NewRequest("POST", queryUrl.String(), body)
	if err != nil {
		return nil, err
	}
...
```

[...]

PR #313, opened 2021-03-05
Fix client request url construction when colon in path

bvwells:
Fix client request url construction when colon in path. Fixes issue #312.


bvwells:
Need to test...

bvwells:
@deepmap-marcinr thanks for the review and merge!

rkojedzinszky:
Unfortunately, this breaks cases when a request parameter contains '/'. Earlier, this got escaped well, now an argument of 'main/i' gets escaped as: `main%252Fi`.

bvwells:
@rkojedzinszky can you provide an example spec and I'll take a look.

rkojedzinszky:
@bvwells I used latest, v1.5.2. The spec is a freenas 11.3 openapi, attached.
[freenas-11.3-openapi.json.gz](https://github.com/deepmap/oapi-codegen/files/6166710/freenas-11.3-openapi.json.gz)

For example, at path `/pool/dataset/id/{id}` the id parameter may be a zfs volume, containing `/`.

deepmap-marcinr:
I've got a fix for this which does what @bvwells  did in the client, but he missed the corresponding server side change to unescape arguments, so I fixed that side as well.

Your code used to work @rkojedzinszky because we got lucky to be consistently wrong :)

bvwells:
Sorry for missing this on the server side. Glad we got to the bottom of it as I couldn't work out what was wrong client side.

PR #322, opened 2021-03-20
Fix broken build due to kin backwards-incompatible change

nevivurn:
There was a backwards-incompatible change in `github.com/getkin/kin-openapi` 0.51.0: getkin/kin-openapi#326

This PR fixes the build & hopefully maintains the old behavior. Unfortunately, this means panicking on error, since that's what kin-openapi used to do: https://github.com/getkin/kin-openapi/blob/v0.50.0/openapi3filter/router.go#L87-L94

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/deepmap/oapi-codegen
 
 require (
 	github.com/cyberdelia/templates v0.0.0-20141128023046-ca7fffd4298c
-	github.com/getkin/kin-openapi v0.50.0
+	github.com/getkin/kin-openapi v0.51.0
 	github.com/go-chi/chi/v5 v5.0.0
 	github.com/golangci/lint-1 v0.0.0-20181222135242-d2cdd8c08219
 	github.com/labstack/echo/v4 v4.2.1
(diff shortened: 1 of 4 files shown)


deepmap-marcinr:
I had to roll this back.

Pulling in that routers/legacy package forces kin-openapi 0.52.0, which uses the //go:embed directive, which breaks on go 1.14, which is our minimum supported version.

nevivurn:
Ah, I thought it was fine since oapi-codegen was using `kin-openapi v0.50.0` when I wrote this. Fair enough.

I'm guessing that issues like these are going to be common as long as kin-openapi is sub-v1 and continues having breaking changes, which is unfortunate, but I guess that's what we get for depending on something sub-v1.

For now, I'll work around this by depending on my own fork. Thanks for the good work!

fenollp:
Hi, I maintain kin-openapi. As you may have noted the dependency on go:embed is only in a test. Let me see if I can have this test run iff go>1.14.

deepmap-marcinr:
Oh cool!

@fenollp  - Thank you so much for your work on kin-openapi. I can't imagine writing API's without it.

PR #323, opened 2021-03-25
Allow single dollar-sign properties

lwlee2608:
I got a [openapi3 spec file ](https://github.com/lwlee2608/oapi-codegen-testbed/blob/master/spec.yaml)that has the following discriminator property
```
      allOf:
        - $ref: '#/components/schemas/baseObject'
        - $ref: '#/components/schemas/childObject'
      discriminator:
        propertyName: $
```
I'm not too sure if this is valid. But online swagger editor seems to accept it.
This PR is a simple fix for this.


lwlee2608:
Without the fix, the generated object become:
```
type BaseObject struct {
	string  `json:"$"`
	Foo        *string `json:"foo,omitempty"`
}
```
instead of 
```
type BaseObject struct {
	DollarSign string  `json:"$"`
	Foo        *string `json:"foo,omitempty"`
}
```

Issue #324, opened 2021-03-26
Unmarshalling objects with AdditionalProperties results in nil for empty maps.


alehed:
When unmarshalling the follwing two json objects

```json
{
    "dict": null
}
```

```json
{
    "dict": {}
}
```

they both result in structs where AdditionalProperties is nil.

Ideally the second one would result in an AdditionalProperties which contains an empty map.

jxsl13:
same issue here

alehed:
This problem should be fixed now by the introduction of nullable-type feature.

PR #326, opened 2021-03-26
Unmarshal AdditionalProperties into a *map instead of a map

alehed:
This allows us to distinguish between `null` and `{}`.

An example of how these changes work can be found in [this go playground](https://play.golang.org/p/QJtC6ATcMJi).

This solves #324.


alehed:
The CI failure seems to be because it changes the generated code.

Since I think that the generated code needs changes, I guess this should be ok.

alehed:
Should this pr be based on v2 instead?

PR #327, opened 2021-03-26
Embed spec v2

dududko:
This PR adds the same feature as #320 except for the part that the old PR was based on the new feature in go 1.16 - embed directives, whereas this PR is based on the old -fashion way of embedding content of the spec file in the golang variable as the base64 encoded zipped value.
Due to this change all builds and tests are passed. 
Feel free to review the code and request changes or additional explanation.


go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/deepmap/oapi-codegen
 
 require (
 	github.com/cyberdelia/templates v0.0.0-20141128023046-ca7fffd4298c
-	github.com/getkin/kin-openapi v0.47.0
+	github.com/getkin/kin-openapi v0.49.0
```
deepmap-marcinr, line 5: Could you rebase and use v0.53.0? The version you're using still forces Go 1.16 because kin-openapi has //embed directives.

deepmap-marcinr (changes requested):

dududko:
@deepmap-marcinr done

wangyoucao577:
It's really great! Thanks!

PR #328, opened 2021-03-28
Descriptions for types

MarkRosemaker:
This PR adds support for type descriptions. 

Instead of
`// Foo defines model for Foo.`
it's possible to write a description in the yaml file and get
`// Foo represents a test object.`


pkg/codegen/templates/typedef.tmpl
```diff
@@ -1,5 +1,5 @@
 {{range .Types}}
-// {{.TypeName}} defines model for {{.JsonName}}.
+// {{.TypeName}} {{ with .Description }}represents {{ . }}{{ else }}defines model for {{.JsonName}}{{ end }}.
```
davidovich, line 2: If I define a description, I would find it strange to have "represents" added in a hard coded manner. What do you think about just taking the description {{.}} as is?

MarkRosemaker: You're right, I've changed it. Thanks!

Issue #329, opened 2021-03-30
[BUG] response body with multiple content type


lwlee2608:
If I have a response body defined as
```
      responses:
        '200':
          description: desc
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/responseObject'
            application/xml:
              schema:
                $ref: '#/components/schemas/responseObject'
```

The parse response function is bugged. The unmarshalled JSON should be set to `JSON200` instead of `XML200`
```
        switch {
        case strings.Contains(rsp.Header.Get("Content-Type"), "json") && rsp.StatusCode == 200:
                var dest ResponseObject
                if err := json.Unmarshal(bodyBytes, &dest); err != nil {
                        return nil, err
                }
                response.XML200 = &dest

        case strings.Contains(rsp.Header.Get("Content-Type"), "xml") && rsp.StatusCode == 200:
                var dest ResponseObject
                if err := xml.Unmarshal(bodyBytes, &dest); err != nil {
                        return nil, err
                }
                response.XML200 = &dest

        }
```
Full spec file
https://github.com/lwlee2608/oapi-codegen-testbed/blob/json-or-xml-response-body/spec.yaml

deepmap-marcinr:
It is bugged indeed, thanks for spotting it.

lwlee2608:
Take a look at PR #330 whenever you're free. I think I've fixed it.

lwlee2608:
Thank you :smile:

PR #330, opened 2021-03-30
fix when responseBody has multiple content types (fixed for #329)

lwlee2608:
This is a fixed for #329 

The bug occurs when `handledCaseClauses` map json casekey has been overridden by xml one.


lwlee2608:
Looks like the test has been broken, unsupported content-type no longer being generated. I can fix this tomorrow.

deepmap-marcinr:
thanks.

Issue #251, opened 2020-11-06
Versioned templates


deepmap-marcinr:
This project has reached a bit of an impasse, in that when we make changes to improve code generation, and people regenerate code, their code built around the generated code frequently breaks.

Semver doesn't work well here because we have two dimensions - the generator itself, and the generated code.

So, to that end, we're going to make one more breaking change, and that's to reorganize the templates for versioning.

We have distinct categories of templates - model types, Echo glue and Chi glue. We'll break those into subdirectories.

Each subdirectory will be prefixed with a version, eg, v1/models , v2/echo

When oapi codegen makes a breaking change to templates, we will ship all versions, but default to the newest version, and allow the user to override.

For example, if code was generated with (echo:v1, models:v2) and we make a breaking change, we will ship oapi-codegen defaulting to (echo:v2, models:v3). If a user is dependent on previous behavior, they could specify --models-version=v2.

wangyoucao577:
Additionally I suggest to use semver on the generator(binary) itself and runtime(can be specified in `go.mod`).         
Both generator's semver and templates(e.g., models, echo, chi, etc.) version should be written in to generated code (`xxx.gen.go`) for visualization, something like `protoc`/`protoc-gen-go` do:         
```go
// Code generated by protoc-gen-go. DO NOT EDIT.
// versions:
// 	protoc-gen-go v1.25.0-devel
// 	protoc        v3.12.3
// source: sample.proto
```      

I'm considering to create a PR to add generator itself's version, which can be used for CLI output or code generation output. On the other hand, it'll affect the release process to below:       
- create a release branch
- change the version info to `vx.x.x` rather than `devel` 
- create github release from the release branch 

Is it ok to you guys? Actually the release process can be automated via Github Actions, and I can help to setup it also. Let me know if any question or other idea.        
Thanks for the excellent project!

deepmap-marcinr:
For those of you interested in the version information in oapi-codegen, what do you think of this comment?

https://github.com/deepmap/oapi-codegen/pull/337#issuecomment-846281759

It's different from @wangyoucao577 's implementation, in that it doesn't produce binary releases, but uses debug.BuildInfo to inject the version when someone does a `go get`. The version isn't visible locally when working on this module itself, though, so within this repo, version just becomes `(devel)`

See the `versiontest` branch.

wangyoucao577:
[...]

PR #337, opened 2021-04-09
Add binary version

wangyoucao577:
## Why 
Thanks for the awesome project, I can generate many codes by `oapi-codegen` from OpenAPI v3 spec. However, I usually have a question: what version of the `oapi-codegen` I use? The project still goes very fast, and I have to upgrade it per several days/weeks to fix bugs or leverage new features. It's so important to know that where am I before upgrade.       
To achieve this, I think there're several steps to do:     
1.  Add version on `oapi-codegen`;    
2. Also write version information to generated codes.      

https://github.com/deepmap/oapi-codegen/issues/251#issuecomment-760804314 describes the idea too.     

## What/How does this PR solve        
This PR intend to solve the problem 1: add version on `oapi-codegen`.       
It did below things:     
- add version related information to `oapi-codegen` binary, EMPTY by default;      
- add release automation(via Github Actions) that will be trigger when create release. It will build and upload binaries to the release page automatically. When build the binaries, version information will be set by `-ldflags`. More specific, the "AppVersion" will be set with the release tag. Several platforms(`linux/amd64`, `darwin/amd64`, `windows/amd64`) have been configured to cover most of use cases.      

https://github.com/wangyoucao577/oapi-codegen/releases/tag/v1.6.0.1-test-bin-version shows the results. If download the release binary and run `-version`, you'll get something like          
```bash
./oapi-codegen -version
Version:    v1.6.0.1-test-bin-version
Build Time: Fri Apr 9 08:30:03 UTC 2021
Git Commit: eba83236752fe3713e26828756884419c727a23a
Git Ref:    refs/tags/v1.6.0.1-test-bin-version
```

I think this approach is better than write version in code because it no need to change any process when release new versions.          

## Next    
If you're good with it, I'll create another PR to solve the step 2: Also write version information to generated codes(as comments).         
After that, we'd better to encourge to use the pre-built release binaries instead of `go get`, otherwise the version information can not be used.     

@deepmap-marcinr How's your idea?

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,39 @@
+name: Release Go Binaries
+
+on: 
+  release:
+    types: [created]
+  workflow_dispatch:
+
+env:
+  CMD_PATH: ./cmd/oapi-codegen
+
+
+jobs:
+  releases-matrix:
+    name: Release Bin
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        goos: [linux, windows, darwin]
+        goarch: [amd64]
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set APP_VERSION env
+      run: echo APP_VERSION=$(echo ${GITHUB_REF} | rev | cut -d'/' -f 1 | rev ) >> ${GITHUB_ENV}
+    - name: Set BUILD_TIME env
+      run: echo BUILD_TIME=$(date) >> ${GITHUB_ENV}
+    - name: Environment Printer
+      uses: managedkaos/print-env@v1.0
+
+    - uses: wangyoucao577/go-release-action@v1.16
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        goos: ${{ matrix.goos }}
+        goarch: ${{ matrix.goarch }}
+        goversion: https://golang.org/dl/go1.16.3.linux-amd64.tar.gz
+        project_path: ./cmd/oapi-codegen
+        build_flags: -v
+        ldflags: -X "main.appVersion=${{ env.APP_VERSION }}" -X "main.buildTime=${{ env.BUILD_TIME }}" -X main.gitCommit=${{ github.sha }} -X main.gitRef=${{ github.ref }}
+        
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 .idea
 .DS_Store
 .vscode/
+
+# bin
+cmd/oapi-codegen/oapi-codegen
+__debug_bin
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -44,6 +44,7 @@ var (
 	flagExcludeSchemas string
 	flagConfigFile     string
 	flagAliasTypes     bool
+	flagPrintVersion   bool
 )
 
 type configuration struct {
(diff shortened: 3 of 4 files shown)


falzm:
This feature is **much needed**, any reason you haven't reacted to this PR @deepmap-marcinr?

deepmap-marcinr:
Thanks for doing this - you've done some magic that I wasn't familiar with. Is the change to versioning that I proposed feasible? I intend to keep good tagging discipline, so when we have a v2 branch, it will continue to work.

wangyoucao577:
> Thanks for doing this - you've done some magic that I wasn't familiar with. Is the change to versioning that I proposed feasible? I intend to keep good tagging discipline, so when we have a v2 branch, it will continue to work.


Yes, the versioning that you proposed still works. What this PR done is mostly on binary version, as well as the repo's version. The versioning you proposed mainly on sub-models version. For example, we could have an `oapi-codegen` version `v2.1`, which includes `v1/echo`, `v2/echo` and `v1/chi`. They're not conflict.

deepmap-marcinr:
I'm still thinking about how to do this, since I'd really like versioning to work if you `go get` the executable, since that's a really easy use case.

I'm thinking that on each release, I can bake the version into the go code itself, eg, (v1.6.1), then post-release, immediately rev it to something (v1.6.2-pre). It's got the downside that the granularity of versions isn't down to the commit, but down to the release, which I think is ok.

davidovich:
@deepmap-marcinr, This is possible now with latest go releases. Here is an example implementation: https://github.com/davidovich/summon/blob/master/cmd/root.go#L172-L193

deepmap-marcinr:
I wasn't aware of the debug.BuildInfo. It seems very limited. It'll work via `go get`, but the version would always be "devel" when working on it... interesting. Need to do some experiments.

deepmap-marcinr:
I made a test repository and tried all kinds of version tags across a v2 mainline and v1 branch, and this buildInfo thing works pretty well.

deepmap-marcinr:
I've made a branch named `versiontest` which has BuildInfo based versions:

```
$ go get  github.com/deepmap/oapi-codegen/cmd/oapi-codegen@versiontest
$ oapi-codegen --version
v1.6.2-0.20210521214225-6862a500dfec
```

Comments at the top of generated files look like this:
```
//
// Code generated by github.com/deepmap/oapi-codegen version v1.6.2-0.20210521214225-6862a500dfec DO NOT EDIT.
```

Maybe this is good enough, and saves having to produce binary releases.

wangyoucao577:
> I've made a branch named `versiontest` which has BuildInfo based versions:
> 
> ```
> $ go get  github.com/deepmap/oapi-codegen/cmd/oapi-codegen@versiontest
> $ oapi-codegen --version
> github.com/deepmap/oapi-codegen/cmd/oapi-codegen
> v1.6.2-0.20210521214225-6862a500dfec
> ```
> 
> Comments at the top of generated files look like this:
> 
> ```
> //
> // Code generated by github.com/deepmap/oapi-codegen version v1.6.2-0.20210521214225-6862a500dfec DO NOT EDIT.
> ```
> 
> Maybe this is good enough, and saves having to produce binary releases.

It looks pretty good! It's able to solve the trace source code problem either!

falzm:
> Maybe this is good enough, and saves having to produce binary releases.

Having a way to tell the current version from the command line would be nice at some point, but this is a good first step 👍

deepmap-marcinr:
I added a —version flag, that’s there

PR #342, opened 2021-04-21
fix: better enum deduplication

### LINKED ISSUE #341: Error building code generated with enum values as digits
If there are numbers in enum values generated code is impossible to build.
yaml example:

```
openapi: 3.0.2

info:
  title: ...
  version: 0.0.0

paths:
  /foo:
    get:
      responses:
        200:
          description: ...
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Bar'

components:
  schemas:
    Bar:
      type: string
      enum: 
        - 1
        - Foo
        - Bar
        - Foo Bar
        - Foo-Bar
        - 1Foo
        - Bar
        - ' Foo'
        - ' Foo '
        - _Foo_
```

The issue is related due to enum values map generator.
To be percise the problem appears here:
```sanitized := SchemaNameToTypeName(SanitizeGoIdentity(n))```
it may be related to @vlastahajek fix https://github.com/deepmap/oapi-codegen/pull/241

Result code for the above yaml will have:
```
const (
	Bar = "1"
	...
```
and 
```
// Bar defines model for Bar.
type Bar string
```

### PR DESCRIPTION
**Pull request by vlastahajek**
Closes #341 

Fixed enum deduplication.
It will cause breaking changes in cases, when enum value starts with a number. Number char was previously removed.

```yaml
    Bar:
      type: string
      enum:
        - Foo
        - 1Foo
```
Before:
```go
const (
	BarFoo Bar = "1Foo"
	BarFoo1 Bar = "Foo"
)
```
Now:
```go
const (
	Bar1Foo Bar = "1Foo"
	BarFoo Bar = "Foo"
)
```

### DIFF
--- a/internal/test/issues/issue-illegal_enum_names/issue.gen.go
+++ b/internal/test/issues/issue-illegal_enum_names/issue.gen.go
@@ -22,7 +22,13 @@ import (
 
 // Defines values for Bar.
 const (
+	Bar1 Bar = "1"
+
+	Bar1Foo Bar = "1Foo"
+
 	BarBar Bar = "Bar"
+
+	BarBar1 Bar = "Bar"
 
 	BarFoo Bar = "Foo"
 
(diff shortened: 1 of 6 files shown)


mr-tron:
Any news? The sooner we merge this pull request, the less problems we'll have later.

deepmap-marcinr:
I've finally fixed this in a different way. Thanks for the PR, though.

PR #344, opened 2021-04-21
feat: allow oneOf with just 'required' property

vlastahajek:
This PR allows generating concrete type instead if general the `interface{}` when `oneOf` property defines just required field:

```yaml
    Bar:
      type: object
      properties:
        id:
          type: string
        name:
          type: string
        foo:
          type: string
        coo:
          type: string
      oneOf:
        - required:
          - id
          - name
          - foo
        - required:
          - id
          - name
          - coo
```

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -169,8 +169,9 @@ func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
 		outSchema.GoType = "interface{}"
 		return outSchema, nil
 	}
-	// We can't support this in any meaningful way
-	if schema.OneOf != nil {
+	// We can't support this in any meaningful way,
+	// except if oneOf defines just restriction on properties via the "required" property
+	if schema.OneOf != nil && !(schema.OneOf[0].Value != nil && schema.OneOf[0].Value.Required != nil) {
 		outSchema.GoType = "interface{}"
 		return outSchema, nil
 	}


pkg/codegen/schema.go
```diff
@@ -169,8 +169,9 @@ func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
 		outSchema.GoType = "interface{}"
 		return outSchema, nil
 	}
-	// We can't support this in any meaningful way
-	if schema.OneOf != nil {
+	// We can't support this in any meaningful way,
+	// except if oneOf defines just restriction on properties via the "required" property
+	if schema.OneOf != nil && !(schema.OneOf[0].Value != nil && schema.OneOf[0].Value.Required != nil) {
```
rogpeppe, line 174: This logic only looks at the first item in the `OneOf` slice. What if there are more items in the slice and they're not all "required"?

vlastahajek: In such a case it wouldn't be correct. The assumption here is that purpose of `oneOf` is that all items are of the same kind (type, restriction). 
Sure, in order to be completely correct all items should be checked.

rogpeppe: Given that this is dealing with arbitrary OpenAPI specs, ISTM that that assumption should be checked in the code, otherwise this probably isn't a reasonable thing to accept into master.

rogpeppe: One other thing: this will panic if `schema.OneOf` is non-nil but zero length.

vlastahajek:
As this very edge case for oneOf and, also as, time went by, there is no need for this.

PR #348, opened 2021-04-27
fix build with kin-openapi 0.61.0

mgabeler-lee-6rs:
Addresses changes from https://github.com/getkin/kin-openapi/pull/351

### DIFF
--- a/examples/petstore-expanded/chi/api/petstore.gen.go
+++ b/examples/petstore-expanded/chi/api/petstore.gen.go
@@ -328,12 +328,12 @@ func PathToRawSpec(pathToFile string) map[string]func() ([]byte, error) {
 // The logic of resolving external references is tightly connected to "import-mapping" feature.
 // Externally referenced files must be embedded in the corresponding golang packages.
 // Urls can be supported but this task was out of the scope.
-func GetSwagger() (swagger *openapi3.Swagger, err error) {
+func GetSwagger() (swagger *openapi3.T, err error) {
 	var resolvePath = PathToRawSpec("")
 
-	loader := openapi3.NewSwaggerLoader()
+	loader := openapi3.NewLoader()
 	loader.IsExternalRefsAllowed = true
-	loader.ReadFromURIFunc = func(loader *openapi3.SwaggerLoader, url *url.URL) ([]byte, error) {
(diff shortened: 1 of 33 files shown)


ondrejbudai:
Changes looking good. Can we get a review from a maintainer and a new release? Thanks! :)

stellirin:
I'm guessing there's a good (legal) reason for kin-openapi to do this, I suggest `swagger` should be fully purged from this repo too, not just make the minimal changes to adapt to upstream.

mgabeler-lee-6rs:
> I'm guessing there's a good (legal) reason for kin-openapi to do this

I was wondering the same thing, but the kin PR makes no mention of any specific motivation for it. Esp. since the spec _itself_ requires using the "swagger" name in fields, I wonder if this was either just paranoia, or desiring to be generic in the naming. Either way, I will ask the kin folks to comment here to hopefully replace speculation with information :)

> I suggest `swagger` should be fully purged from this repo too, not just make the minimal changes to adapt to upstream.

This certainly would be safer, though I wonder if it would be reasonable to do in separate steps?

mgabeler-lee-6rs:
Added #354 which includes the changes here, and also makes an attempt to remove the Swagger name from all APIs

fenollp:
Hi, kin-openapi maintainer here.
Swagger is a name trademarked by an entity (SmartBear) other than the Linux Foundation.
OpenAPIv2 (as well as v1) do have a field named `swagger` and for legacy reasons it's here for good and free of conflicts. OpenAPIv3 does not, though.

Please don't call us paranoid. I already spend enough time looking over my shoulder.

mgabeler-lee-6rs:
> Please don't call us paranoid. I already spend enough time looking over my shoulder.

Apologies if my wording came across poorly. What I meant was that I wasn't sure if this was a pro-active change made to err on the side of safety, or if there had been a specific request / letter from the Swagger folks to not use their name in your library.

lukestoward:
I came looking for this PR as I have a project in which I attempted to update all of the dependencies, only for it to fail due to the generated spec using properties defined by `kin-openapi v0.53.0` and the update command attempts to build my source using `v0.61.0`. I worked around this by updating all other dependencies individually, which isn't a big issue, but this PR would allow me to update everything in one go. 😄

thenewwazoo:
Another +1 for merging this. The instructions as written currently don't work.

PR #354, opened 2021-05-06
fix 

mgabeler-lee-6rs:
Addendum to #348, which makes Breaking Changes, to similarly remove the Swagger name from all APIs


mvrhov:
@mgabeler-lee-6rs, it seems this needs to be rebased before it can be merged.

mgabeler-lee-6rs:
rebased :+1:

PR #356, opened 2021-05-10
Added extension for adding extra tags

elliotmr:
This pull request adds the ability to specify a oapi-codegen specific extension for generating struct types with extra tags. 

Web applications often need input validation and object relational mapping for databases. Several of the most popular libraries in the Go ecosystem use custom tags for implementing those functions. By allowing the user to specify extra tags we can then directly use the generated structures in those packages.


elliotmr:
This currently just takes a single string as an input. However, I think it might be better to take a list of strings instead.

I will test it out and update the pull request if it feels better.

amanbolat:
Can't wait it to be implemented. Having this feature will make life easier. It would be very helpful for adding some validation tags and just calling `validator.Validate(&request)` on Router level.

Issue #147, opened 2020-03-08
Support for `Content-Type: application/vnd.[...]+json`


chlunde:
Currently `application/vnd.$VENDOR+json` is not fully supported, only `application/json`: types are generated but `...WithResponse` response types will not deserialize the data, and requests with bodies are generated without a serializer.

For example, [docker distribution API ](https://docs.docker.com/registry/spec/api/) uses `application/vnd.docker.distribution.manifest.v2+json` in both `Accept` and `Content-Type`.  [Github API V3](https://developer.github.com/v3/media/) uses `Accept: application/vnd.github[.version].param[+json]` but responses have `Content-Type: application/json`.

Additionally, the current implementation does not work when schemas define content-types with encodings, i.e. `mediatype; charset=UTF-8`
# Example schema

```json
        "requestBody" : {
          "description" : "Authentication data",
          "content" : {
            "application/vnd.APP-v1+json; charset=UTF-8" : {
              "schema" : {
                "$ref" : "#/components/schemas/AuthData"
              }
            }
          },

[...]

          "500" : {
            "description" : "Internal server error. See error message for details.",
            "content" : {
              "application/vnd.APP-v1+json; charset=UTF-8" : {
                "schema" : {
                  "$ref" : "#/components/schemas/Error"
                }
              }
            }
```
# Request
## Current
```go
// The interface specification for the client above.                                                                                                                                 
 type ClientInterface interface {                                                                                                                                                     
     // UsernamePassword request  with any body                                                                                                                                       
     UsernamePasswordWithBody(ctx context.Context, contentType string, body io.Reader) (*http.Response, error)                              

```

## Expected
```go
+    UsernamePassword(ctx context.Context, body UsernamePasswordJSONRequestBody) (*http.Response, error)                                                                              
```
# Response
## Current

```go
 type usernamePasswordResponse struct {
        Body         []byte
        HTTPResponse *http.Response
 }
```
## Expected
```go
+       JSON200      *Session
+       JSON500      *Error
```
# Quick hack/POC
```diff
diff --git a/pkg/codegen/operations.go b/pkg/codegen/operations.go
index 916643d..541fd7c 100644
--- a/pkg/codegen/operations.go
+++ b/pkg/codegen/operations.go
@@ -281,5 +281,5 @@ func (o *OperationDefinition) GetResponseTypeDefinitions() ([]TypeDefinition, er
                                        var typeName string
                                        switch {
[...]

PR #364, opened 2021-05-18
Generate Go structs for vendor-specific JSON media types

slee1728:
Resolves https://github.com/deepmap/oapi-codegen/issues/147

Description:
- Modify oapi-codegen to generate structs for vendor-specific JSON media types.
- Introduce a commandline flag to customize generated type names. For example, one can extract the version in a media type and use it as the suffix (type name in schema = body, media type = application/vnd.company.v1+json => BodyV1 in `types.gen.go`)


slee1728:
I'd love to get some feedback from the maintainers at this point.

jamietanna:
We've now got support for this, as part of #581

PR #374, opened 2021-05-28
Add support for all integer types

DerekStrickland:


pkg/codegen/schema.go
```diff
@@ -340,12 +340,24 @@ func resolveType(schema *openapi3.Schema, path []string, outSchema *Schema) erro
 		// We default to int if format doesn't ask for something else.
 		if f == "int64" {
 			outSchema.GoType = "int64"
-		} else if f == "uint64" {
-			outSchema.GoType = "uint64"
 		} else if f == "int32" {
 			outSchema.GoType = "int32"
+		} else if f == "int16" {
```
deepmap-marcinr, line 345: these are not valid OpenAPI formats.

PR #382, opened 2021-06-09
Create non-pointer map/slice fields from optional properties

### LINKED ISSUE #266: Create slice/map/string fields without pointers (for optional fields)
Default value for slice/map is `nil`. So slice/map field will not be serialized if it is not initialized.

I mean from serialization point of view these two structs are the same:

```
struct {
  GroupKey *[]string `json:"group_key,omitempty"`
}
```
```
struct {
  GroupKey []string `json:"group_key,omitempty"`
}
```

The later version is more convenient to work with..

https://play.golang.org/p/0c0nVaoOPNK

Also the same applies to `string` type with `omitempty` json tag.

Pointer to interface `*interface{}` struct fields are also useless.

karupanerura:
Partially resolves #266.

No support for string, as it is still under discussion.
map/slice can represent nil without making it a pointer, so this patch should be able to resolve the issue about for map/slice.

### DIFF
--- a/examples/petstore-expanded/chi/api/petstore.gen.go
+++ b/examples/petstore-expanded/chi/api/petstore.gen.go
@@ -52,7 +52,7 @@ type Pet struct {
 type FindPetsParams struct {
 
 	// tags to filter by
-	Tags *[]string `json:"tags,omitempty"`
+	Tags []string `json:"tags,omitempty"`
 
 	// maximum number of results to return
 	Limit *int32 `json:"limit,omitempty"`
--- a/examples/petstore-expanded/chi/api/petstore.go
+++ b/examples/petstore-expanded/chi/api/petstore.go
@@ -47,7 +47,7 @@ func (p *PetStore) FindPets(w http.ResponseWriter, r *http.Request, params FindP
 	for _, pet := range p.Pets {
 		if params.Tags != nil {
 			// If we have tags,  filter pets by tag
-			for _, t := range *params.Tags {
+			for _, t := range params.Tags {
 				if pet.Tag != nil && (*pet.Tag == t) {
 					result = append(result, pet)
 				}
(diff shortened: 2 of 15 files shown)


ckarenz:
I'd love to see this go in.  Is there an issue blocking merge?

The build failure appears to come from the main branch.

karupanerura:
Thank you. I think you are right, the build failure appears to come from the main branch.
Now I'm waiting for the fix and this patch to be reviewed and accepted. No TODO.

justjhon:
@deepmap-marcinr is there any updates on this change?

deepmap-marcinr:
This will break existing generated code. I've taken the approach of making these kinds of changes changed by flag. Your change moves the decision from templates (`{{if not .Requored}}`) to code. Could you extend the code generation options to provide the old behavior with an option? Please see: https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/configuration.go#L37

Whenever changes liek this go in, they cause some hate mail.

dlsniper:
Is there any interest in this PR?

I just bumped into this and would like to get it sorted.

@karupanerura do you have any interest in doing the change yourself? Should I fork the PR and send an upgraded one to address anything else required to merge it? 
Do maintainers have any interest to get it done?

Thank you?

karupanerura:
@dlsniper Sorry for the late reply. I'm happy if you fork from this branch. I don't have enough time to solve the problem for a while.

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the `prefer-skip-optional-pointer-on-container-types` configuration option now exists and handles both maps and slices. This PR's exact feature has been implemented as an opt-in configuration option.

PR #390, opened 2021-06-24
Added implementation of x-json-ignore

sparshev:
The implementation of x-go-json-ignore to use with GORM when we don't need to serialize specific fields (when ObjectID and Object fields is in the same struct).

### DIFF
--- a/pkg/codegen/extension.go
+++ b/pkg/codegen/extension.go
@@ -6,10 +6,11 @@ import (
 )
 
 const (
-	extPropGoType    = "x-go-type"
-	extGoName        = "x-go-name"
-	extPropOmitEmpty = "x-omitempty"
-	extPropExtraTags = "x-oapi-codegen-extra-tags"
+	extPropGoType       = "x-go-type"
+	extGoName           = "x-go-name"
+	extPropGoJsonIgnore = "x-go-json-ignore"
+	extPropOmitEmpty    = "x-omitempty"
+	extPropExtraTags    = "x-oapi-codegen-extra-tags"
 )
 
 func extString(extPropValue interface{}) (string, error) {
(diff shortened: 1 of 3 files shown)


sparshev:
Hi @deepmap-marcinr, could you please suggest who can check the PR? If it's good - I can work on the tests and other required improvements to make sure it will fit perfectly.

deepmap-marcinr:
I'm traveling on vacation right now, I'll take a look soon.

pkg/codegen/schema.go
```diff
@@ -465,6 +472,7 @@ func GenFieldsFromProperties(props []Property) []string {
 			tags[i] = fmt.Sprintf(`%s:"%s"`, k, fieldTags[k])
 		}
 		field += "`" + strings.Join(tags, " ") + "`"
+
```
jamietanna, line 475:
```suggestion
```

pkg/codegen/extension.go
```diff
@@ -51,3 +52,18 @@ func extExtraTags(extPropValue interface{}) (map[string]string, error) {
 	}
 	return tags, nil
 }
+
+func extParseGoJsonIgnore(extPropValue interface{}) (bool, error) {
+
```
jamietanna, line 57:
```suggestion
```

jamietanna (commented):
Looking good! Would you mind updating the branch, to make sure CI is happy?

We will look at confirming functionality wise it looks good / whether we need any new tests

sparshev:
I actually thought this change will never be here... That probably will be a pain, but let me try to recall what happened here.

sparshev:
Ok, I removed the logic optimization and left just the code related to the json-ignore. @jamietanna could you please approve the workflows?

pkg/codegen/schema.go
sparshev: fixed

pkg/codegen/extension.go
sparshev: fixed

sparshev:
I hope it's not another year to wait...

PR #404, opened 2021-07-18
Support standard encoding.TextUnmarshaler interface

hexdigest:
Hi,

First of all thank you for this wonderful tool it really makes my life so much easier!

There are various standard types supporting standard encoding.TextUnmarshaler interface.
Here are few very famous ones for example: 
* big.Int 
* net.IP
* google.UUID

And there are thousands of other custom types implementing this interface can be found on GitHub
Currently if you try to use such types as a parameters you're getting binding error like:
```Invalid format for parameter id: error binding string parameter: can not bind to destination of type: array``` in case of net.IP or google.UUID

I know that there was a similar PR #262 but unfortunately it doesn't allow you to use thousands of existing types and forces you to create a new type with Bind(string) error method implemented on it. This is very inconvenient and this PR solves the problem.

### DIFF
--- a/pkg/runtime/bindparam.go
+++ b/pkg/runtime/bindparam.go
@@ -14,6 +14,7 @@
 package runtime
 
 import (
+	"encoding"
 	"encoding/json"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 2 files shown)


hexdigest:
@deepmap-marcinr Hi, is there any chance to get this merged?

deepmap-marcinr:
Thank you.

PR #415, opened 2021-08-03
updated the version of echo & go

Taillook:
Updated echo v4.2.1 to v4.5.0
and go1.15

because problem by github.com/dgrijalva/jwt-go.
Includes security upgrades.

> So for JWT middleware users Go 1.15+ is required

ref: https://github.com/labstack/echo/releases/tag/v4.5.0


Taillook:
@deepmap-marcinr 
Hi.
If possible, please check as soon.

Taillook:
resolved conflict.

Taillook:
@deepmap-marcinr 
fixed conflict.

deepmap-marcinr:
I've started working on a V2 of this library, which will make compatibility easier, and that one's already on 1.16. If I merge this, we break compatibility with Go 1.14, which would mean increasing the semver major version. I guess we could push a release with a new minor version.

Taillook:
@deepmap-marcinr 
recreated pull request on #430 for v2-dev.

Issue #418, opened 2021-08-10
Deprecate github.com/pkg/errors


andig:
Influx uses this package for code generation. With go 1.13 as minimum version it should be possible to use the newer Go mechanisms of errors.Is/As instead of pkg/errors for the generated code.

Refs https://github.com/influxdata/influxdb-client-go/issues/262

oncilla:
I think this can be closed now that #420 is merged.

PR #420, opened 2021-08-13
Abolished github.com/pkg/errors

Taillook:
Fix for #418
To move `github.com/pkg/errors` to `erros`.

### DIFF
--- a/internal/test/components/components.gen.go
+++ b/internal/test/components/components.gen.go
@@ -20,7 +20,6 @@ import (
 	"github.com/deepmap/oapi-codegen/pkg/runtime"
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/labstack/echo/v4"
-	"github.com/pkg/errors"
 )
 
 // Has additional properties of type int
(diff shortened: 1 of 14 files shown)


Taillook:
Fixed test at https://github.com/deepmap/oapi-codegen/pull/410

internal/test/components/components.gen.go
```diff
@@ -171,7 +170,7 @@ func (a ParamsWithAddPropsParams_P1) MarshalJSON() ([]byte, error) {
 	for fieldName, field := range a.AdditionalProperties {
 		object[fieldName], err = json.Marshal(field)
 		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("error marshaling '%s'", fieldName))
+			return nil, fmt.Errorf("error marshaling '%s'", fieldName)
```
ckarenz, line 173: Throughout this change, you're losing the wrapped error.  This would be equivalent to the old error:
```suggestion
			return nil, fmt.Errorf("error marshaling '%s': %w", fieldName, err)
```

ckarenz:
I might also recommend converting `'%s'` to `%q` globally, but that would change output and probably deserves its own discussion and PR.

ckarenz (changes requested):
Edit: Removed duplicate comment.

internal/test/components/components.gen.go
Taillook: got it, thanks.

Taillook:
> Throughout this change, you're losing the wrapped error. This would be equivalent to the old error:

resolved format.

Taillook:
> I might also recommend converting '%s' to %q globally, but that would change output and probably deserves its own discussion and PR.

Yes, Not supported here.
(And need many changes...

ckarenz (approved):
Looks great!

Issue #221, opened 2020-08-22
Can't pass security validation when using oapi validation middleware


bandirsen:
I use this library on sever side, while on client side do manually since the client is for mobile platform.

This is my openapi spec :
I follow [openapi documentation](https://swagger.io/docs/specification/authentication/bearer-authentication/) to define security scheme
```
paths:
  /admins:
    get:
      description: Get AdminList
      security:
        - bearerAuth: []
      operationId: getAdminList
    
components:
  securitySchemes:
    bearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
```

And in my request I put Authorization header on request
`Authorization: Bearer ---random JWT token ----`

But I always get 'Security requirements failed' from validation middleware (the JWT token was checked and valid)
Am I missing something ?

PS: also it return http error 403, while above documentation recommending to return http 401 error for unauthorized request.

zaccari:
I'm seeing this error as well, with the global `security` set (not at the path level):

```yaml
paths:
  /models:
    post:
    ...
components:
  securitySchemes:
    bearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
security:
  - bearerAuth: []
```

Removing the global `security` lets the validation pass:

```bash
curl -X POST "http://127.0.0.1:8090/models" -H "Authorization: Bearer a503faf9-45b5-4fec-8334-337284a66ea4" -H  "accept: application/json" -H  "Content-Type: application/json" -d "{\"name\":\"string\",\"type\":\"string\",\"options\":\"string\",\"parentId\":\"string\"}"
```

I'm looking into the issue now, just wanted to make a note of it here for reference

zaccari:
@bandirsen I was able to resolve the issue by providing a custom validator function to the options, like so:

```go
	validatorOptions := &oapimiddleware.Options{}

	validatorOptions.Options.AuthenticationFunc = func(c context.Context, input *oapifilter.AuthenticationInput) error {
		fmt.Println(">>>> INSIDE AuthenticationFunc")
		return nil
	}

	e.Use(oapimiddleware.OapiRequestValidatorWithOptions(swagger, validatorOptions))

```
If I don't pass a validator function, I get the `Security requirements failed` error

bandirsen:
@zaccari
 
[...]

PR #423, opened 2021-08-16
Add NoopAuthenticationFunction in RequestValidation

BananaNosh:
Adresses #221.

When using security schemes, it is not possible to pass the RequestValidation, because it is enherited from openapi3filter which enforces an AuthenticationFunction in the Options. 
This function, however, is not used by the echo/chi server and can be just set to Noop.
This way the RequestValidation will pass.
For the security the server must implement some other kind of Authorization Middleware, anyway.

### DIFF
--- a/pkg/chi-middleware/oapi_validate.go
+++ b/pkg/chi-middleware/oapi_validate.go
@@ -34,6 +34,8 @@ func OapiRequestValidatorWithOptions(swagger *openapi3.T, options *Options) func
 		panic(err)
 	}
 
+	addNoopAuthentication(options)
+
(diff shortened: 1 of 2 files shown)


BananaNosh:
I do not really get, why there is a diff in the CI. Can someone explain.

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the middleware code has been extracted to separate repositories (e.g., `github.com/oapi-codegen/echo-middleware`). No validation middleware exists in the main repo. This PR targets code that no longer lives here.

PR #424, opened 2021-08-17
[#211] Generate OpenAPI dictionaries without fixed keys as maps

### LINKED ISSUE #211: infer map from additionalProperties
Currently [the spec](https://swagger.io/docs/specification/data-models/dictionaries/) suggests you use `additionalProperties` to define typed maps. Is it possible to infer `map[string]Xxx` from `struct { AdditionalProperties map[string]Xxx }`? Clearly if other fields are present, the overflow map is required, but it would be nice to have things flattened more.

natsukagami:
Closes #211.

Generate all objects with the `additionalProperties` value set, but no `properties` set as maps.

For example, these schemas:
```yaml
anyMap:
    type: object
    additionalProperties: true
numberMap:
    type: object
    additionalProperties:
        type: integer
mapWithFixedKeys:
    type: object
    additionalProperties:
        type: string
    properties:
        key:
            type: string
```

produces the following

```go
type AnyMap map[string]interface{}

type NumberMap map[string]int

type MapWithFixedKeys struct {
    Key                  string
    AdditionalProperties map[string]string
}
```

instead of

```go
type AnyMap struct {
    AdditionalProperties map[string]interface{}
}

type NumberMap struct {
    AdditionalProperties map[string]int
}

type MapWithFixedKeys struct {
    Key                  string
    AdditionalProperties map[string]string
}
```

**This is a small breaking change.**

### DIFF
--- a/internal/test/components/components.gen.go
+++ b/internal/test/components/components.gen.go
@@ -57,9 +57,7 @@ type AdditionalPropertiesObject4_Inner struct {
 }
 
 // Has additional properties with schema for dictionaries
-type AdditionalPropertiesObject5 struct {
-	AdditionalProperties map[string]SchemaObject `json:"-"`
-}
+type AdditionalPropertiesObject5 map[string]SchemaObject
 
 // ObjectWithJsonField defines model for ObjectWithJsonField.
 type ObjectWithJsonField struct {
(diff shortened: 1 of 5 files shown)


deepmap-marcinr:
Sorry to take so long to get to this.

Would you mind adding a configuration option for old schema behavior in the new Configuration struct, and allow disabling this new behavior? I guess we could call it "DisableFlattenAdditionalProperties" or something liek that - so that your new behavior is the default, but we can override it off. I'd be happy to merge this.

deepmap-marcinr:
If not, I can take care of it, but I'm not entirely sure how to do it in github so you get a commit credit :)

natsukagami:
Please take care of it. It's been a long time since I touched the codebase, I am no longer familiar with it.

As for the commit credit, I think as long as you merge or squash with the "Co-author" fields in the commit message intact GitHub will treat it well :P

Issue #301, opened 2021-02-19
Open to new routers?


Jleagle:
Hi, i was thinking about adding Fiber https://github.com/gofiber/fiber as a server option, but before i started wanted to check if you are open to new server options?

Thanks.

deepmap-marcinr:
Absolutely, I also would like to add Gin support as well.

This code was an experiment, and as such, isn't very modular. I'd like to have a consistent type description available to templates, then allow anyone to add templates for any server. If you go down this path today, it'll get messy. If you're willing, please make a subdirectory in templates for fiber, and be aware that the data passed to the templates isn't always very stable.

kotyara85:
Fiber support would be great. Is anyone working on this?

VandrerenErik:
Hello there.

Are there any prospects on this endeavour? I'm willing to help with the efforts.

princefishthrower:
Also interested in helping out, are there any branches or efforts that have been committed to this repository so far?

Jleagle:
I didn't get far when i started this last year (project requirements changed). Probably better to start from scratch.

Jleagle:
Having a go at this again at https://github.com/Jleagle/oapi-codegen. Looked at the chi server for how to do it but it seems lots of deepmap code requires the standard http libs, which Fiber does not use.

`go generate ./examples/petstore-expanded/fiber/...` to generate.

gnirb:
> Having a go at this again at https://github.com/Jleagle/oapi-codegen. Looked at the chi server for how to do it but it seems lots of deepmap code requires the standard http libs, which Fiber does not use.
> 
> `go generate ./examples/petstore-expanded/fiber/...` to generate.

Nice! I've also done some work on this recently (and ran into similar challenges with the http libs you mentioned). If its to any help for you, my fork is available here https://github.com/four-fingers/oapi-codegen. We've used it for a couple of internal projects so far.

Jleagle:
@gnirb Awesome, thanks. That's very helpful.

PR #427, opened 2021-08-22
#301 Clean up code to support multiple servers

kotyara85:


wolfeidau:
@kotyara85 @deepmap-marcinr this PR has broken the ability to override templates as the command line flag doesn't permit providing a folder. When i try to override the server template for echo I am unable to provide templates nested inside a directory https://github.com/deepmap/oapi-codegen#making-changes-to-code-generation.

Would be good to restore the flat structure rather than nesting them in sub folders.

PR #428, opened 2021-08-23
Allow overriding primitive types with custom types

natsukagami:
## Short Description

Allow primitive types (ints, floats, string types) to be replaced with an option, `primitive-mapping`, in the YAML configuration.

## Motivation

Sometimes it is desirable to use a custom wrappers or different types entirely for the hard-coded types generated in `oapi-codegen`. 
One of such reasons is for tighter controls over JSON marshalling (e.g. check for missing fields or perform some custom validation). 

This can already be done with `x-go-type` extension tag, but it is very painful to do for e.g. every integer parameter in the whole YAML spec.

## What has changed?

This is not a breaking change, no code generation change is made if there is no configuration change.

- A lot of functions now take the `*Options` object as an additional parameter.
- A new option has been added, `primitive-mapping`, which maps the primitive type to the actual type (`int32` => `null.Int32` for example). 
- README has been updated to inform users of the new functionality.

Note that `primitive-mapping` is only available in the YAML configuration at the moment (no flags). I am not sure if it is desirable to add the flags / what is the best way to do it, so I left it out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,6 +542,47 @@ This tells us that in order to resolve references generated from `some_spec.yaml
 need to import `github.com/deepmap/some-package`. You may specify multiple mappings
 by comma separating them in the form `key1:value1,key2:value2`.
 
+### Replacing primitive types
+
+Sometimes you would want to use a different type for the primitives (int, floats etc). 
+This can be done with the `primitive-mapping` option in the YAML configuration file.
+A list of supported primitives is shown below.
+
+If used inside a query parameter, the given replacement should implement `Binder`:
+```go
+type Binder interface {
+    Bind(string) error
+}
+```
+
+| Primitive | Default type used | 
+| -- | -- |
+| `int64` | `int64` |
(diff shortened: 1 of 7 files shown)


jamietanna:
@deepmap-marcinr is this one we can take inspiration from / will be replaced by your work in the v2 side of things?

jamietanna:
Apologies, meant to close #458 🫣

mromaszewicz:
Thank you for doing this @natsukagami . I tried to fix up this commit to be compatible with the current code, which has diverged greatly since when you originally submitted it, but I can't update your PR since the linked branch now contains many other changes, and i don't want to break your work. I've made a new PR, which does the same thing but is implemented differently as #2223 and made you a co-author.

natsukagami:
No worries! It's been a long time since my PR and I don't personally use oapi-codegen anymore, so it's great to have someone take over the changes!

Issue #432, opened 2021-08-25
Client Response template generating Response Parser that has a defer without error check.


tomkaith13:
Hello,

The present client-with-responses template generates a response that does not check errors from a `resp.Body.Close()` that trips up our linter tool called [errcheck](https://github.com/kisielk/errcheck) which complains about the generated file not checking for errors: 
![image](https://user-images.githubusercontent.com/1534550/130813394-9f410c9a-bf7c-4cbe-8f3a-5c4fbe5c0938.png)

Does it make sense to mask such errors using a patch like this: 
![image](https://user-images.githubusercontent.com/1534550/130813751-493fed40-c49b-48b3-a978-94714715f9fb.png)

tomkaith13:
Hi oapi-codegen team, does this issue and fix make sense to you? This is a minimally invasive patch to keep the behavior the same as the original code.

PR #433, opened 2021-08-25
fix: client response template to ignore errors

tomkaith13:
ref: https://github.com/deepmap/oapi-codegen/issues/432 

Presently rsp.Body.Close masks away the error without checking for them.
This code will do the same but give linters the assumption that it alteast
considering to look at them.


deepmap-marcinr:
So, I'm open to merging this, but why does this matter? The error is discarded in either case.

tomkaith13:
hello @deepmap-marcinr, the error generated  is still discarded but it wont trip any errors from [errcheck](https://github.com/kisielk/errcheck) linters that complaint about errors not being checked.  This is a pretty useful linter that checks for errors not being consumed but in this case, we can probably make an exception, no ?

tomkaith13:
Thanks for merging this in @deepmap-marcinr .. we were really hoping we dont have to fork this repo to get around this problem. ❤️

Issue #375, opened 2021-05-30
Invalid types output for enum property inside of array


brumhard:
For the following openapi spec: 
```yaml
openapi: 3.0.3
paths:
  /:
    get:
      responses:
        default:
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/NotWorking"
components:
  schemas:
    NotWorking:
      title: NotWorking
      type: object
      description: This is not working.
      properties:
        anything:
          type: array
          items:
            type: object
            properties:
              testing:
                type: string
                enum:
                  - test1
                  - test2
            required:
              - testing
      required:
        - anything
```

The types that are output are the following: 
```go
// Package main provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/deepmap/oapi-codegen version v1.7.0 DO NOT EDIT.
package main

// Defines values for NotWorkingAnythingTesting.
const (
	NotWorkingAnythingTestingTest1 NotWorkingAnythingTesting = "test1"

	NotWorkingAnythingTestingTest2 NotWorkingAnythingTesting = "test2"
)

// Defines values for NotWorkingAnythingTesting.
const (
	NotWorkingAnythingTestingTest1 NotWorkingAnythingTesting = "test1"

	NotWorkingAnythingTestingTest2 NotWorkingAnythingTesting = "test2"
)

// This is not working.
type NotWorking struct {
	Anything []struct {
		Testing NotWorkingAnythingTesting `json:"testing"`
	} `json:"anything"`
}

// NotWorkingAnythingTesting defines model for NotWorking.Anything.Testing.
type NotWorkingAnythingTesting string

// NotWorkingAnythingTesting defines model for NotWorking.Anything.Testing.
type NotWorkingAnythingTesting string
```

As you can see the `NotWorkingAnythingTesting` enum is defined twice. From testing I found out that this should have something to do with the enum inside of an array. When defining another enum as another property of `NotWorking` it works as expected.

brumhard:
Can confirm that the passed example works now. Thanks @tfaller.

PR #442, opened 2021-09-11
Fix duplicate enum declaration.

tfaller:
This is a fix for #375 
The recursive nature of GetAdditionalTypeDefs caused that types of array items were generated multiple times.
Just simple assign AdditionalTypes in case of an array.


iwata:
When the fixed version will be released?

PR #449, opened 2021-10-03
chore: add linting and fix identified issues

stevenh:
Add support for golangci-lint and fix the issues it identifies.

This switches validation from travis to github actions which is typically faster and more widely supported.

To eliminate formatting issues with the code generated with github.com/cyberdelia/templates, which is deprecated, switched to new core embed package. This required bumping the min golang version to v1.16.

Issues fixed include:
* Add missing error checks.
* goimports format code.
* Remove dead code.
* Fix broken flow for date times in assignPathValues.
* Use constants for repeated literals.
* Add missing doc comments.
* Use generate.go vs doc.go for generation.
* Use nolint hints instead _ assign.
* Remove unneeded type assignments.
* Use errors.As for error type checks to support wrapped errors.
* Use %w for fmt.Errorf error wrapping.
* Use %q instead of '%s' for quoting.
* Remove else statements with lead block which returns to simply flow.
* Use simple if statements.
* Add full stop to end of comments.
* Use lower case for errors.
* Add t.Helper() to test helpers.
* Remove leading and trailing blank lines.
* Use typed strings for context.Values.
* Combine following if statements into one.
* Remove commented out code.
* Convert multi else if to switch statements.


stevenh:
Sorry this is a bit large, but its very hard to split out linting fixes.

You can see the new [github actions in action here](https://github.com/stevenh/oapi-codegen/runs/3784984142?check_suite_focus=true)

Hope this helps.

Ehco1996 (commented):
LGTM

stevenh:
Oh wow forgot I had done this already, now I've done it twice ;-)

PR #453, opened 2021-10-05
chi: Route directly from BaseURL

Karitham:
- [x] Edit chi template to clear off groups
- [x] Use a router.Route prefix based for options.BaseURL
- [x] Add an empty options.BaseURL check


Karitham:
Wrong merge target, sorry about that :/

PR #462, opened 2021-10-09
feat: implement x-oapi-codegen-middlewares (#461)

### LINKED ISSUE #461: Ability to create operation/route-specific middlewares for chi-server.
At the moment, the only way middlewares work are through all-or-nothing middlewares. This can be slightly inhibiting, as sometimes you want to have middlewares for only particular endpoints. 

I have a PR almost ready that addresses this problem.

hhhapz:
This change adds the x-oapi-codegen-middlewares extension tag for the
chi server and a new TaggedMiddleware field to the server interface.
This makes it possible for users to define middlewares that are only
called on specific endpoints, and also allows middlewares to be reused.
The middlewares are called after the global middlewares, and in the
order they are defined.

This change is fully backwards compatible with all older versions of
oapi-codegen, however one disadvantage to this is the silent failure
when a tagged middleware was not found. Perhaps that can be changed in
v2.

Closes #461


thejdavid:
would be very useful great work!

hhhapz:
> would be very useful great work! need any help ?

hey, this has already been implemented and further improved upon in a fork that myself and other people in the discord gophers community are maintaining. You can check it out at https://github.com/discord-gophers/goapi-gen

aaletov:
@hhhapz Did you found some temporary solution?

PR #471, opened 2021-10-26
oneOf/anyOf support (including discriminator)

Warboss-rus:
Implemented using json.RawMessage to delay parsing to concrete type until one of the getter (As...) methods is called. If the schema has a discriminator and its mapping, then ValueByDiscriminator method can be used to get an empty interface that holds one of the union types. Also, when using setters (From... methods), discriminator value will be filled automatically, according to mapping.
This implementation only supports JSON, so will not work in parameters (body only). ~~Also reverts to generating empty interface if "type" or other special properties are specified (for example, if only required properties differ in each variant)~~ fixed that behavior, will now work properly with fixed properties
Tests for this feature are added to internal/test/components.


jonasrmichel:
This is great work @Warboss-rus, thank you!

Any plans to support arrays whose contents are oneOf / anyOf?

For example...

```json
{
   "type":"array",
   "items":{
      "oneOf":[
         {
            "$ref":"#/components/schemas/TypeA"
         },
         {
            "$ref":"#/components/schemas/TypeB"
         },
         {
            "$ref":"#/components/schemas/TypeC"
         }
      ]
   }
}
```

Warboss-rus:
Thanks, added a test for this case, will fix it this week

Warboss-rus:
I've reworked the code to support oneOf\anyOf along fixed properties, fixed type generation for array items that have additional properties or union parts (thanks @jonasrmichel), added some tests and updated information in README

jonasrmichel:
Brilliant work @Warboss-rus, thank you!

@deepmap-marcinr @carmo-evan @Karitham -- It'd be wonderful if these contributions could be merged for the next release. :pray:

tamj0rd2:
Nice! Is there a release schedule? Really looking forward to being able to use this

jonasrmichel:
Nudge.

@deepmap-marcinr these changes look good to me. They include new functionality, tests, and necessary documentation updates.

What's required for merge approval? It'd be great to see these features in the next minor release.

pkg/codegen/schema.go
```diff
@@ -500,6 +535,9 @@ func GenStructFromSchema(schema Schema) string {
 		objectParts = append(objectParts,
 			fmt.Sprintf("AdditionalProperties map[string]%s `json:\"-\"`", addPropsType))
 	}
+	if len(schema.UnionElements) != 0 {
+		objectParts = append(objectParts, "union json.RawMessage")
```
veger, line 539: Thanks a lot for your PR, we use it for quite a while already!
Recently we encountered an issue with it thought, so hopefully you are willing to update the PR, so it won't be an issue when it merged merged anymore:

When using `anyOf` as a type in combination with `additionalProperties` like this:
```
variables:
  type: object
  description: Extra variables of type bool, number, or string
  additionalProperties:
    oneOf:
      - type: boolean
      - type: number
      - type: string
```
([spec](https://swagger.io/specification/#schema-object) states that for `additionalObject` "... inline or referenced schema MUST be of a Schema Object", so this seems to be allowed)

it is not possible to get the actual value, because of the private `union`.

For example the `Get()` function is generated as follows:
```
func (a Variables) Get(fieldName string) (value struct {
	union json.RawMessage
}, found bool) {
	if a.AdditionalProperties != nil {
		value, found = a.AdditionalProperties[fieldName]
	}
	return
}
```
Which returns a `struct` with the private `union` field. Since it comes from a different package, the field is not accessible.

A similar scenario can be imagined when using the `Set()` function (it requires the struct with the private field, which is not accepted as the same type by the compiler)

Also, using `for-range` to iterate over the `Variables.AdditionalProperties` is not possible due to the private field.

Changing `union` to `Union` (and fixing the template (`union.tmpl`) to use `Union`) fixes this issue for me.

Warboss-rus: Thanks, I've added a test case and forced generator to create a new type for additional properties, if this type has union elements or additional properties of its own. Also found and fixed merging union cases with allOf

veger: Thanks a lot, your addition works nicely for our case :+1:

deepmap-marcinr:
Impressive change, thank you.

PR #477, opened 2021-11-02
expand chi server wrapper errors with more information

paulmach:
Support for an error callback function was added to the chi server a few weeks ago https://github.com/deepmap/oapi-codegen/pull/459

This PR expands the errors sent to that callback with the offending parameter name. Existing behavior does not change.

cc: @carmo-evan @Karitham

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,14 @@
 language: go
 go:
-- 1.13.x
+  - 1.16.x
 env:
   global:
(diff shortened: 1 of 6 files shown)


carmo-evan:
@paulmach we ended up doing a hard-fork of this project. It would be awesome if you could make the change there as well https://github.com/discord-gophers/goapi-gen

PR #478, opened 2021-11-09
Unmarshal into a *map instead of a map

alehed:
This allows us to distinguish between `null` and `{}`.

This is a rebasing of #326, which was accidentally closed.


alehed:
This problem should be now fixed by #1404.

PR #487, opened 2021-11-23
Updated labstack/echo/v4 from v4.2.1 to v.6.1 - security vulnerability fix

jaswanthm:
Labstack/echo/v4 v4.2.1 uses a dependency called jwt-go which had security vulnerability. 

The latest version of echo points to the new jwt-go library that has addressed this issue. 

More information can be found here. https://github.com/labstack/echo/blob/c6f0c667f145b5e5347ba812c9de5a5a4280bac5/CHANGELOG.md#v450---2021-08-01


deepmap-marcinr:
This has long ago happened.

Issue #486, opened 2021-11-23
Full XML support not implemented


matyat:
Handling XML responses and requests is severely limited in the generated code as the full OpenAPI spec WRT XML is not fully implemented.

Given this example API definition:
```yaml
openapi: 3.0.3
info:
  version: 1.0.0
  title: test-api

paths:
  /foo:
    get:
      responses:
        200:
          description: OK
          content:
            application/xml:
              schema:
                $ref: "#/components/schemas/document"

components:
  schemas:
    document:
      type: object
      required:
      - name
      - item
      properties:
        name:
          type: string
        item:
          type: object
          required:
          - type
          - value
          properties:
            type:
              type: string
              xml:
                attribute: true
            value:
              type: string
```

The following type is generated:
```go
// Package Testapi provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/deepmap/oapi-codegen version v1.9.0 DO NOT EDIT.
package Testapi

// Document defines model for document.
type Document struct {
        Item struct {
                Type  string `json:"type"`
                Value string `json:"value"`
        } `json:"item"`
        Name string `json:"name"`
}
```

However, there are no `xml` tags present at all. The expected output would be something like:
```go
// Package Testapi provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/deepmap/oapi-codegen version v1.9.0 DO NOT EDIT.
package Testapi

import "encoding/xml"

// Document defines model for document.
type Document struct {
        XMLName xml.Name `xml:"document"`
        Item    struct {
                Type  string `json:"type" xml:"type,attr"`
                Value string `json:"value" xml:"value"`
        } `json:"item" xml:"item"`
        Name string `json:"name" xml:"name"`
}
```

Representing XML in OpenAPI 3: https://swagger.io/docs/specification/data-models/representing-xml/

matyat:
Put together a fairly minimal patch to address this here https://github.com/deepmap/oapi-codegen/pull/488. Let me know if it's of any use :).

kennykarnama:
Hi @matyat, thanks for the patching. I am looking for the same thing. I am testing your patch. Will let you know if there is a problem 😄

Giftzwerg02:
Are there any news to this?

PR #488, opened 2021-11-24
Add support for generating XML tags

matyat:
Add support for the properties "xml/name" and "xml/attribute" detailed
[here][1].
    
[1]: https://swagger.io/docs/specification/data-models/representing-xml/

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -26,6 +26,8 @@ type Schema struct {
 
 	Description string // The description of the element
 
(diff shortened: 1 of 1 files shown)


hae-andrew-young:
This would be very nice to have for us. Please merge this in.

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the parameterized struct tags system provides a general mechanism for adding custom tags, including XML. This PR against 2021-era code is too outdated to merge. There is a tracked issue (#486) noting the XML gap.

PR #436, opened 2021-08-29
Fix problem with code redeclarations

migueleliasweb:
Fixes: #256 #200 and might also fix: #386.

This PR addresses the problems when the generated code contains multiple redeclarations of `methods`, `constants` and `types`.

You can verify this problem with the Docker OpenAPI reference yaml:

```bash
# this converts the yaml from v2 to v3 so we can generate code with oapi-codegen
curl "https://converter.swagger.io/api/convert?url=https://docs.docker.com/engine/api/v1.41.yaml" -H "Accept: application/yaml" -o v1.41.yaml
```

Run the generation with `v1.8.2`:

```bash
oapi-codegen -package openapigen -generate types > types.go
```

Run `staticcheck` linter:

```bash
staticcheck types.go
```

Outputs:

```bash
types.go 
src/openapigen/types.go:247:2: 	other declaration of SwarmSpecCAConfigExternalCAsProtocolCfssl (compile)
src/openapigen/types.go:252:2: SwarmSpecCAConfigExternalCAsProtocolCfssl redeclared in this block (compile)
src/openapigen/types.go:683:6: 	other declaration of ContainerSummary_Labels (compile)
src/openapigen/types.go:688:6: 	other declaration of ContainerSummary_NetworkSettings_Networks (compile)
src/openapigen/types.go:693:6: ContainerSummary_Labels redeclared in this block (compile)
src/openapigen/types.go:698:6: ContainerSummary_NetworkSettings_Networks redeclared in this block (compile)
src/openapigen/types.go:2679:6: 	other declaration of SwarmSpec_CAConfig_ExternalCAs_Options (compile)
src/openapigen/types.go:2685:6: 	other declaration of SwarmSpecCAConfigExternalCAsProtocol (compile)
src/openapigen/types.go:2689:6: SwarmSpec_CAConfig_ExternalCAs_Options redeclared in this block (compile)
src/openapigen/types.go:2695:6: SwarmSpecCAConfigExternalCAsProtocol redeclared in this block (compile)
src/openapigen/types.go:5452:34: 	other declaration of Get (compile)
src/openapigen/types.go:5460:35: 	other declaration of Set (compile)
src/openapigen/types.go:5468:35: 	other declaration of UnmarshalJSON (compile)
src/openapigen/types.go:5490:34: 	other declaration of MarshalJSON (compile)
src/openapigen/types.go:5505:52: 	other declaration of Get (compile)
src/openapigen/types.go:5513:53: 	other declaration of Set (compile)
[...]

PR #493, opened 2021-12-06
fix(response): added an option to override the response suffix

### LINKED ISSUE #386: 'Response' suffix causes duplicated structs
I have some endpoints in my spec like this one:

```yaml
  /v1/auth/phone/login:
    post:
      summary: 用手机登陆/注册
      tags:
        - Auth
      operationId: LoginByPhone
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/LoginWithPhoneRequest'
      responses:
        "200":
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/LoginWithPhoneResponse'
```

If I generate server code it is ok, but when I generate client code `oapi-codegen` creates structs with `Response` suffix which causes deduplication of my structs. 

Maybe it would be better to make this suffix configurable or change it and other suffix/prefix literals to with `_` at the beginning so it would be `OperationName_Response`?

Initially I wanted to use a custom template, but this suffix is generated by the function.

wolfeidau:
This should resolve #386 which is still an issue for clients, the changes in #436 fixed it for types and server but not clients.

This small change enables users to override the suffix where it could clash with existing names.

I have updated the client template to ensure use of the provided suffix.

### DIFF
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -35,28 +35,30 @@ func errExit(format string, args ...interface{}) {
 }
 
 var (
-	flagPackageName    string
-	flagGenerate       string
-	flagOutputFile     string
-	flagIncludeTags    string
-	flagExcludeTags    string
-	flagTemplatesDir   string
-	flagImportMapping  string
-	flagExcludeSchemas string
-	flagConfigFile     string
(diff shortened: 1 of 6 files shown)


leedsalex:
@deepmap-marcinr Any chance this can get pulled into the next release. My team is running into the same issue that this resolves.

Issue #500, opened 2021-12-16
URL Query params should be hinted as "form", not "json"


djeer:
I've run into in issue that my parsed query params are always nil. It was resolved by manually changing query params hint from "json" to "form" in the generated types. The problem may be reproduced using [petstore-expanded.yaml](https://github.com/OAI/OpenAPI-Specification/blob/main/examples/v3.0/petstore-expanded.yaml), generated code (not working):

```
type FindPetsParams struct {
	// tags to filter by
	Tags *[]string `json:"tags,omitempty"`

	// maximum number of results to return
	Limit *int32 `json:"limit,omitempty"`
}
```

code I expect to be generated (working):

```
type FindPetsParams struct {
	// tags to filter by
	Tags *[]string `form:"tags,omitempty"`

	// maximum number of results to return
	Limit *int32 `form:"limit,omitempty"`
}
```

Code I am using to parse query params:
```
func ParseFilter(c *gin.Context) (myapi.FindPetsParams, error) {
	f := tasksapi.FindPetsParams{}
	if err := c.BindQuery(&f); err != nil {
		return f, err
	}
	return f, nil
}
```

djeer:
I've found exact place where it is generated https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/schema.go#L447
```
if p.Required || p.Nullable || !omitEmpty {
	fieldTags["json"] = p.JsonFieldName
} else {
	fieldTags["json"] = p.JsonFieldName + ",omitempty"
}
```

djeer:
Checked on the code from master branch, auto-generated server reads parameters as "form", so it's just annotations that are wrong (annotations are not used by generated server). Created [PR](https://github.com/deepmap/oapi-codegen/pull/501) to fix this. Build passed green.

![image](https://user-images.githubusercontent.com/3585469/146442363-5d13a8c2-b7dc-462e-8116-3fc4c04fa93e.png)

deepmap-marcinr:
The annotation on those fields isn't related to the parameter type in BindQueryParameter. There is something missing in the problem description. By changing the annotation to `form:"..."`, you are changing the way JSON is parsed, so perhaps that's changing things up in a way that make them work.

deepmap-marcinr:
Oh, I see what's going on.

We need both annotations, then, both `form` and `json` on those objects.

Perhaps we add the `form` annotation for gin.

djeer:
[...]

PR #501, opened 2021-12-16
Query Params struct tag should be "form" instead of "json"

djeer:
Fix for issue https://github.com/deepmap/oapi-codegen/issues/500
Query params will be generated with tag "form" instead of tag "json"

### DIFF
--- a/examples/petstore-expanded/chi/api/petstore.gen.go
+++ b/examples/petstore-expanded/chi/api/petstore.gen.go
@@ -48,10 +48,10 @@ type Pet struct {
 // FindPetsParams defines parameters for FindPets.
 type FindPetsParams struct {
 	// tags to filter by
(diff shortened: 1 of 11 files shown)


djeer:
image

djeer:
Checked on the code from master branch, auto-generated server reads parameters using "form" argument, so it's type hints that are wrong  
image

deepmap-marcinr:
Sorry to respond to this so late. I'm happy to merge this - but can we have both annotations, 'json' and 'form'? There's lots of existing JSON serialization which will break.

djeer:
Sorry for late response as well 😄 Looks like this PR can be closed since the fix was applied https://github.com/deepmap/oapi-codegen/commit/6a3482493b5731b60c267147522ec3aaa09fc3a3

PR #505, opened 2022-01-05
GIN Middelware: return correct http status codes based on validation

msvechla:
Hello, 

Currently, the gin middleware always returns HTTP status `400`, even for unauthorized requests.
This PR fixes this, by returning a different status code based on the validation results.

Tests have been aligned as well.

Thanks a lot, let me know if you have further suggestions


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the Gin middleware code has been extracted to a separate repository (`github.com/oapi-codegen/gin-middleware`). This PR targets code that no longer lives in the main repo.

PR #506, opened 2022-01-07
Fix customProvider RequestEditorFn example

reidmv:
The ctx and req arguments in the example appear to be ordered incorrectly.
This commit modifies the example to list ctx first, which appears to be the
correct ordering.

I'm just learning golang so I haven't dug deeply to confirm this. Just noted
that the example wasn't working as-is in my test project, but compiles
when I invert them, and ctx documentation says it should usually come first.


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the `RequestEditorFn` definition is correct in the current codebase. Either the fix was applied or the code was always correct.

PR #515, opened 2022-01-31
Recursively load subdirectory template files

shearnd:


kandreev:
Works for me, thank you!

Issue #468, opened 2021-10-20
Custom struct fieldname 


octomad:
This is a feature request to support custom struct field names.

We have a situation where we are using `allOf` to reference a JSON schema that has a property with a similar name. After the name gets sanitized for Go, it ends up duplicating a field name that already exists.

Example:

**parentschema.json**
```json
{
  "title": "node",
  "type": "object",
  "description": "Represents a node",
  "properties": {
    "@id": {
      "type": "string",
      "readOnly": true
    }
}
```

**components in the OpenAPI Spec**
```yaml
components:
  schemas:
    resourcenode:
      title: resourcenode
      allOf:
        - $ref: ./parentschema.json
        - type: object
          properties:
            id:
              type: string
              format: uuid
              description: ID of the resource
              readOnly: true
```

Because there's a property in the parentschema.json called `@id`, and a property in the OpenAPI schema called `id`, the generated code will output

```go
type Resourcenode struct {
	Id string `json:"@id"`
	Id string `json:"id"`
}
```

Which obviously won't compile because of the duplicate field in the struct. If we could add support for custom field names, that would help.

An alternative approach would be to add an X-Extension for `x-go-fieldname-prefix` to prefix the property so that it generates this instead

```go
type Resourcenode struct {
	NodeId string `json:"@id"`
	Id string `json:"id"`
}
```

octomad:
I don't mind implementing the change if you give me some pointers.

emilekm:
Hey @octomad, I've implemented the change you request - please take a look and tell me if it satisfies you and if it works for your case.

octomad:
SHIP IT! :shipit:

mromaszewicz:
This has been implemented via the x-go-name extension (commit dcb982c, Apr 2022). You can now override Go struct field names to avoid collisions from sanitization. Example: annotate @id with x-go-name: NodeID. See examples/extensions/xgoname/ and the README documentation. Closing as completed.

PR #517, opened 2022-02-04
feat: Support `x-go-name` property extension

emilekm:
This implements a feature requested in https://github.com/deepmap/oapi-codegen/issues/468.

The extension name `x-go-name` comes from [go-swagger](https://goswagger.io/use/spec.html#known-vendor-extensions).

Example schema:
```yaml
components:
  schemas:
    Person:
      type: object
      properties:
        id:
          type: string
          x-go-name: ID
        pet_id:
          type: string
          x-go-name: PetID
```
Output:
```go
type Person struct {
	ID string `json:"id"`
	PetID string `json:"pet_id"`
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,10 @@ which help you to use the various OpenAPI 3 Authentication mechanism.
   will override any default value. This extended property isn't supported in all parts of
   OpenAPI, so please refer to the spec as to where it's allowed. Swagger validation tools will
   flag incorrect usage of this property.
+- `x-go-name`: specifies Go field name. It allows you to specify the field name for a schema, and
+  will override any default value. This extended property isn't supported in all parts of
+  OpenAPI, so please refer to the spec as to where it's allowed. Swagger validation tools will
+  flag incorrect usage of this property.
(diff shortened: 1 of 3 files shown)


octomad:
Excellent. Exactly what was needed.

octomad:
Just checking in. Will this be merged soon?

Penthious:
Also checking when this will be merged in.

Jacob-Roberts:
This has broken the generated code. The `types` file that is created will correctly respect this. But the client that is generated will use the non-x-go-name version

emilekm:
> This has broken the generated code. The `types` file that is created will correctly respect this. But the client that is generated will use the non-x-go-name version

Could you provide a sample of the code that is broken? I don't think it's broken and you are referring to URL parameters (since query params work fine). If I am right, then [this MR](https://github.com/deepmap/oapi-codegen/pull/574) should fix this for you (the param ref name will be used instead of param's `name`)

emilekm:
I'm currently on vacation, but when I get back I'll try to fix this (unless
it's already being worked on). Good catch

Issue #507, opened 2022-01-11
Support RouterGroup when registering handlers with gin


msvechla:
Hello, 

Currently, it's not possible to register the handlers on a gin `RouterGroup`. 
This could be helpful, because one might register the openapi handlers on a router group with special middleware configured, with different middleware on some other `RouterGroup`.

One possible way would be changing the `*gin.Engine` to `*gin.RouterGroup` in `RegisterHandlers`: https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/templates/gin/gin-register.tmpl#L8.

However, this would be a breaking change. I am open to any other solutions on this.

I also thought about using the `options.Middlewares` to specify special middleware used only with the routes from the OpenAPI spec, however, this one does not support the regular `gin.Handlerfunc` usually used with middleware.

Do you have any suggestions here?
Thanks a lot!

metinogurlu:
I have same issue. Gin's grouping function totally usesles because of oapi's definition. @msvechla could you find any solution?

msvechla:
The code generation allows you to supply template files. I just copied over the Gin-Register.tmpl and made the changes I mentioned above. During template generation I then specified my modified file as template, so the generated code has the changes

Issue #485, opened 2021-11-17
gin: included middleware cannot abort request


erikbos:
For Gin generated code, when supplying middleware:

```
	RegisterHandlersWithOptions(router, handler, GinServerOptions{
		Middlewares: []MiddlewareFunc{
			auth.AuthenticateAndAuthorize,
		},
	})
```

A request cannot be abort by middleware, the request is fully being handled.. which mean authentication middleware cannot abort a request for example.

(@stephen-opal noticed the issue in initial pull request: https://github.com/deepmap/oapi-codegen/pull/385#discussion_r655001461)

tomsharratt:
Any chance this will be getting resolved?

re-cheid:
For the time being you can just take my modified gin template files from #926 and use them as custom templates during generation as described here: https://github.com/deepmap/oapi-codegen#making-changes-to-code-generation

I just put them in my project directory:
![image](https://user-images.githubusercontent.com/25147494/212836295-1c555a73-cfed-4f7b-a3ce-4fe0404ea62f.png)

Then I call `oapi-codegen -generate gin -templates scripts/oapi-codegen-templates/ -o "gen/openapi/openapi_api.gen.go" -package "rest_service" "api/openapi/rest_service.yaml"`

alexstojda:
This should be resolved by #914, @deepmap-marcinr probably good to close this :)

mromaszewicz:
This has been fixed. The Gin wrappers template now checks c.IsAborted() after each middleware call and returns early if the request was aborted. This allows authentication middleware and similar to properly stop request processing. Closing as completed.

PR #530, opened 2022-02-22
Improve gin support (interface, error, middlewares)

zzh8829:
1. Use gin.IRouter interface instead of *gin.Engine
    this will allow us to support nested router mounting using engine/routergroup/other routers
    fixes https://github.com/deepmap/oapi-codegen/issues/507 
2. switch custom middleware code to gin middleware
    the custom middleware code doesn't support context abort or c.Next() in normal gin middleware, this also fixes https://github.com/deepmap/oapi-codegen/pull/494  https://github.com/deepmap/oapi-codegen/issues/485
3. add errors from validation middleware to gin context
    huge fix for debugging authentication func. before this change the nested e.Errors[] is completely hidden not surfaced anywhere.

### DIFF
--- a/pkg/codegen/templates/gin/gin-register.tmpl
+++ b/pkg/codegen/templates/gin/gin-register.tmpl
@@ -1,21 +1,26 @@
 // GinServerOptions provides options for the Gin server.
 type GinServerOptions struct {
     BaseURL string
-    Middlewares []MiddlewareFunc
+    Middlewares []gin.HandlerFunc
 }
 
+// keeping for backward compatibility
+type MiddlewareFunc = gin.HandlerFunc
+
 // RegisterHandlers creates http.Handler with routing matching OpenAPI spec.
-func RegisterHandlers(router *gin.Engine, si ServerInterface) *gin.Engine {
+func RegisterHandlers(router gin.IRouter, si ServerInterface) gin.IRouter {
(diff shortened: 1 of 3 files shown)


ericvolp12:
Hey, this is neat but it ended throwing off how we were constructing the gin engine, looks like now the pattern needs to be:
```go
// Register the business logic handlers
apiRoutes := api.RegisterHandlers(r, apiServer).(*gin.RouterGroup)

r.Use(apiRoutes.Handlers...)
```

Instead of what was previously:
```go
// Register the business logic handlers
r := api.RegisterHandlers(r, apiServer)
```

I'm gonna submit a MR to update the Gin example since this new patch breaks the example.

jxsl13:
returning concrete types should be preferred over returning interfaces as concrete types that implement an interface may provide more functionality than the interfave.
 on the other hand requiring interfaces to be passed as parameters makes parameters more versatile.

deepmap-marcinr:
huh, yeah. Agreed, returning concrete types is better. I missed that. I wonder whether to roll back this change. I hate breaking existing code like this.

ericvolp12:
We could update it to return a `gin.RouterGroup` instead of the `IRouter` but I'm not 100% sure if there are any use cases in which we'd want an `IRouter` over the `RouterGroup`.

jxsl13:
release a new monor version and retract the last one maybe in go.mod?

Issue #531, opened 2022-02-23
Required inheritance and generated objects when using allOf


florentchauveau:
Considering the following schema:

```yaml
    FooProperties:
      type: object
      properties:
        uuid:
          type: string
          format: uuid
          readOnly: true
        firstname:
          type: string
        lastname:
          type: string

    Foo:
      allOf:
        - $ref: "#/components/schemas/FooProperties"
        - type: object
          required:
            - uuid
            - firstname
            - lastname

    NewFoo:
      allOf:
        - $ref: "#/components/schemas/FooProperties"
        - type: object
          required:
            - firstname
```

According to the specs, also confirmed in the https://editor.swagger.io/, and https://github.com/swagger-api/swagger-editor/issues/1212, `NewFoo` should have `firstname` as required, and `Foo` should have all properties as required.

The validation works as expected.

However, the resulting objects in Go are:

```go
type Foo struct {
	// Embedded struct due to allOf(#/components/schemas/FooProperties)
	FooProperties `yaml:",inline"`
	// Embedded fields due to inline allOf schema
}

// FooProperties defines model for FooProperties.
type FooProperties struct {
	Firstname *string `json:"firstname,omitempty"`
	Lastname  *string `json:"lastname,omitempty"`
	Uuid      *string `json:"uuid,omitempty"`
}

// NewFoo defines model for NewFoo.
type NewFoo struct {
	// Embedded struct due to allOf(#/components/schemas/FooProperties)
	FooProperties `yaml:",inline"`
	// Embedded fields due to inline allOf schema
}
```

Because of the struct embedding, the properties are all pointers instead of values (like when using "required" directly in objects).

Now, because the validation works as expected, this might not be an issue at all. I wonder however if in this case, if the generated code should embed `FooProperties` as is, or if a new struct should be generated per "merge".

This has been discussed, and sometimes fixed, in several other projects:
- https://github.com/OAI/OpenAPI-Specification/issues/1870
- https://github.com/cyclosproject/ng-openapi-gen/issues/108
- https://github.com/sylvainlaurent/swagger-validator-maven-plugin/issues/19
- https://github.com/acacode/swagger-typescript-api/issues/275
- https://github.com/p1c2u/openapi-schema-validator/issues/15

deepmap-marcinr:
This is a hard one, and not a part of the spec that I was aware of when I was initially writing this code.

[...]

Issue #502, opened 2021-12-17
Generating code for "anyOf"/"allOf" with only one possible ref type


lkarlslund:
I'm using oapi-codegen for generation of the Microsoft Graph API, and there are many places they use this declaration:

````
"optionalClaims": {
  "anyOf": [
    {
      "$ref": "#/components/schemas/microsoft.graph.optionalClaims"
    }
  ],
  "description": "Application developers can configure optional claims in their Azure AD applications to specify the claims that are sent to their application by the Microsoft security token service. For more information, see How to: Provide optional claims to your app.",  "nullable": true
},
````

This results in an interface{} type in the output code. I know there are efforts to improve the oneOf, anyOf options, and a pull request is waiting for merge (and so am I).

But in all cases wouldn't anyOf / oneOf / allOf with only ONE provided option be the same as just having a $ref directly, and shouldn't the code then output that single type?

I'm using the JSON from https://graphexplorerapi.azurewebsites.net/openapi?operationIds=*&openApiVersion=3&graphVersion=beta&format=json - it has a few problems with unsupported datatypes, which I search&replace by hand:

"format": "decimal" -> "format": "double"
{:space:}NaN -> {:space:}"NaN MS BORK"

Generating the client code for this seems syntactically correct, but the Go compiler was overwhelmed when I tried to compile it, so I'm using my own client code and the generated datatypes. My code is in https://github.com/lkarlslund/azureimposter for reference.

mromaszewicz:
Sorry for the long silence on this one.

The code for handling `oneOf`, `anyOf` has changed radically since this issue was filed. It still doesn't detect the pattern of a single element `anyOf` or `oneOf`, and automatic squashing will break existing users. We could potentially add an output option to flatten single element aggregates.

PR #537, opened 2022-03-08
POC for flattening allOf

alex-broad:
Based on a need I have, and it seems others have too (#531, #502)
I thought I'd see what the appetite is for a change that looks something like this being accepted back upstream. What it does now is it flattens any allOf structs, either by referencing the struct directly if there is only one ref and no extra props, or by flattening and pulling in the fields directly into the current struct.

Still to do:
- [ ] Add mechanism to switch flattening on/off - possibly by specifying a spec extension? `x-oapi-codegen-flatten-allof: true` perhaps? 
- [ ] Add tests

Looking for suggestions and thoughts on whether this could be accepted upstream - thanks in advance.

### DIFF
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -368,6 +368,10 @@ func GenerateConstants(t *template.Template, ops []OperationDefinition) (string,
 // dereference schemas from another part of the tree.
 var Schemas = make(map[string]*openapi3.SchemaRef)
 
+// Even worse for now, we need to have a list of additional schemas to go
+// create to avoid anonymous structs
+var AdditionalSchemas = make(map[string]*openapi3.SchemaRef)
+
 // Generates type definitions for any custom types defined in the
 // components/schemas section of the Swagger spec.
 func GenerateTypesForSchemas(t *template.Template, schemas map[string]*openapi3.SchemaRef, excludeSchemas []string) ([]TypeDefinition, error) {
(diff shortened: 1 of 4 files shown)


pkg/codegen/codegen.go
```diff
@@ -363,24 +363,28 @@ func GenerateConstants(t *template.Template, ops []OperationDefinition) (string,
 	return GenerateTemplates([]string{"constants.tmpl"}, t, constants)
 }
 
+// Not the best, but let's do it this way for now.
+// Define a package variable that holds the Schemas to allow us to
+// dereference schemas from another part of the tree.
+var Schemas = make(map[string]*openapi3.SchemaRef)
```
alex-broad, line 369: This is obviously a considerable change from the existing philosophy, but something is required to allow us to find any other Schema to dereference it. One thing I could do is wrap this in accesor functions rather than just make a package-wide variable. Thoughts?

pkg/codegen/schema.go
```diff
@@ -541,29 +541,83 @@ func MergeSchemas(allOf []*openapi3.SchemaRef, path []string) (Schema, error) {
...
-			// }
 			goType, err := RefPathToGoType(ref)
 			if err != nil {
 				return "", err
 			}
-			objectParts = append(objectParts,
-				fmt.Sprintf("   // Embedded struct due to allOf(%s)", ref))
-			objectParts = append(objectParts,
-				fmt.Sprintf("   %s `yaml:\",inline\"`", goType))
+			if true {
```
alex-broad, line 596: Obviously this needs to switch based on something, perhaps an extension e.g. `x-oapi-codegen-flatten-allof: true`

deepmap-marcinr:
I did this completely differently, please check the latest commits and see how they work for you. There are more complexities with AllOf than just merging the schemas in our own abstraction, because a schema in an AllOf also might be an AllOf, and you need it to be recursive.

I think we can close this PR if the latest code works for you

alex-broad:
I will, thank you for your attention!

PR #539, opened 2022-03-13
538 add default query param support

wjase:
Added support in the Schema object for the openapi schema default value.
There is now a boolean and a string which can be accessed from the templates.
See the echo server template change for exmaple usage.

Now we can specify a default value and it will be reflected in the request passed to the handler if it is not overridden in the request params


wjase (commented):
ok something weird is going on here - url.Values has a Has method. I'm missing something

wjase:
The build failure doesn't make sense at first glance and I can't reproduce locally

PR #540, opened 2022-03-13
538 default param values

wjase:
Fix for issue 538


wjase:
Found it. go1.17 url has a Has() method - go1.16 doesn't

wjase:
Hi @deepmap-marcinr Just wondering what needs doing to have this merge ready?

deepmap-marcinr:
The reason we don't have this yet is because it's difficult. What it the default value is an object, not a simple string/primitive?

With the code as-is, you'll get errors processing those OpenAPI specs.

wjase:
Hi There, Thanks for the feedback - I'll test drive a fix to address your comments.

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because default value support via `ApplyDefaults()` methods has been implemented elsewhere in the codebase. This 2022 PR is too outdated to merge — any work would need to be done fresh.

Issue #480, opened 2021-11-11
Build error in generated gin-server code: "undefined r"


yamamoto-febc:
version: 

```bash
$ oapi-codegen -version
github.com/deepmap/oapi-codegen/cmd/oapi-codegen
v1.9.0
```

spec:

```yaml
openapi: "3.0.0"
paths:
  /foo:
    get:
      summary: foo
      operationId: foo
      parameters:
        - name: X-Foo
          in: header
          required: true
          schema:
            type: string
            enum:
              - bar
      responses:
        '200':
          description: example
          content:
            application/json:
              schema:
                type: object
                required:
                  - id
                properties:
                  id:
                    type: integer
                    format: int64
```

generated by: 

```bash
$ oapi-codegen -generate types,client,gin -o example.go example.yaml
```

After the code is generated, `go build` fails with `./example.go:283:13: undefined: r`

PR #541, opened 2022-03-17
Fix loadTemplateOverrides function. It must manage subfolder to allow…

maurik77:
Improvements made to the loadTemplateOverrides function. 
Now It manages subfolder to allow the override of server template files like 'gin'.
Without the fix It is impossible to override the server template files because, for instance, the 'gin' manager looks for the following keys in the templates map:
- "gin/gin-interface.tmpl"
- "gin/gin-wrappers.tmpl"
- "gin/gin-register.tmpl"

In addition the pull request contains the fix related to https://github.com/deepmap/oapi-codegen/issues/480

### DIFF
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -174,20 +174,43 @@ func loadTemplateOverrides(templatesDir string) (map[string]string, error) {
 		return templates, nil
 	}
 
-	files, err := ioutil.ReadDir(templatesDir)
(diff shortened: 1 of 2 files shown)


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because user template overrides with subfolder paths (e.g., `gin/gin-interface.tmpl`) already work in the current codebase. The template loading code properly handles nested paths.

Issue #239, opened 2020-10-14
Improvement : oapi validation middleware should only validate path/route/endpoint that included in oapi spec file.


bandirsen:
In current implementation, oapi validation middleware validate all route/path of the server application including path/route that wasn't included in oapi spec file, 

I have a case when my application is serving oapi request but also serve static files from a webroot (I'm using echo framework)
```
e := echo.New()
e.Static('/', "path_to_static_files_dir")
...
e.Use(OapiRequestValidator(swagger))   
```
The middleware will reject request to static files such as http://host/index.html or http://host/javascript/any.js or http://host/image/any.png etc and send 400 error "Path was not found", 

In my opinion those requests should not validated by the middleware, since was not oapi end-point. I knew we could use echo middleware Skipper, but it will be tedious job if we have many static files/folder in webroot directory.

My suggestion is if the middleware can not find a matching route from a request, it should letting the upstream application code  to handle it.

Anyone agree with me ?

iggy:
Did you ever find a workaround for this? My use case is metrics endpoints. I'm trying to use the echo Prometheus middleware, but the validator keeps blocking `/metrics` requests. I kind of sort of worked around it by writing a custom `Skipper` function. That feels like it could get cumbersome with more and more exceptions.

```go
	validatorOptions.Skipper = func(ctx echo.Context) bool {
		if ctx.Path() == "/metrics" {
			return true
		}

		return false
	}
	e.Use(middleware.OapiRequestValidatorWithOptions(swagger, validatorOptions))
```

aliakseiz:
Using `e.Group` solved similar issue for me
```go
e := echo.New()
noCodegen := e.Group("")
// Add non-codegen handlers here

codegen := e.Group("")
// Add codegen handlers and middleware here.
```

Full example:
```go
// Add non-codegen handlers here
noCodegen := e.Group("")
noCodegen.Static("/fonts", "fonts")
noCodegen.Static("/css", "css")

// Add codegen handlers and middleware here
codegen := e.Group("")

// Instead of setting swagger.Servers to nil, set it to the base path similar to router
// https://github.com/deepmap/oapi-codegen/issues/806
swagger.Servers = []*openapi3.Server{{URL: "/api", Description: "Needed to match path"}}
// Use our validation middleware to check all requests against the OpenAPI schema.
codegen.Use(middleware.OapiRequestValidator(swagger))

// We now register our store above as the handler for the interface
[...]

PR #552, opened 2022-04-18
Add an option to skip route error

k-yomo:
## About
- Add an option to skip route error to let users have flexibility to define the routes outside oapi.
- Return route error to let users handle it.

Fix https://github.com/deepmap/oapi-codegen/issues/239

### DIFF
--- a/pkg/middleware/oapi_validate.go
+++ b/pkg/middleware/oapi_validate.go
@@ -63,6 +63,7 @@ type Options struct {
 	ParamDecoder openapi3filter.ContentParameterDecoder
 	UserData     interface{}
 	Skipper      echomiddleware.Skipper
+	SkipRouteError bool
 }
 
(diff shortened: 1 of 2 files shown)


deepmap-marcinr:
This is already easy to do, so I'm wondering whether this is necessary. We try not to complicate our code when there's already a way to do it.

```
e := echo.New()
noCodegen := e.Group("")
// Add non-codegen handlers here

codegen := e.Group("")
// Add codegen handlers and middleware here.
```

If you would like to modify the middleware to skip routes, let's do this with a skipper function, which allows for arbitrary skipping configuration.

k-yomo:
@deepmap-marcinr 
Thanks for the comment!  Actually, my usecase is I would like to handle `not found path` as `404`, and it can't be achieved by your approach.

If skipping route error is not a good way, I would like at least expose internal route error so we can handle it.

PR #553, opened 2022-04-19
Update dependencies in `go.mod'

seeforschauer:


seeforschauer:
Hi @deepmap-marcinr I would love to see this merged, what are your thoughts? It's currently blocking me bcz protobuf is used in https://github.com/tigrisdata/tigrisdb and I've panics in using https://github.com/tigrisdata/tigrisdb 

```panic: proto: file "server/v1/api.proto" is already registered
	previously from: "github.com/tigrisdata/tigrisdb-client-go/api/server/v1"
	currently from:  "github.com/tigrisdata/tigrisdb/api/server/v1"
See https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict
```
In the new version of protobuf it's fixed.

PR #554, opened 2022-04-19
Update gin example to use most recent release changes

ericvolp12:
The most recent release of `oapi-codegen` (v1.10.0) included changes
that broke the existing examples which were introduced in https://github.com/deepmap/oapi-codegen/pull/530

This change updates the `gin` example by regenerating the code-generated files and
updating the router registration to work with the new generated code.

It's worth considering whether we should update the `gin` template to return a concrete
type instead of the `IRouter` interface.

When the `IRouter` interface was introduced it seemed to exist just to allow mocking of the
interface (https://github.com/gin-gonic/gin/pull/1607) so I think it may be reasonable to just return 
a `gin.RouterGroup` from the `RegisterHandlersWithOptions` and `RegisterHandlers` generated functions.


ericvolp12:
This example was https://github.com/deepmap/oapi-codegen/pull/530 which has since been rolled back, I don't think this should have been merged unless we decide to re-roll out the changes in https://github.com/deepmap/oapi-codegen/pull/530 with a version bump that makes sense since they are breaking.

deepmap-marcinr:
Thank you.

PR #555, opened 2022-04-19
Return a struct from `gin`'s RegisterHandlers instead of an interface

jazware:
This change provides an alternative to https://github.com/deepmap/oapi-codegen/pull/554 based on the discussions in https://github.com/deepmap/oapi-codegen/pull/530.

For setting up a Gin engine, we want to handle the `IRouter` interface when registering handlers, but as this only thinly wraps a `*gin.GroupRouter`, we can return a `*gin.GroupRouter` from the registration function.

We _could_ refactor the `RegisterHandlers` and `RegisterHandlersWithOptions` to just take a `*gin.GroupRouter` to begin with, I'm not really sure what we gain by using the `IRouter` interface here. Does anyone have opinions on doing that?

### DIFF
--- a/examples/petstore-expanded/gin/api/petstore-server.gen.go
+++ b/examples/petstore-expanded/gin/api/petstore-server.gen.go
@@ -36,11 +36,8 @@ type ServerInterface interface {
 
 // ServerInterfaceWrapper converts contexts to parameters.
 type ServerInterfaceWrapper struct {
-	Handler            ServerInterface
-	HandlerMiddlewares []MiddlewareFunc
+	Handler ServerInterface
 }
-
-type MiddlewareFunc func(c *gin.Context)
 
 // FindPets operation middleware
 func (siw *ServerInterfaceWrapper) FindPets(c *gin.Context) {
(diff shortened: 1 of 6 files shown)


ericvolp12:
I included autogeneration of new examples via `go generate` directives in the `gin` example folder that were missing which is likely what lead to this breaking change being merged without notice. If the changes make the generated code change in a breaking way, it will fail in `make test` now when it tries to build the example packages. `make test` runs after `make generate` so we should catch breaking changes this way.

mromaszewicz:
Thank you for submitting this PR, @jazware — but we need to close it. Goal (1) — accepting `gin.IRouter` instead of `*gin.Engine` — was fully achieved by #879 (commit `e8dc342d`). Goal (2) — returning a concrete `*gin.RouterGroup` instead of `*gin.Engine` — was resolved differently: the maintainers ultimately chose to make `RegisterHandlers` void, a different design decision. So while the input-type concern is fixed, the return-type direction wasn't taken.

Apologies for the long wait, and thanks for the contribution.

PR #556, opened 2022-04-21
Swap UUID format to byte array

rliebz:
With #546 merged in, UUIDs are defined as an alias for a string.

Before that implementation gets mass adoption, I wanted to propose a
slightly different alternative: basing the custom type off of
`uuid.UUID` over `string`.

This has the advantages of:
- Being convertable back to a `uuid.UUID` without having to parse the
  string a second time, or opening the conversion up to failure
  scenarios
- Still being easily convertable to `string` with the `uuid.UUID`'s
  `String` method
- Taking up less space in memory
- Reducing the number of invalid representations, since only `[16]byte`s
  can be represented, as opposed to all strings
- Not adding any dependencies, since `uuid.UUID` is already being
  imported

### DIFF
--- a/pkg/types/uuid.go
+++ b/pkg/types/uuid.go
@@ -7,13 +7,10 @@ import (
 	"github.com/google/uuid"
 )
 
-type UUID string
+type UUID uuid.UUID
 
 func (u UUID) MarshalJSON() ([]byte, error) {
-	if _, err := uuid.Parse(string(u)); err != nil {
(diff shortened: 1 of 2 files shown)


alexdulin:
@deepmap-marcinr any possibility of a review on this? This would substantially lessen the burden on users to need to implement custom methids for marshaling/unmarshaling actual UUID types and greatly improve the usability of this project.

deepmap-marcinr:
You're right, thanks. Odd, the tests were passing.

PR #557, opened 2022-04-23
Add ErrorHandler for middleware

andrewstucki:
This adds roughly the same custom error handling logic across all three openapi validation middleware. For the particular API that I'm working with I want to standardize error messages across all API endpoints and responses originating from middleware. Since the current middleware make opinionated decisions around how errors coming from the validation middleware are sent, the thought is just to add an `ErrorHandler` option (similar to `ErrorHandlerFunc` in the generated server code) that allows you to do some custom error formatting.

### DIFF
--- a/pkg/middleware/oapi_validate.go
+++ b/pkg/middleware/oapi_validate.go
@@ -29,8 +29,10 @@ import (
 	echomiddleware "github.com/labstack/echo/v4/middleware"
 )
 
-const EchoContextKey = "oapi-codegen/echo-context"
-const UserDataKey = "oapi-codegen/user-data"
+const (
+	EchoContextKey = "oapi-codegen/echo-context"
+	UserDataKey    = "oapi-codegen/user-data"
+)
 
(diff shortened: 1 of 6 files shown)


pkg/middleware/oapi_validate.go
```diff
@@ -90,7 +99,7 @@ func OapiRequestValidatorWithOptions(swagger *openapi3.T, options *Options) echo
 
 // ValidateRequestFromContext is called from the middleware above and actually does the work
 // of validating a request.
-func ValidateRequestFromContext(ctx echo.Context, router routers.Router, options *Options) error {
+func ValidateRequestFromContext(ctx echo.Context, router routers.Router, options *Options) *echo.HTTPError {
```
andrewstucki, line 102: switched the return type here since it's only called from the above function, all branches return `*echo.HTTPError` types, and returning the concrete type rather than an `error` allows for easier introspection in the `ErrorHandler`.

Issue #545, opened 2022-03-30
Support shared "types" package


kwtj43:
Sorry if this feature (or work around) is already available, I didn't find anything in the documentation.

I'd like to be able to generate the types in one package and then use that package for client/server (that also reside in different packages)...

oapi-codegen -package types -generate types my.yaml > types/types.gen.go
oapi-codegen -package client -generate client **-type-package types** my.yaml > client/client.gen.go
oapi-codegen -package server -generate server **-type-package types** my.yaml > server/server.gen.go

The generate client/server code would use the types in the "types" package.

tuxofil:
That's a must have feature. In other case typical cyclic deps arose:
* `server` pkg contains implementation that imports `tools` pkg;
* `tools` pkg has to import `server` in order to process arbitrary types.
Separate `types` package could break that vicious circle.

tuxofil:
https://github.com/deepmap/oapi-codegen/pull/389

tuxofil:
I've adapted #389 to the latest version of `oapi-codegen` in #558

tuxofil:
The difference from origin patch is you have to specify full Golang package name like:
```
oapi-codegen -package types -generate types my.yaml > types/types.gen.go
oapi-codegen -package client -generate client -type-package a.com/b/c/types my.yaml > client/client.gen.go
oapi-codegen -package server -generate server -type-package a.com/b/c/types my.yaml > server/server.gen.go
```

pitmullerIngka:
This is now possible.
An example is provided in one of the original PRs for this: https://github.com/deepmap/oapi-codegen/pull/389#issuecomment-1243080859

pcriv:
Is there a way to provide additional-imports via cli?

mromaszewicz:
This is now supported via import-mapping. You can generate types in one package and reference them from client/server code in separate packages using external $ref references and the import-mapping configuration. A complete working example is provided in examples/import-mapping/multiplepackages/. Closing as resolved.

PR #558, opened 2022-04-24
Types can be placed separately from the server side logic

tuxofil:
For https://github.com/deepmap/oapi-codegen/issues/545


mromaszewicz:
Thanks for doing this. I'm revisiting old PR's and this functionality is now fully present, implemented in a different way.

PR #561, opened 2022-04-25
Types can be placed separately from the server side logic

### LINKED ISSUE #545: Support shared "types" package

chrnorm:
Extends upon #558 to close #545.

I was using @tuxofil's fork locally but I found that the empty strings being passed to `GenerateGoSchema` ([here, as an example](https://github.com/deepmap/oapi-codegen/pull/558/commits/34f2cf042d5d2601007663e10e886f6f24a3a016#diff-78dc9f981f29394bfdbfac801231649f4119db8a57213336a1e4d0eaf2cdf6aaR442)) caused the generated client code to not correctly pick up the types from the `types` package.

### DIFF
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -43,6 +43,7 @@ var (
 	flagTemplatesDir   string
 	flagImportMapping  string
 	flagExcludeSchemas string
+	flagTypesPackage   string
 	flagConfigFile     string
 	flagAliasTypes     bool
 	flagPrintVersion   bool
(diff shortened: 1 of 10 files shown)


deepmap-marcinr:
Take a look at an alternate way of doing this. We already _mostly_ support moving types out, all that was missing was the ability to configure addition al imports.

You can always:

```
import (
  .  <path to models>
)
```

This allows importing your models without modifying the code dependent upon them. See the authenticated-api example for how that works in this PR:
https://github.com/deepmap/oapi-codegen/pull/595

PR #563, opened 2022-05-02
Broken json marshal/unmarshal for nested additional properties

johanneswuerbach:
json marshal/unmarshal is broken for nested additional properties like:

```yaml
nested:
  description: Has nested additional properties
  type: object
  additionalProperties:
    additionalProperties: true
    type: "object"
```

I've looked into fixing this issue, but I'm not sure on the best approach. One option would be to modify the https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/templates/additional-properties.tmpl#L63 to handle such nested additionalProperties. Another one to generate named structs also for such cases with their own marshalling / unmarshalling function. 

Do you have any preference or a different idea?

### DIFF
--- a/internal/test/components/components.gen.go
+++ b/internal/test/components/components.gen.go
@@ -60,6 +60,13 @@ type AdditionalPropertiesObject5 struct {
 	AdditionalProperties map[string]SchemaObject `json:"-"`
 }
 
+// Has nested additional properties
+type AdditionalPropertiesObject6 struct {
+	AdditionalProperties map[string]struct {
+		AdditionalProperties map[string]interface{} `json:"-"`
+	} `json:"-"`
+}
+
 // ObjectWithJsonField defines model for ObjectWithJsonField.
 type ObjectWithJsonField struct {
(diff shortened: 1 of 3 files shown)


deepmap-marcinr:
Oh, this is a hard one. AdditionalProperties is very difficult to synthesize.

Can you reformat your schema like this:
```
components:
  schemas:
    InnerObject:
      additionalProperties: true
    OuterObject:
      type: object
      additionalProperties:
        $ref: "#/components/schemas/InnerObject"

```

johanneswuerbach:
Yes, but I wondered if the autogenerated code could also generate „inner“ structs automatically instead of an anonymous one. This would solve this issue without a need to modify the schema. A naming convention could be the name of the surrounding stuct +Inner for each level. Wdyt?

deepmap-marcinr:
That's the right long term fix, it's just going to require some refactoring. Hopefully the workaround is ok for now.

johanneswuerbach:
Closing this one for now.

PR #565, opened 2022-05-04
Fix utils, generation of operator by the prefix name, add more prefix

Flatch78:
Updated generation of the constants variable name's with specific operators.


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the `typeNamePrefix()` function in `utils.go` already comprehensively handles operator prefixes (Minus, Plus, And, Or, Tilde, Equal, GreaterThan, LessThan, Hash, Dot, Asterisk, Caret, Percent, Underscore, DollarSign, and numeric prefixes).

Issue #568, opened 2022-05-06
Fails to unmarshal date field with empty value


noritama:
When I wrote date format with omitempty, generated struct like this.
```
type Foo {
  MyDate openapi_types.Date `json:"myDate,omitempty"`
}
```

But, a json with empty MyDate, fails unmarshal to this struct.
```
jsonString := `{"myDate":""}`
var foo Foo
err := json.NewDecoder(strings.NewReader(jsonString)).Decode(&foo);
// parsing time "" as "2006-01-02": cannot parse "" as "2006"
```

omitempty should ignoring empty string.
https://pkg.go.dev/encoding/json#Marshal

PR #569, opened 2022-05-06
fix: Unmarshal date field with empty value

noritama:
https://github.com/deepmap/oapi-codegen/issues/568

### DIFF
--- a/pkg/types/date.go
+++ b/pkg/types/date.go
@@ -21,6 +21,9 @@ func (d *Date) UnmarshalJSON(data []byte) error {
 	if err != nil {
 		return err
 	}
+	if dateStr == "" {
+		return nil
+	}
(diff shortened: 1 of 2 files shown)


jxsl13:
Marshal is not Unmarshal. In your issue you refer to the Marshal documentation that does not add a key (nor its value) to the JSON object if the corresponding struct field contains the zero value (e.g. empty string).
This behavior that you try to implement is seemingly (from the top of my head) not the same as the behavior of the standard library "encoding/json".
Especially in the case that you actually want to have a valid date time to be present in a json string. an empty string is not a valid date time value, thus it's invalid.
You would need to remove the key and the value in order for it tonbe valid, as it is not set at all, contrary to set to en empty string.

PR #572, opened 2022-05-13
handle oapi3.MultiError messages

### LINKED ISSUE #441: If `MultiError` is enabled via `chi_middleware.OapiRequestValidatorWithOptions()`, the sever always return 500 as status code about invalid request.
I'm using `chi` with `github.com/deepmap/oapi-codegen/pkg/chi-middleware` to validate the request body.
But if `MultiError` is enabled looks like below, it always return 500 as status code when the request is invalid. I expect it returns 400.
```go
	r := chi.NewRouter()
	r.Use(
		middleware.RequestID,
		middleware.Recoverer,
		chi_middleware.OapiRequestValidatorWithOptions(swagger, &chi_middleware.Options{
			Options: openapi3filter.Options{
				MultiError: true,
			},
		}),
	)
```

I think this switch statement don't care about `openapi3.MultiError` in `validateRequest()`
https://github.com/deepmap/oapi-codegen/blob/master/pkg/chi-middleware/oapi_validate.go#L75-L88
```go
	if err := openapi3filter.ValidateRequest(context.Background(), requestValidationInput); err != nil {
		switch e := err.(type) {
		case *openapi3filter.RequestError:
			// We've got a bad request
			// Split up the verbose error by lines and return the first one
			// openapi errors seem to be multi-line with a decent message on the first
			errorLines := strings.Split(e.Error(), "\n")
			return http.StatusBadRequest, fmt.Errorf(errorLines[0])
		case *openapi3filter.SecurityRequirementsError:
			return http.StatusUnauthorized, err
		default:
			// This should never happen today, but if our upstream code changes,
			// we don't want to crash the server, so handle the unexpected error.
			return http.StatusInternalServerError, fmt.Errorf("error validating route: %s", err.Error())
		}
	}
```

So, it always return `http.StatusInternalServerError`.

## Environment

- `opai-codegen`: v1.8.2
- Go: 1.17

joerocklin:
This adds support for handling the MultiError responses from the ValidateResponse method for the various middlewares. It returns the entire error message verbatim, which may not be the desired response, but I wanted to start somewhere. Open to alternate solutions for the return messages.

fixes #441

### DIFF
--- a/pkg/chi-middleware/oapi_validate.go
+++ b/pkg/chi-middleware/oapi_validate.go
@@ -5,6 +5,7 @@ package middleware
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
@@ -80,6 +81,11 @@ func validateRequest(r *http.Request, router routers.Router, options *Options) (
 	}
 
 	if err := openapi3filter.ValidateRequest(context.Background(), requestValidationInput); err != nil {
+		me := openapi3.MultiError{}
+		if errors.As(err, &me) {
+			return http.StatusBadRequest, me
+		}
+
 		switch e := err.(type) {
 		case *openapi3filter.RequestError:
 			// We've got a bad request
--- a/pkg/chi-middleware/oapi_validate.go
+++ b/pkg/chi-middleware/oapi_validate.go
@@ -19,10 +19,14 @@ import (
 // ErrorHandler is called when there is an error in validation
 type ErrorHandler func(w http.ResponseWriter, message string, statusCode int)
 
+// MultiErrorHandler is called when oapi returns a MultiError type
+type MultiErrorHandler func(openapi3.MultiError) (int, error)
+
 // Options to customize request validation, openapi3filter specified options will be passed through.
 type Options struct {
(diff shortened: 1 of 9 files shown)


Alphasite:
Is there anything I can do to help get this reviewed or merged? This issue effects all 3 server implementations as best i can tell...

joerocklin:
I'm happy to make changes to this to get it accepted, or abandon it if another fix is provided.

jmileson:
I'm also interested in this change

deepmap-marcinr:
I think this needs more versatility around turning MultiError into a HTTP response code. What you have will turn all errors into a BadRequest, even though something might be a different error code. It' difficult to predict how to handle this correctly, so I would propose that you add a function to generate responses from a MultiError, provide a default behavior, but allow it to be overridden.

joerocklin:
@deepmap-marcinr your comment came at the perfect time for me to just get this done today. The new version has a default method that can be overridden as part of the options for each of the middleware. Tests added to verify that the override is used.

deepmap-marcinr:
Thank you!

PR #573, opened 2022-05-15
Support Go Initialisms

### LINKED ISSUE #454: Respect initialisms for schema generation
Per the [Go code review guide](https://github.com/golang/go/wiki/CodeReviewComments#initialisms), initialisms like ID, DB, etc. should have consistent casing like `id` or `ID`. However, the schema generation functionality is just converting the properties to UpperCamelCase and disregarding this rule, which produces structs with fields like `Id` and `Db`.

The culprit seems to be that [schema.go](https://github.com/deepmap/oapi-codegen/blob/546b69d513f7cc98da2eeee46c2aaf43ad895e35/pkg/codegen/utils.go#L524) calls [ToCamelCase](https://github.com/deepmap/oapi-codegen/blob/546b69d513f7cc98da2eeee46c2aaf43ad895e35/pkg/codegen/utils.go#L59), a homegrown case converter. Packages like [iancoleman/strcase](https://github.com/iancoleman/strcase#custom-acronyms-for-tocamel--tolowercamel) support custom acronyms so that users can specify these initialism overrides. We don't necessarily need to import this new package but it's a good illustration of how this might be implemented.

One option is to maintain the current behavior but add a new configuration flag `-initialisms` with values like `ID,HTTP,DB` that would override those parts of the property name.

Another option is to create a new property extension that would allow users to override the struct field name in the Open API definition, e.g. `x-oapi-codegen-name`. Maybe something similar for overriding the json tag as well if necessary. The risk with this approach is that we will alienate people that are using an OpenAPI doc that they do not own and therefore cannot modify to add these extended properties.

Some test cases:
- `dbName`
  - Expected: _DBName string \`json:"dbName"\`_
  - Actual:  _DbName string \`json:"dbName"\`_
- `id`
  - Expected: _ID string \`json:"id"\`_
  - Actual: _Id string \`json:"id"\`_
- `someHttpThing`
  - Expected: _SomeHTTPThing string \`json:"someHttpThing"\`_ (json tag is not changed)
  - Actual: _SomeHttpThing string \`json:"someHttpThing"\`_
- `some_http_thing`
  - Expected: _SomeHTTPThing string \`json:"some_http_thing"\`_
  - Actual: _SomeHttpThing string \`json:"some_http_thing"\`_

jaredpetersen:
Fixes #454 

I'll leave it up to the library owners to determine whether or not this is breaking behavior. Generated code will have different names:

Before:
```go
// Identifier defines model for Identifier.
type Identifier struct {
	Id string `json:"id"`
}
```

After:
```go
// Identifier defines model for Identifier.
type Identifier struct {
        ID string `json:"id"`
}
```

Users still have some ability to override this behavior by adding in uppercase characters, e.g `word_word_WORD` -> `WordWordWORD`. We can override this behavior now that we have a more intense camelCase function (`WordWordWord`) but that might be too drastic of a change so I didn't include that.

### DIFF
--- a/pkg/codegen/utils.go
+++ b/pkg/codegen/utils.go
@@ -57,33 +57,40 @@ func LowercaseFirstCharacter(str string) string {
 // So, "word.word-word+word:word;word_word~word word(word)word{word}[word]"
 // would be converted to WordWordWordWordWordWordWordWordWordWordWordWordWord
 func ToCamelCase(str string) string {
-	separators := "-#@!$&=.+:;_~ (){}[]"
+	specialCharRegexStr := "[-#@!$&=.+:;_~ (){}\\[\\]]"
+
+	// Support initialisms
+	// https://github.com/golang/go/wiki/CodeReviewComments#initialisms
+	// https://staticcheck.io/docs/configuration/options/#initialisms
+	acronyms := []string{"ACL", "API", "ASCII", "CPU", "CSS", "DNS", "EOF",
+		"GUID", "HTML", "HTTP", "HTTPS", "ID", "IP", "JSON", "QPS", "RAM",
+		"RPC", "SLA", "SMTP", "SQL", "SSH", "TCP", "TLS", "TTL", "UDP", "UI",
+		"GID", "UID", "UUID", "URI", "URL", "UTF8", "VM", "XML", "XMPP",
+		"XSRF", "XSS", "SIP", "RTP", "AMQP", "DB", "TS"}
+
+	acronymsSepRegexStr := fmt.Sprintf("($|%s.)|", specialCharRegexStr)
+	acronymsInitialRegexStr := strings.Join(acronyms, acronymsSepRegexStr) +
+		acronymsSepRegexStr
+	acronymsTrailingRegexStr := strings.Join(acronyms, acronymsSepRegexStr) +
+		acronymsSepRegexStr
+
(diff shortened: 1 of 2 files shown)


jaredpetersen:
Looks like there's a few failures related to the diff not lining up with what's expected. I'd like to get some initial thoughts on whether this is something that will likely be merged before continuing. The hard part is done.

deepmap-marcinr:
I'm concerned about this breaking existing code. Can we put this new capital behavior behind a flag? You can default-on your behavior, but allow a way to override.

Please see this flag for an example. https://github.com/deepmap/oapi-codegen/blob/master/cmd/oapi-codegen/oapi-codegen.go#L81

jaredpetersen:
Yeah, that makes sense. It can be done but it will be a bit of a chore to route the configuration to the utility.

deepmap-marcinr:
There's a global options, no need to route anything:
https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/schema.go#L336

jaredpetersen:
This regex implementation has a number of issues when you look closer, particularly if you pass in a string that is already camelCase.

I'm going to close this out for now.

deepmap-marcinr:
Thanks for trying.

I originally thought, "oh, generating some code to save work should be easy", and never in a thousand years, did I imagine how complex this would end up being. It's better not to mess with things because the potential for damage is high, while the gain from code that's more idiomatic but no more correct is low.

PR #574, opened 2022-05-15
Use lower-cased `x-go-name` for parameters var name

emilekm:
Small change to how variable names are generated (a follow up to a [comment](https://github.com/deepmap/oapi-codegen/pull/517#issuecomment-1118934346) on my previous MR and extension to it).
```yaml
components:
  parameters:
    PetID:
      name: pet_id
      x-go-name: PetID
      in: path
      schema:
        type: string
      required: true
```
Before:
```go
func (c *ClientWithResponses) UpdatePetWithResponse(ctx context.Context, petId PetID,  ...)
```
Now:
```go
func (c *ClientWithResponses) UpdatePetWithResponse(ctx context.Context, petID PetID,  ...)
```

### DIFF
--- a/pkg/codegen/operations.go
+++ b/pkg/codegen/operations.go
@@ -119,7 +119,13 @@ func (pd ParameterDefinition) GoVariableName() string {
 }
 
 func (pd ParameterDefinition) GoName() string {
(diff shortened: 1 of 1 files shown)


emilekm:
Breaks compatibility, I'll look into doing this differently.

Issue #286, opened 2021-01-28
Inconsistent Chi(net/http) middleware func signature 


wangyoucao577:
Hi there,     

Many thanks to @maclav3 , I'm able to use the chi middlewares per handler(PR #259) with the request validator chi-middleware (PR #278) to validate restful request via openapi spec. However, I found that the middleware func interface is inconsistent.       

`oapi-codegen` generated middleware func singnature is `type MiddlewareFunc func(http.HandlerFunc) http.HandlerFunc`, but most of industry guys use `func(http.Handler) http.Handler` instead(e.g., https://www.alexedwards.net/blog/making-and-using-middleware).  PR #278 satisfied `func(http.Handler) http.Handler` as well.           

As above, I have to write some ugly codes like this to make them work:       
```go
	sampleSwagger, err := sample.GetSwagger()
	if err != nil {
		glog.Errorf("Error loading sample swagger spec\n: %s", err)
		return
	}
	sampleSwagger.Servers = nil
	sampleValidatorMiddleware := chimiddleware.OapiRequestValidator(sampleSwagger)

	sample.HandlerWithOptions(sampleService, sample.ChiServerOptions{
		BaseRouter: mux,

                // ugly adaptor code
		Middlewares: []sample.MiddlewareFunc{func(next http.HandlerFunc) http.HandlerFunc {
			return http.HandlerFunc(sampleValidatorMiddleware(next).ServeHTTP)
		}},
	})
```

So I'll suggest to change the MiddlewareFunc singature to `type MiddlewareFunc func(http.Handler) http.Handler` when generate code.       

@maclav3  How's your opinion?  

Thanks!

maclav3:
Hey @wangyoucao577 

That sounds like a perfectly reasonable idea; in fact after using the existing signature for a bit, I agree that it is a bit unfortunate 😅

Now the only problem that arises is that we'd be possibly breaking someone's code.
If we wanted to be strict, we should bump the major version.

But:
1) The innovation is recent, so the impact should be small
2) The change should be very easy to make
3) Bumping major impacts everybody, changing the signature only impacts those who have started using the middlewares option.

Let's ask for the opinion of the main maintainer @deepmap-marcinr ?

wangyoucao577:
@maclav3  Excellent! Personally I agree with you that the impact should be small, maybe we don't need to be so that strict, just make the changes fastly.

deepmap-marcinr:
[...]

PR #578, opened 2022-05-19
chi : Change the signature of the `MiddlewareFunc` for compatibilities.

uhey22e:
This PR changes the signature of the `MiddlewareFunc` for chi, and get more compatibility in the ecosystem.

Before:

```
type MiddlewareFunc func(http.HandlerFunc) http.HandlerFunc
```

Now:

```
type MiddlewareFunc func(http.Handler) http.Handler
```

Almost another famous middleware libraries for `net/http` in Go has a function signature below.

For example,
- [go-chi/chi/middleware](https://pkg.go.dev/github.com/go-chi/chi/middleware)
- [rs/cors](https://github.com/rs/cors/blob/a4a5ce87d5a262d09854a195cb6cb1ba1e28c352/cors.go#L219)
- [auth0/go-jwt-middleware](https://github.com/auth0/go-jwt-middleware/blob/e82be78f2661fe7f35dbe798d7c42bcedaf45f52/middleware.go#L50)
- And many more...

### DIFF
--- a/examples/petstore-expanded/chi/api/petstore.gen.go
+++ b/examples/petstore-expanded/chi/api/petstore.gen.go
@@ -86,7 +86,7 @@ type ServerInterfaceWrapper struct {
 	ErrorHandlerFunc   func(w http.ResponseWriter, r *http.Request, err error)
 }
 
-type MiddlewareFunc func(http.HandlerFunc) http.HandlerFunc
+type MiddlewareFunc func(http.Handler) http.Handler
 
 // FindPets operation middleware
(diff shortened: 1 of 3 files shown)


uhey22e:
Related issue #286 

As @deepmap-marcinr said, this PR breaks people's code.
But that's not a big deal, IMHO.

There is an easy work-around to fix the template version.
Following command runs oapi-codegen with version written in the `go.mod` file. So, developers can put breaking changes on ice for some time.

```
go run -mod=mod github.com/deepmap/oapi-codegen/cmd/oapi-codegen
```

PR #582, opened 2022-05-20
Allow custom JSON media types to be marked as `IsJson`

### LINKED ISSUE #581: Allow marking custom vendored JSON types as JSON
In https://github.com/deepmap/oapi-codegen/blob/6a3482493b5731b60c267147522ec3aaa09fc3a3/pkg/codegen/operations.go#L54-L61 we're checking for specifically JSON, but we may have:

- `application/json-patch+json`
- `application/vnd.something+json`
- `application/json; charset=UTF-8`

We should look at adding support for these, too

jamietanna:
As noted in #581, we won't currently support custom media types using
the suffixes, such as custom or vendor media types.

This also adds coverage for the OpenAPI spec's requirement for only a
single MediaType to be defined.

Closes #581.

### DIFF
--- a/pkg/codegen/operations.go
+++ b/pkg/codegen/operations.go
@@ -23,6 +23,7 @@ import (
 	"text/template"
 	"unicode"
 
+	"github.com/deepmap/oapi-codegen/pkg/util"
 	"github.com/getkin/kin-openapi/openapi3"
 )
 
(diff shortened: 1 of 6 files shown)


jamietanna (commented):
This should also support `GenerateBodyDefinitions` to allow marking vendored JSON as `NameTag: "JSON"` to allow for generation on the client-side

jamietanna (commented):
I wonder if we should move this into a `util` method like `util.IsJSON`?

pkg/codegen/operations.go
```diff
@@ -621,7 +624,7 @@ func GenerateBodyDefinitions(operationID string, bodyOrRef *openapi3.RequestBody
 		var defaultBody bool
 
 		switch {
-		case contentType == "application/json":
+		case contentType == "application/json" || strings.HasSuffix(contentType, "+json"):
```
deepmap-marcinr, line 627: Do you think you could make a helper function which takes a contentType and determines if it's JSON? I guarantee this will end up getting copy pasted a lot more.

deepmap-marcinr (changes requested):

PR #585, opened 2022-05-20
Add support for Gorilla generation

### LINKED ISSUE #465: Add support for gorilla/mux
Add support for gorilla/mux https://github.com/gorilla/mux

```
oapi-codegen -generate gorilla-mux
```

jamietanna:
So far this seems to have worked for one of my services, but would be good to hear if others have success.

I'll also be adding the `examples/` version so folks can see how that works too.

Closes #465.


jamietanna:
Hey @deepmap-marcinr sorry for the nudge, but wondering if there's any chance you'd be able to have a review of this, to see if it fits in with how we'd want to do this? 🙏

jamietanna:
Thanks! I think that's right, but maybe I should double check. This was mostly followed by using Chi as a base 🤔

jamietanna:
Having a look through [Gorilla's middleware handling code](https://github.com/gorilla/mux/blob/91708ff8e35bafc8612f690a25f5dd0be6f16864/middleware.go#L24-L28) it's similar to [Chi's middleware handling code](https://github.com/go-chi/chi/blob/b75b525e1d50900a814c222113c415b81648e45c/mux.go#L100-L105), which both appear to be based on how `net/http` handles it, so happy for this to be merged when you're happy with it :raised_hands:

micklove:
@jamietanna , nice work, this is great, have been testing for the last few days, works well 👏 
Will your gorilla changes be in the next release? 

FYI - I used an earlier commit, on the master branch , to test your stuff, works well. 
However, today, when I used the latest master (my build pipeline is just pulling the latest master, till the next release), it appears the mux import is no longer being written out)... example below...

image

jamietanna:
Thanks Mick, yep I'm currently using it in production (https://deliveroo.engineering/2022/06/27/openapi-design-first.html) pinned to a commit off HEAD.

That's odd, mind sharing the config file you're using to generate with? I pulled HEAD yesterday and it still seemed to generate OK for me 🤔

micklove:
Cheers Jamie, it appears to be something going on in the GA pipeline (latest HEAD working locally for me too), I'll sort it :)
(I'll dump the details here if they are relevant, and not just PEBKAC)

V Nice blog post BTW, will have a look at some of those tools you mentioned too, your hard work is much appreciated

corani:
I had hacked together something similar but would like to switch to this implementation. A few differences I noticed that may be worth to discuss:

1. You're generating type *aliases* (e.g. `type Status = string`) while I was generating *new* types (e.g. `type Status string`).
2. For `enum` values you're generating constants named directly after the string value. To avoid collisions I was prefixing them by the type. 

Example of both:
```yaml
status:
  type: string
  enum: [ success, failed ]
```

```go
// Yours:
type Status = string

const (
    Success Status = "success"
    Failed  Status = "failed"
)

// Mine:
type Status string

const (
    StatusSuccess Status = "success"
    StatusFailed  Status = "failed"
)
```

3. Using new types allows you to define methods on them, so for `enum` types I was generating a custom `UnmarshalJSON` method that validated that the provided enum value was actually allowed by the spec (not sure if this is too much "magic").
4. It appears some useless code is being generated (or some code is missing) for optional query parameters:

```go
if paramValue := r.URL.Query().Get("name"); paramValue != "" {
    // this is empty
}

// BindQueryParameter
```

5. When setting the `Methods` for the Gorilla `HandleFunc` you're using strings (e.g. `"GET"`) while I used the constants in the http package (e.g. `http.MethodGet`).

corani:
I've created new issues for them:

1. #656
2. reuse #625
3. #657
4. #658
5. #659

Issue #584, opened 2022-05-20
gin server append error with c.Error() instead of dumping it with c.JSON()


chase-isabelle-roostify:
When the gin server code is generated, it provides a wrapper for the service interface. This wrapper does some request validation logic; however, when it encounters an error, it dumps it to the response payload...

```
// FindPetById operation middleware
func (siw *ServerInterfaceWrapper) FindPetById(c *gin.Context) {
	...
	err = runtime.BindStyledParameter("simple", false, "id", c.Param("id"), &id)
	if err != nil {
		c.JSON(http.StatusBadRequest, gin.H{"msg": fmt.Sprintf("Invalid format for parameter id: %s", err)})
		return
	}
	...
}
```

What is expected is...

```
// FindPetById operation middleware
func (siw *ServerInterfaceWrapper) FindPetById(c *gin.Context) {
	...
	err = runtime.BindStyledParameter("simple", false, "id", c.Param("id"), &id)
	if err != nil {
		c.Error(err)
		return
	}
	...
}
```

...this way a middleware can handle the error. Is there a way to either:
1. Generate the wrapper so it behaves as ☝️ ?
2. Omit the wrapper and implement the validation myself?

This is the command I'm using to generate the gin server:
```
oapi-codegen -generate gin,types,spec -package api api/openapi/api.yaml > internal/api/openapi/api.go
```

chaseisabelle:
I would like to submit a PR for this, but it appears I don't have rights: 
```
$ git push origin gin-error-handler-option
ERROR: Permission to deepmap/oapi-codegen.git denied to chase-isabelle-roostify.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists
```
Could someone grant me rights?

deepmap-marcinr:
You have to send a pull request from your own branch in your own repo, and I will pull the change.

Please see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request

You don't need any push access to our repo to do this, you must push to your own fork.

chaseisabelle:
@deepmap-marcinr thank you! i've added my pr, please have a look when you have a moment. https://github.com/deepmap/oapi-codegen/pull/587

chaseisabelle:
I managed to mitigate this issue using custom templates: https://github.com/deepmap/oapi-codegen#making-changes-to-code-generation

PR #587, opened 2022-05-21
gin server custom error handler option

chaseisabelle:
allows for a custom error handler option to the gin server code generation. default/existing behavior is preserved.

i really want my employer to use this tool going forward, but gin is a requirement, and is custom error handling.

Closes: https://github.com/deepmap/oapi-codegen/issues/584.

### DIFF
--- a/pkg/codegen/templates/gin/gin-register.tmpl
+++ b/pkg/codegen/templates/gin/gin-register.tmpl
@@ -2,7 +2,7 @@
 type GinServerOptions struct {
     BaseURL string
     Middlewares []MiddlewareFunc
-    ErrorHandler func(*gin.Context, error, int) bool
+    ErrorHandler func(*gin.Context, error, int)
 }
 
 // RegisterHandlers creates http.Handler with routing matching OpenAPI spec.
@@ -16,9 +16,8 @@ func RegisterHandlersWithOptions(router *gin.Engine, si ServerInterface, options
     errorHandler := options.ErrorHandler
 
     if errorHandler == nil {
-        errorHandler = func(c *gin.Context, err error, statusCode int) bool {
+        errorHandler = func(c *gin.Context, err error, statusCode int) {
             c.JSON(statusCode, gin.H{"msg": err.Error()})
-            return false
         }
     }
 
--- a/pkg/codegen/templates/gin/gin-register.tmpl
+++ b/pkg/codegen/templates/gin/gin-register.tmpl
@@ -2,6 +2,7 @@
 type GinServerOptions struct {
     BaseURL string
     Middlewares []MiddlewareFunc
+    ErrorHandler func(*gin.Context, error, int) bool
 }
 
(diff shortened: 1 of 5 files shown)


pkg/codegen/templates/gin/gin-register.tmpl
```diff
@@ -2,6 +2,7 @@
 type GinServerOptions struct {
     BaseURL string
     Middlewares []MiddlewareFunc
+    ErrorHandler func(*gin.Context, error, int) bool
```
deepmap-marcinr, line 5: What does the `bool` return value signify here? When would you ever return true, to ignore an error?

chaseisabelle: Good question. The `bool` is to tell the wrapper func to continue execution or not. I can't think of a scenario where one would return `true`, but I wanted to provide that flexibility to the developer.

jxsl13:
bool return value vs error return value?

chaseisabelle:
Well, the error is what the handler is handling, so it seems superfluous to return it back to the caller. Or am I misunderstanding what you mean?

chase-isabelle-roostify:
@deepmap-marcinr @jxsl13 Hey, all. Just a friendly checkup. Any update on this?

jxsl13:
I just had the thought that it might (or might not) make sense to return error values from the error handler in order to do whatever one wants to do with them, log, report, send somewhere, etc.

It's more or less a discussion topic, I'd guess as maybe someone might have a use case for returning an error instead of a bool.

If not, bool is fine.

chaseisabelle:
@deepmap-marcinr Sorry to bug ya. Was just wondering if these changes might be good to go?

deepmap-marcinr:
I've been very busy, I'll take a look when I can

pkg/codegen/templates/gin/gin-register.tmpl
deepmap-marcinr: I've looked through this a bit, and I think there are several parts here that are somewhat difficult to work around.

First, it's impossible to tell from within the error handler where it's been called from - handling a parameter? Bad request body? You can parse it out of the error string.

I think that for now, we can skip returning that `bool`. When would you ever continue a request when you encountered the kind of error that needs handling? 

Ideally, we use kin-openapi's filter as part of middleware to validate requests - and our own code should never get involved in figuring out why a request is invalid. I've been trying to push as much of this logic out of generated code and into middleware as possible, due to flexibility.

chaseisabelle:
I've managed to mitigate this with custom templates: https://github.com/deepmap/oapi-codegen#making-changes-to-code-generation

deepmap-marcinr:
I've been super busy and not putting much work into oapi-codegen recently. All this needs is to have the conflict resolved. I'll see if I can have a look a little later today.

jeroendk:
Awesome, thanks!

PR #590, opened 2022-05-23
Use updated configuration value in docs for `embedded-spec`

jamietanna:


deepmap-marcinr:
This makes the documentation incorrect, as there is no `embedded-spec` option to `--generate`. There is an `embedded-spec` property of the new-style configuration file.

jamietanna:
Ah, woops!

PR #595, opened 2022-05-27
Add additional imports to config

deepmap-marcinr:
This can be used to work around #545, by allowing additional imports,
you can put the models into a separate package.


pkg/codegen/templates/imports.tmpl
```diff
@@ -30,4 +30,7 @@ import (
 	{{- range .ExternalImports}}
 	{{ . }}
 	{{- end}}
+	{{- range $k, $v := .AdditionalImports}}
+	{{ $k }} "{{ $v }}"
```
davidovich, line 34: This form will force giving an import alias. 

Also, I believe that because of the map, you will not be able to add more than one import keyed by "." or "". 

Maybe this could be keyed by import name (with optional alias as value)?

deepmap-marcinr: You're right. I didn't think of that. It's easy enough to take a list with optional alias.

PR #596, opened 2022-05-27
Add a client geneneration option to error on invalid responses

SoMuchForSubtlety:
If the server returns an invalid response (for example status `503` with content type `text/html` from nginx), it can be helpful if the API client returns an error instead of failing silently. Not having this feature has led to a few nil pointer dereferences for me.

It's disabled by default, and can be enabled with `output-options.error-on-invalid: true`.

with the option disabled:
```go
switch {
case strings.Contains(rsp.Header.Get("Content-Type"), "json"):
	var dest Error
		return nil, err
	}
	response.JSONDefault = &dest
}
```

with it enabled:
```go
switch {
case rsp.StatusCode == 204:
	break // No response body
case strings.Contains(rsp.Header.Get("Content-Type"), "json"):
	var dest Error
		return nil, err
	}
	response.JSONDefault = &dest
default:
	return response, errors.New("server returned invalid response")
}
```


jxsl13:
I like this :)

I also do work extensively with the generated client.

Could you update the README.md in this pull request

SoMuchForSubtlety:
> Could you update the README.md in this pull request

Individual options are not really addressed in the README, there is just a reference to the go struct where they are defined. Should I add a section that describes all possible parameters?

SoMuchForSubtlety:
Any chance for a review by a maintainer? 🙂

PR #601, opened 2022-05-31
Generate type names in GoDoc comments

jamietanna:
To make sure that our commented types appear as "official" GoDoc
comments, we need to prefix each

To avoid duplicating too much code, we can create an overloaded
`StringWithTypeNameToGoComment` which delegates to a shared
`stringToGoCommentWithPrefix`.

We then need to move this execution into the template, as we do not have
access to the `TypeName` at the point prior that we're attempting to
generate our comments.

This also includes the fields that are generated inside structs.

This includes any `make generate` changes, too.

### DIFF
--- a/examples/authenticated-api/echo/api/api.gen.go
+++ b/examples/authenticated-api/echo/api/api.gen.go
@@ -27,10 +27,10 @@ const (
 
 // Error defines model for Error.
 type Error struct {
-	// Error code
+	// Code Error code
 	Code int32 `json:"code"`
 
-	// Error message
+	// Message Error message
 	Message string `json:"message"`
 }
 
--- a/examples/petstore-expanded/chi/api/petstore.gen.go
+++ b/examples/petstore-expanded/chi/api/petstore.gen.go
@@ -20,40 +20,40 @@ import (
 
 // Error defines model for Error.
 type Error struct {
-	// Error code
+	// Code Error code
 	Code int32 `json:"code"`
 
-	// Error message
+	// Message Error message
 	Message string `json:"message"`
 }
 
 // NewPet defines model for NewPet.
 type NewPet struct {
-	// Name of the pet
+	// Name Name of the pet
 	Name string `json:"name"`
 
-	// Type of the pet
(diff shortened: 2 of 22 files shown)


deepmap-marcinr:
I would prefer that people write Go-style descriptions in the OpenAPI spec. While this makes linters happy, I think it makes the comments in the code less readable. I intentionally don't mess with these things in the code generator - trying to pass through as much stuff unedited as possible. Someone who cares can fix their spec.

jamietanna:
Hmm, that works when we're in control of the OpenAPI but when ie using something like Petstore, or a vendor OpenAPI there's no chance for us to update it to make it match the Go-style comment by prefixing the type, and that's trivial for us.

Out of interest why do you see this as less readable?

jamietanna:
Hey @deepmap-marcinr @yangzh-deepmap any thoughts on this PR?

jamietanna:
Note that we already do this for `struct`s, so it's simply aligning what we use elsewhere, [i.e.](https://github.com/deepmap/oapi-codegen/blob/master/examples/petstore-expanded/gorilla/api/petstore.gen.go#L39-L40)

jamietanna:
@deepmap-marcinr mind re-reviewing, especially with respect to https://github.com/deepmap/oapi-codegen/pull/601#issuecomment-1172300318 ?

deepmap-marcinr:
You asked why I consider it less readable - it prefixes the field name in the comment just to make linters happy, which could butcher the original sentence.

It's almost impossible to generate lint-friendly Go code from arbitrary OpenAPI specs. I can merge this, as it's really a no-op to code, however, I don't like the comments it produces, I guess that's subjective :)

I have a lower bar for generated code, and I don't assume we can always generate something to satisfy every nitpicky tool, so I err on the side on not modifying stuff that I don't have to.

jamietanna:
That's fair. I guess to clarify, this isn't for linters, just for suggestion/autocomplete hints.

It's not necessarily to get perfect Go comments, but at least to get some information about struct fields.

Up to you ☺

PR #602, opened 2022-05-31
Add basic automated builds

jamietanna:
To ensure that contributions are well-formed, we can set up basic CI.

We can do this following [0] as inspiration, and using our `make` tasks
to automate the process.

We can also create a separate workflow for validating that the `make
generate` has been run, and no untracked changes are present.

[0]: https://blog.kowalczyk.info/article/8dd9c2c0413047c589a321b1ccba7129/using-github-actions-with-go.html
[1]: https://www.jvt.me/posts/2022/04/29/git-uncommitted-changes/

Example failed execution from `generate.yml`: https://github.com/jamietanna/oapi-codegen/runs/7560083866?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,20 @@
+name: Build project
+on: [push, pull_request]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v3
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Test
+        run: make test
+
+      - name: Build
+        run: go build ./cmd/oapi-codegen
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -1,0 +1,20 @@
+name: Ensure generated files are up-to-date
+on: [push, pull_request]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v3
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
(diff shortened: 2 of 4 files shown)


jxsl13:
```yaml
on: [push, pull_request]
name: Test
jobs:
  test:
    strategy:
      matrix:
        go-version: [1.13, 1.14, 1.15, 1.16, 1.17, 1.18]
        platform: [ubuntu-latest, macos-latest, windows-latest]
    runs-on: ${{ matrix.platform }}
    steps:
    - name: Install Go
      uses: actions/setup-go@v3
      with:
        go-version: ${{ matrix.go-version }}
    - name: Checkout code
      uses: actions/checkout@v3
    - name: Code Coverage
      run: go clean -testcache && go test ./... -timeout 30s -race -covermode=atomic -coverprofile=coverage.txt
    - name: Upload coverage to Codecov
      uses: codecov/codecov-action@v3
      with:
        token: ${{ secrets.CODECOV_TOKEN }}
        files: ./coverage.txt
        fail_ci_if_error: false
        verbose: false
```

https://github.com/jxsl13/backupfs/blob/master/.github/workflows/test.yaml

Maybe some inspiration

deepmap-marcinr:
We only support go 1.16+ at this point, and the code will break on anything earlier than 1.16, so there is no need to test all of these versions. Also, there is no platform specific code in here whatsoever, so I'm not sure what value there is in testing on windows, mac and linux. Does this solve a particular problem for you?

I'm happy to merge this, however, I would only test a single version of Go, 1.16, because Go's backward compatibility guarantee will ensure the rest work.

jamietanna:
@deepmap-marcinr should be good to go now 👍 the last run on https://github.com/jamietanna/oapi-codegen/actions?query=branch%3Afeature%2Fci shows what this looks like for the happy-path

deepmap-marcinr:
Thanks for doing this, it should catch some basic errors.

PR #608, opened 2022-06-03
Render UUID in parameters as string

hikhvar:
We make heavy use of the new UUID feature introduced in #571, #546 and #556.
However, we observed that using a UUID as parameter creates malformed
strings in the requests made by the generated clients. This adds a
special handling for types.UUID in the StyleParamWithLocation function.
I'm not that happy with this approach. I would like to use an interface
"MarshalParameter" for this with semantics similar to MarshalJSON. But
that doesn't work for aliased types as UUIDs.

I'm open for any recommendations with a better and more general solution of this problem.

### DIFF
--- a/pkg/runtime/styleparam.go
+++ b/pkg/runtime/styleparam.go
@@ -62,6 +62,15 @@ func StyleParamWithLocation(style string, explode bool, paramName string, paramL
 		t = v.Type()
 	}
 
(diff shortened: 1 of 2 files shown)


jamietanna:
This should be possible to close now. we've had a couple of relevant PRs closed 👍

Issue #609, opened 2022-06-03
For outType = "interface{}" outSchema.SkipOptionalPointer = true is missing


jens1205:
For non required objects without a type, as a fallback `interface{}` is used. But this will later be translated to `*interface{}` which makes no sense.

At lease for `interface{}`, the optional pointer should be skipped.

PR #610, opened 2022-06-03
fix: don't generate an "optional pointer" for unknown types

jens1205:
see https://github.com/deepmap/oapi-codegen/issues/609


jamietanna:
Thanks very much and apologies for the delay!

jamietanna (approved):

Issue #458, opened 2021-10-07
Option to generate a struct without sorting fields


mf-sakura:
I want to generate a struct without sorting fields, in the order listed in the OpenAPI doc.

I wrote my OpenAPI component schema so that the fields that are important to the user are at the top
And I want to return JSON fields on my server response in the same order.
However, for now, generated fields are sorted in increasing order.

So I want an option to generate a struct without sorting field.

openapi.yml
```yml
components:
  schemas:
    Pet:
      properties:
        id:
          type: integer
          format: int64
          description: Unique id of the pet
        name:
          type: string
          description: Name of the pet
        age:
          type: integer
          format: int64
          description: Age of the pet
```

Current
```go
// Pet defines model for Pet.
type Pet struct {
  // Age of the pet
  Age *int64 `json:"age,omitempty"`

  // Unique id of the pet
  Id *int64 `json:"id,omitempty"`

  // Name of the pet
  Name *string `json:"name,omitempty"`
}
```

What I want
```go
// Pet defines model for Pet.
type Pet struct {
  // Unique id of the pet
  Id *int64 `json:"id,omitempty"`

  // Name of the pet
  Name *string `json:"name,omitempty"`

  // Age of the pet
  Age *int64 `json:"age,omitempty"`
}
```

weusego:
+100 -

great job with oapi-codegen, but we are very irritated to see all the fields from our nicely designed structures appear in alphabetical order...

the underlying https://github.com/getkin/kin-openapi/ seems to be the problem, parsing it into a map where the ordering is lost.

gnvk:
@deepmap-marcinr Is there a way to fix / work around this?

psyhatter:
Often this problem is solved by the `x-order` extension, for example in https://github.com/go-openapi/spec or https://github.com/go-swagger/go-swagger

goforbroke1006:
I have check dependecy for parsing openapi specification - github.com/getkin/kin-openapi
openapi3/schema.go:156 - field that keep properties
openapi3/loader.go:159 - unmarshalling yaml to struct

Look like it is a behaviour of "github.com/invopop/yaml" unmarshalling mechanism.
I tried "gopkg.in/yaml.v3" , but does not work.

What suggestions how to fix it without writing a new yaml parser?

renom:
https://github.com/getkin/kin-openapi did use https://github.com/ghodss/yaml but they switched to https://github.com/invopop/yaml for some reason.
[...]

PR #627, opened 2022-06-13
Add support for the x-order extension.

psyhatter:
Fix for #458.
Can be used when you need to set the order of fields in the model:
```yaml
    NewPet:
      type: object
      required:
        - name
      properties:
        name:
          type: string
          x-order: 1
        tag:
          type: string
          x-order: 2
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,6 +533,34 @@ which help you to use the various OpenAPI 3 Authentication mechanism.
   ```
   Name string `json:"name" tag1:"value1" tag2:"value2"`
   ```
+- `x-order`: specifies the order of the fields in the structure. It allows you to specify the order
+  of the fields in the generated structure and will override any default value. This extended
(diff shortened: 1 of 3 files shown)


stevenh:
While this works, could it be simplified to just an `x-ordered` e.g.
```yaml
    NewPet:
      type: object
      required:
        - name
      properties:
        x-ordered: true    
        name:
          type: string
        tag:
          type: string
```
This then uses the order of the properties, without having to put a flag on every field?

filintod:
I created this PR https://github.com/deepmap/oapi-codegen/pull/1190 based of this one, resolving issues with master and minor refactors

jamietanna:
Closed by #1190

psyhatter:
> I created this PR #1190 based off this one, resolving issues with master and minor refactors

@filintod Great work, thank you so much!

PR #630, opened 2022-06-15
Update README to match the new syntax

lzap:
This patch from 2 weeks ago changed the config syntax:

https://github.com/deepmap/oapi-codegen/commit/c0f382b0286d0e19c9fd7eb02f16dfa72ea108bc

however, no change was made to the README. This fixes it, already confused few people I assume :)

Cheers


deefdragon:
It took me an hour to notice that types changed to models. Can you also add a migration guide to the top pointing out the big changes ( server-> server of choice and types-> models) I thought it was just a drop in replace for string to object in the configs that I already had.

PR #631, opened 2022-06-16
Echo: generate `Unimplemented` implementation

urso:
Update echo-interface.tmpl to generate an implementation for ServerInterface that always returns a 404.

The `Unimplemented` server can be used like this (assuming the code is generated in package `spec`):

```
type MyAPIServer struct {
  spec.Unimplemented
}

...
```

Using `Unimplemented` is optional, but might help when prototyping APIs or in case the OpenAPI spec is managed in a separate repo and mutliple developers.

grpc-go uses a similar strategy. For example see the [examples/route_guide/server/server.go](https://github.com/grpc/grpc-go/blob/master/examples/route_guide/server/server.go#L57) file.


pkg/codegen/templates/echo/echo-interface.tmpl
```diff
@@ -5,3 +5,12 @@ type ServerInterface interface {
 {{.OperationId}}(ctx echo.Context{{genParamArgs .PathParams}}{{if .RequiresParamObject}}, params {{.OperationId}}Params{{end}}) error
 {{end}}
 }
+
+// Unimplemented server implementation that returns 404 for each endpoint.
```
jamietanna, line 9:
```suggestion
// Unimplemented server implementation that returns HTTP 501 Not Implemented for each endpoint.
```

pkg/codegen/templates/echo/echo-interface.tmpl
```diff
@@ -5,3 +5,12 @@ type ServerInterface interface {
 {{.OperationId}}(ctx echo.Context{{genParamArgs .PathParams}}{{if .RequiresParamObject}}, params {{.OperationId}}Params{{end}}) error
 {{end}}
 }
+
+// Unimplemented server implementation that returns HTTP 501 Not Implemented for each endpoint.
+type Unimplemented struct {}
+{{range .}}{{.SummaryAsComment }}
+// ({{.Method}} {{.Path}})
+func (_ Unimplemented) {{.OperationId}}(ctx echo.Context{{genParamArgs .PathParams}}{{if .RequiresParamObject}}, params {{.OperationId}}Params{{end}}) error {
```
jamietanna, line 13:
```suggestion
func (Unimplemented) {{.OperationId}}(ctx echo.Context{{genParamArgs .PathParams}}{{if .RequiresParamObject}}, params {{.OperationId}}Params{{end}}) error {
```

jamietanna: Mirrors how we've done it in https://github.com/deepmap/oapi-codegen/pull/1090

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the `Unimplemented` struct pattern now exists for all 7 server frameworks (Echo, Chi, Fiber, Gin, Gorilla, Iris, stdhttp). This PR only targeted Echo's legacy template and the feature is already implemented.

PR #638, opened 2022-06-20
Allow binding types as parameters through `String()` methods

jamietanna:
If trying to use a type, such as a UUID in a path parameter, we'd
previously receive an error at runtime to say that `types.UUID` isn't
supported.

As this is a common use case, we should support it. Instead of adding
support for it manually, we can instead look at whether the there's a
`String` method, and if so use that to serialise the field.

### DIFF
--- a/pkg/runtime/styleparam.go
+++ b/pkg/runtime/styleparam.go
@@ -377,7 +377,12 @@ func primitiveToString(value interface{}) (string, error) {
 	case reflect.String:
 		output = v.String()
 	default:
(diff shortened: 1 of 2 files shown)


jxsl13:
fmt.Stringer

jamietanna:
Looks like it may already be covered by https://github.com/deepmap/oapi-codegen/pull/608 ?

jamietanna:
@deepmap-marcinr @yangzh-deepmap sorry for the small nudge, but would it be possible to have a look? 🙏

Looks like this will help a few folks!

PR #643, opened 2022-06-23
Support param styling for UUIDs

rliebz:
Previously attempting to style a param would result in the error
message:

    unsupported type uuid.UUID

This follows the same pattern as supporting time.Time and types.Date to
now support types.UUID.


jamietanna:
See also #638 and #608.

Issue #636, opened 2022-06-19
Non-Existent Types Referenced by `Response` `Schema`


linden:
I'm not sure why but sometimes generated `type GetExampleResponse struct {}`s reference non-existent types whose names start with a status code (illegal to have a type start with a digit so this refuses to format). This is only happening when the [Schema](https://pkg.go.dev/github.com/getkin/kin-openapi/openapi3#Schema) type lacks a `Title`, so I'm guessing it's something to do with the substitution.

```go
type GetExampleResponse struct {
	Body         []byte
	HTTPResponse *http.Response
	JSON200      *struct {
		Limit   int                 `json:"limit"`
		Offset  int                 `json:"offset"`
		Results []200_Results_Item `json:"results"`
		Total   int                 `json:"total"`
	}
}
```

Here's a simplified example of what it looks like, notice the `Results` field references a (non-existent) type named `200_Results_Item`.
I'm trying to generate a wrapper around [this spec](https://github.com/hirosystems/stacks-blockchain-api/blob/master/docs/openapi.yaml), let me know if you need more information.

jamietanna:
I think this is the same as I saw in #626, which appears to mirror your issue

alexpyc:
Not sure if my issue is related to this one: https://github.com/deepmap/oapi-codegen/issues/1496

mromaszewicz:
Closing as stale — this appears to have been resolved by subsequent work.

**Reproduction attempt (2026-04):** A minimal spec mirroring the report (inline response object, nested `results` array of inline objects, no `Title` on any schema) now generates a valid anonymous struct:

```go
type GetExampleResponse struct {
    Body         []byte
    HTTPResponse *http.Response
    JSON200      *struct {
        Limit   int `json:"limit"`
        Offset  int `json:"offset"`
        Results []struct {
            Name  string `json:"name"`
            Value int    `json:"value"`
        } `json:"results"`
        Total   int `json:"total"`
    }
}
```

Compiles and passes `gofmt` — no `200_Results_Item`-style reference.

**Why it no longer reproduces:** Two independent guards now prevent the broken output:

1. Inline nested schemas in response bodies are emitted as anonymous `struct{}` literals, so no generated name is needed.
2. If a name *were* synthesized from a status code, `typeNamePrefix` in `pkg/codegen/utils.go` prepends `N` whenever the first character is a digit (added in #735, four months after this was filed).

[...]

PR #648, opened 2022-06-30
Generate anonymous objects referenced in schemas

### LINKED ISSUE #626: Generation of anonymous object schemas doesn't seem to work
Given the following OpenAPI:

```yml
openapi: '3.1.0'
info:
  title: ''
  version: ''
servers: []
paths:
  "/healthcheck":
    get:
      summary: "Healthcheck endpoint"
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                additionalProperties: false
                properties:
                  health:
                    type: string
                    enum:
                      - "ok"
```

We'll get generated the following struct:

```go
type GetHealthcheckResponse struct {
	Body         []byte
	HTTPResponse *http.Response
	JSON200      *struct {
		Health *N200Health `json:"health,omitempty"`
	}
}
```

**However**, `N200Health` isn't defined anywhere 😅

jamietanna:
Closes #626, #636.


examples/authenticated-api/echo/api/api.gen.go
```diff
@@ -288,7 +288,8 @@ type ClientWithResponsesInterface interface {
 type ListThingsResponse struct {
 	Body         []byte
 	HTTPResponse *http.Response
-	JSON200      *[]ThingWithID
+
+	JSON200 *[]ThingWithID
```
jamietanna, line 292: Having a bit of difficulty tweaking the whitespace trimming in the template to avoid this diff noise 🤔

internal/test/components/components.gen.go
```diff
@@ -111,9 +111,12 @@ type AdditionalPropertiesObject3 struct {
 
 // Has anonymous field which has additional properties
 type AdditionalPropertiesObject4 struct {
-	Inner                AdditionalPropertiesObject4_Inner `json:"inner"`
-	Name                 string                            `json:"name"`
-	AdditionalProperties map[string]interface{}            `json:"-"`
+	Inner struct {
+		Name                 string                 `json:"name"`
+		AdditionalProperties map[string]interface{} `json:"-"`
+	} `json:"inner"`
```
jamietanna, line 117: We shouldn't be doing this here, in the case that the object already is defined 🤔 

Looks like it may only be the `client-with-responses` that is actually not generating these correctly so I'll have a look

pkg/codegen/schema.go
```diff
@@ -580,7 +580,16 @@ func GenFieldsFromProperties(props []Property) []string {
...
+		if p.Schema.AdditionalTypes == nil {
+			field += fmt.Sprintf("    %s %s", goFieldName, p.GoTypeDef())
+		} else {
+			for _, td := range p.Schema.AdditionalTypes {
+				if p.Schema.TypeDecl() == td.TypeName {
+					out, _ := GenerateGoSchema(p.Schema.OAPISchema.NewRef(), nil)
+					field += fmt.Sprintf("    %s %s", goFieldName, out.GoType)
+				}
+			}
+		}
```
AlainODea, line 592: I'm working on the same problem in a very different way in #654. I have to admit, this is an incredibly elegant and tight solution you have here if it works. Does it work on enums?

jamietanna: It doesn't appear to :grin: We'll need to use a combination of the two, methinks, to get this working

examples/authenticated-api/echo/api/api.gen.go
veger: It is because of the if-statements you introduced. Their tempalte closing tags are `}}`, while `-}}` would prevent the new lines (I guess, didn't try it)

veger:
I am very interested to try it, when it is finished, as I think it is solving a similar issue I am experiencing... :+1:

riftsin:
Hi guys, I was wondering if there was any update on that issue ?

jamietanna:
@deepmap-marcinr I think I may need a hand with the `PascalSchema` error, currently trying to work out the intent of those tests to work out what handling needs to be added

pkg/codegen/codegen.go
```diff
@@ -996,3 +1089,52 @@ func GetParametersImports(params map[string]*openapi3.ParameterRef) (map[string]
...
+				v := operation.Responses[statusCodeOrDefault]
+				sortedContentKeys := SortedContentKeys(v.Value.Content)
+				for _, mediaTypeStr := range sortedContentKeys {
+					mediaType := v.Value.Content.Get(mediaTypeStr)
+					// fmt.Printf("mediaType.Schema: %v\n", mediaType.Schema)
+					// fmt.Printf("mediaType.Schema: %#v\n", mediaType.Schema)
+					// TODO **************************************************
+					// TODO: this is where things aren't working
+					// TODO **************************************************
+					imprts, err := GoSchemaImports(mediaType.Schema)
```
jamietanna, line 1109: Latest debugging has shown that in here, we're getting an `sref` of:

```
sref: &{#/components/schemas/pascalSchema 0x1400021a1a0}
```

Which is because at the point it's being retrieved, it's true, because it's being retrieved from inside the other spec.

Maybe this means that way that I'm pulling out all the referenced things in `GenerateTypesForPaths` is being _too_ exhaustive, and is reaching into other referenced files, which isn't ideal, as we should only be generating things from the base file

jamietanna:
I've got this working finally :clap: 

If you're happy with it, @deepmap-marcinr, I'll sort out the Git history then merge it.

I'll test it out with a few other specs that have been mentioned to be affected, too

jamietanna:
Got a couple of tweaks to make:

- anonymous `requestBodies`
- anonymous `parameters`
- handle when there's i.e. `content: {}`

pkg/codegen/codegen.go
```diff
@@ -462,6 +468,127 @@ func GenerateTypesForSchemas(t *template.Template, schemas map[string]*openapi3.
...
+				for _, mediaTypeStr := range sortedContentKeys {
+					mediaType := v.Value.Content.Get(mediaTypeStr)
+
+					typeName := ""
+					if operation.OperationID != "" {
+						typeName += ToCamelCase(operation.OperationID)
+					} else {
+						typeName += ToCamelCase(strings.ToLower(operationName)) + ToCamelCase(pathName)
+					}
+					typeName += fmt.Sprintf("ResponseBody%s", ToCamelCase(statusCodeOrDefault))
```
deepmap-marcinr, line 549: Is this sufficient to avoid type conflict? What if a request has two response bodies with different content types for the same response code? What happens in that case?

jamietanna: In that case, yep, it'll handle the conflict well enough

[i.e.](https://github.com/deepmap/oapi-codegen/pull/648/files#diff-6686d60bb4293c58e363e155ee4c6b689ce7fc4ab57109d44d691a3b0099bf8bR27-R31):

```go
// NewPetResponseBody200ApplicationYaml defines model for the response body for POST /pets (200, application/yaml).
type NewPetResponseBody200ApplicationYaml = map[string]interface{}

// NewPetResponseBody200TextPlain defines model for the response body for POST /pets (200, text/plain).
type NewPetResponseBody200TextPlain = map[string]interface{}
```

mmorells:
Bump

massenz:
Is there any updated on this?
It is impacting using `go generate` for k8s `Object` DeepCopy etc, as the inlined anonymous structs cannot be consumed to generate the functions.

Simplifying the issue, we have this generated code:

```
type MyBucket struct {
	Spec *struct {
		
		ArchivePeriod *BucketSpecArchivePeriod `json:"archivePeriod,omitempty"`
	 	Cors          *struct {
	 		 ...
	 	}
	 	// other fields
}       
```
when running `go generated ./generated` we get the following error:

```
└─( go generate ./generated
command-line-arguments:-: name requested for invalid type: struct{Cors *struct{MaxAgeSeconds *float32 "json:\"maxAgeSeconds,omitempty\""; Methods *[]string "json:\"methods,omitempty\""; Origins *[]string "json:\"origins,omitempty\""; ResponseHeaders *[]string "json:\"responseHeaders,omitempty\""} "json:\"cors,omitempty\""; Status *struct{FromProvider *map[string]interface{} "json:\"fromProvider,omitempty\""} "json:\"status,omitempty\""}
Error: not all generators ran successfully
run `controller-gen object paths=types.gen.go -w` to see all available markers, or `controller-gen object paths=types.gen.go -h` for usage
generated/types.gen.go:6: running "controller-gen": exit status 1
```
By trying a few different combinations, it appears that the anonymoust `struct`s are the cause of the error.
Thanks!

gainsley:
Another bump for this.

Tommy-42:
Hey @jamietanna do you have any news for this feature ? We would benefit a lot from this and would like to see it merge. 

Do you have any blockers ?

Cheers

jamietanna:
Superceded by https://github.com/oapi-codegen/oapi-codegen/pull/2366

PR #649, opened 2022-07-01
Allow providing custom error funcs in strict chi server

chadweimer:
The new strict server option is great, but it doesn't provide an option to supply custom error functions like the standard server offers. This change adds options to optionally supply custom functions for both request and response errors via separate funcs. It defaults to functions that should preserve existing behavior.

One use case for this is when you don't want to send the detailed error back to the client, thus revealing details of the server, and instead want to log it and return a generic error in its place.


Warboss-rus (approved):
Great change, no compatibility issues. Would love for this to be merged.

PR #653, opened 2022-07-01
Add template overrides from files in new config

### LINKED ISSUE #607: Template overrides in new config style not possible?
Firstly, thanks for the great work.

While converting our configs to the new style I found that the config generator adds the actual templates as plain text under `output-options.user-templates` instead of referencing the template location provided with `--templates`. Having to update every config file after the template changes defeats the purpose of having a template. 

Is there currently an option in the new config style that works similar to the old `--templates` flag? If not, is it possible to add an alternative to the new config that behaves like the old flag?

jacklee1792:
Closes #607.

This PR introduces template file overrides from filenames. Overrides listed in `output-options.user-template-files` will just override whatever is given by `output-options.user-templates`, so this doesn't break any existing configuration files.

Example usage:

```yaml
package: mypackage
generate:
  models: true
  client: true
output:
  test.gen.go
output-options:
  user-template-files:
    client.tmpl: ./templates/my-client.tmpl
    client-with-responses.tmpl: ./templates/my-client-with-responses.tmpl
```

### DIFF
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -18,8 +18,11 @@ import (
 	"bufio"
 	"bytes"
 	"embed"
+	"errors"
 	"fmt"
 	"io/fs"
+	"os"
+	"path"
 	"runtime/debug"
 	"sort"
 	"strings"
@@ -130,6 +133,36 @@ func Generate(spec *openapi3.T, opts Configuration) (string, error) {
 
 	// Override built-in templates with user-provided versions
 	for _, tpl := range t.Templates() {
+		// Check for template in provided template directory
+		if dir := opts.OutputOptions.UserTemplatesDir; dir != "" {
+			fp := path.Join(dir, tpl.Name())
+			_, err := os.Stat(fp)
+			if err != nil && !errors.Is(err, os.ErrNotExist) {
+				return "", fmt.Errorf("error accessing user-provided template %q: %w", fp, err)
+			}
+			if err == nil {
+				utpl := t.New(tpl.Name())
+				data, err := os.ReadFile(fp)
+				if err != nil {
+					return "", fmt.Errorf("error reading user-provided template %q: %w", fp, err)
+				}
+				if _, err := utpl.Parse(string(data)); err != nil {
+					return "", fmt.Errorf("error parsing user-provided template %q: %w", fp, err)
+				}
+			}
+		}
+		// Check for template in a provided file path
+		if fp, ok := opts.OutputOptions.UserTemplateFiles[tpl.Name()]; ok {
+			utpl := t.New(tpl.Name())
+			data, err := os.ReadFile(fp)
+			if err != nil {
+				return "", fmt.Errorf("error reading user-provided template %q: %w", fp, err)
+			}
+			if _, err := utpl.Parse(string(data)); err != nil {
+				return "", fmt.Errorf("error parsing user-provided template %q: %w", fp, err)
+			}
+		}
+		// Check for template provided inline in the configuration
 		if _, ok := opts.OutputOptions.UserTemplates[tpl.Name()]; ok {
 			utpl := t.New(tpl.Name())
 			if _, err := utpl.Parse(opts.OutputOptions.UserTemplates[tpl.Name()]); err != nil {
--- a/pkg/codegen/codegen_test.go
+++ b/pkg/codegen/codegen_test.go
@@ -3,18 +3,20 @@ package codegen
 import (
 	"bytes"
 	_ "embed"
+	"fmt"
 	"go/format"
 	"io"
 	"net/http"
 	"testing"
 
-	examplePetstoreClient "github.com/deepmap/oapi-codegen/examples/petstore-expanded"
-	examplePetstore "github.com/deepmap/oapi-codegen/examples/petstore-expanded/echo/api"
-	"github.com/deepmap/oapi-codegen/pkg/util"
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/golangci/lint-1"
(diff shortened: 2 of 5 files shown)


Celant:
Replacing the templates in the startup seems to break the template functions for me. I've tested overriding the templates in `pkg/codegen/codegen.go` instead, and that seems to work properly. I've tested this [in my fork](https://github.com/deepmap/oapi-codegen/compare/master...Celant:oapi-codegen:feat/user-template-files), and it works correctly. I'd suggest updating your PR to move the config overrides into that file, or if you'd prefer, I can make a PR for my fork.

jacklee1792:
@Celant Interesting, do you know why moving the overrides to `codegen.go` changes the behaviour? As far as I can tell, if you have `old-config-style` set to false it should be the same.

In any case I think your solution is better, feel free to open a PR for your fork. I'll close mine once yours is up.

w32blaster:
@Celant hey mate, can you make a release for your branch, please? I heavily relied on template overriding in my project. Thanks!

jeroendk:
any plans on moving this PR forward?

stevenh:
Liking this functionality, allows for individual templates to be picked from different directories or file names,

That said its still not as easy to use as `-templates <dir>` on the command line, which I suspect is actually the most command and useful case.

I just submitted https://github.com/deepmap/oapi-codegen/pull/852 as having to have go .tmpl files in a .yaml file really isn't usable. It restores compatibility with the old command switches which still have their place.

With all this in mind would you consider adding a `user-template-dir` to this PR?

cmd/oapi-codegen/oapi-codegen.go
```diff
@@ -265,6 +268,27 @@ func updateConfigFromFlags(cfg configuration) (configuration, error) {
 	return cfg, nil
 }
 
+// updateTemplatesFromFiles populates the UserTemplates field in OutputOptions using the
+// UserTemplateFiles field. If there is a shared key between UserTemplates and UserTemplateFiles,
+// the contents of the file take precedence.
+func updateTemplatesFromFiles(cfg configuration) (configuration, error) {
```
stevenh, line 274: As per above
```suggestion
func updateTemplatesFromFiles(cfg *configuration) error {
```

cmd/oapi-codegen/oapi-codegen.go
```diff
@@ -265,6 +268,27 @@ func updateConfigFromFlags(cfg configuration) (configuration, error) {
...
+	oo := &cfg.OutputOptions
+	if oo.UserTemplateFiles == nil {
+		return cfg, nil
```
stevenh, line 277:
```suggestion
		return nil
```

cmd/oapi-codegen/oapi-codegen.go
```diff
@@ -265,6 +268,27 @@ func updateConfigFromFlags(cfg configuration) (configuration, error) {
...
+	}
+	if oo.UserTemplates == nil {
+		oo.UserTemplates = make(map[string]string)
+	}
+	for key, filename := range oo.UserTemplateFiles {
+		data, err := ioutil.ReadFile(filename)
+		if err != nil {
+			return configuration{}, err
```
stevenh, line 285: Wrap the error to provide the needed context to the user to solve the issue.
```suggestion
			return fmt.Error("load user templates %q: %w", key, err)
```

cmd/oapi-codegen/oapi-codegen.go
```diff
@@ -265,6 +268,27 @@ func updateConfigFromFlags(cfg configuration) (configuration, error) {
...
+		}
+		oo.UserTemplates[key] = string(data)
+	}
+	return cfg, nil
```
stevenh, line 289:
```suggestion
	return nil
```

stevenh (changes requested):
Some style comments below, but in addition to this I would move this into the `codegen` package so it works for uses of that and not just `oapi-codegen` command line.

jacklee1792:
In the latest commits I've combined ideas from https://github.com/deepmap/oapi-codegen/compare/master...Celant:oapi-codegen:feat/user-template-files together with https://github.com/deepmap/oapi-codegen/pull/707.

Template overrides can now be specified by any of the following, in order of lowest to highest precedence:
- `user-templates-dir`
- `user-template-files`
- `user-templates`

jacklee1792:
Hey @jamietanna @deepmap-marcinr, are there any plans to move forward with this PR or to provide custom template functionality some other way? As it is right now custom templates are basically unusable without either providing them inline in the config file or writing a custom wrapper around `codegen`.

debuggerpk:
This PR is incredibly useful. Thanks @jacklee1792

kNoAPP:
> Hey @jamietanna @deepmap-marcinr, are there any plans to move forward with this PR or to provide custom template functionality some other way? As it is right now custom templates are basically unusable without either providing them inline in the config file or writing a custom wrapper around `codegen`.

I would love to see some love or activity on this PR. It is a pretty useful feature.

jamietanna:
Apologies on the delay with this one folks! Will try and get a look at it for the next release - would it be possible to get this branch up-to-date with the current state of the repo? :pray:

jacklee1792:
> Apologies on the delay with this one folks! Will try and get a look at it for the next release - would it be possible to get this branch up-to-date with the current state of the repo? 🙏

@jamietanna In rebasing I found https://github.com/deepmap/oapi-codegen/pull/968 already allows loading from files, this PR doesn't do much more than adding another option to load from a directory instead of specifying the file mapping one-by-one. So feel free to close!

jamietanna:
Interesting, thanks for letting me know! Appreciate the work (and apologies for the delay) with this one.

Issue #513, opened 2022-01-28
`types` generation not generating response body types


danthegoodman1:
The imported response bodies are not getting generated as go types with the following file set. As you can see the `matchmakeResponse.yml` type is not generated.

I have the following set of files:

openapi.yml
```yml
openapi: '3.1.0'
info:
  version: 1.0.0
  title: Matchmaking
servers:
  - url: ':8080'
    description: localhost
paths:
  /hc:
    $ref: paths/hc.yml

  /lobbyPrices:
    $ref: paths/lobbyPrices.yml

  /matchmake:
    $ref: paths/matchmake.yml

components:
  securitySchemes:
    Bearer:
      $ref: ./bearerAuth.yml
```

paths/matchmake.yml

```yml
post:
  security:
    - Bearer: []
  tags:
    - Matchmaking
  summary: Find a match for a given game
  requestBody:
    content:
      application/json:
        schema:
          $ref: ../schemas/matchmakeRequest.yml
  responses:
    '200':
      description: Match found
      content:
        application/json:
          schema:
            $ref: '../schemas/matchmakeResponse.yml'
```

schemas/matchmakeRequest.yml
```yml
type: object
required:
  - game
properties:
  game:
    type: string
    enum:
      - CYBERWORMS
    x-oapi-codegen-extra-tags:
      validate: required
  lobbyPrice:
    type: number
    description: The lobby price in cents.
```

schemas/matchmakeResponse.yml
```yml
type: object
required:
  - token
  - address
properties:
  token:
    type: string
    description: The Bearer JWT token that must be passed to the game server as the
  address:
    type: string
    description: The address of the game server to connect to
```

This is the configuration file:

```yml
output: httpServer/http.types.gen.go
package: httpServer
generate:
  - types
```

And this is the golang types file generated:
```go
// Package httpServer provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/deepmap/oapi-codegen version v1.9.1 DO NOT EDIT.
package httpServer

const (
	BearerScopes = "Bearer.Scopes"
)

// GetLobbyPricesParams defines parameters for GetLobbyPrices.
type GetLobbyPricesParams struct {
	// The game to get lobby prices for, will find in the region the backend deems the most relevant.
	Game GetLobbyPricesParamsGame `json:"game"`
}

// GetLobbyPricesParamsGame defines parameters for GetLobbyPrices.
type GetLobbyPricesParamsGame string

[...]

Issue #467, opened 2021-10-19
Generated code references missing enum type


AlbinoDrought:
```yml
openapi: 3.0.2

servers:
  - url: /api

info:
  title: Demo
  version: 0.0.1

paths:
  /demo:
    get:
      responses:
        200:
          $ref: "#/components/responses/DemoBrokenResponse"

components:
  responses:
    DemoBrokenResponse:
      description: |
        Generated struct tries to reference an enum,
        but no type gets generated for that enum.
        Code does not compile.
      content:
        application/json:
          schema:
            type: object
            required: [some_property]
            properties:
              some_property:
                type: string
                enum: ["Some Value", "Some Other Value"]
```

Running `oapi-codegen -o repro.gen.go -package openapi -generate types repro.yml ` gives you

```go
// Package openapi provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/deepmap/oapi-codegen version v1.8.3 DO NOT EDIT.
package openapi

// DemoBrokenResponse defines model for DemoBrokenResponse.
type DemoBrokenResponse struct {
	SomeProperty DemoBrokenResponseSomeProperty `json:"some_property"`
}
```

The `DemoBrokenResponseSomeProperty` type is missing.

![image](https://user-images.githubusercontent.com/852873/137999234-44c7fc0c-a22b-44d0-bee2-8e03cc276de9.png)

Running v1.8.3

sirotinhighline:
Are there any updates on this one? We are facing the same issue and our workaround is to define it as a string instead which is not great. Thank you!

drummonds:
I think I have got to the same issue

obrhoff:
Bumping this issue up. While referencing in a schema works as a workaround, it would be ideal to have this resolved to avoid cluttering schemas with types that are only used once in this specific context.

wgarunap:
I am having the same issue and workaround doesnt work for me as i have separate my spec into multiple packages :|

mromaszewicz:
The issue reported that inline enum properties inside components/responses schemas produced a struct referencing an undefined type. Running the exact spec from the issue against the current binary (v2.6.1) correctly generates both the enum type alias (DemoBrokenResponseSomeProperty string) and its const values. The fix is in schema.go line 621 where AdditionalTypes is populated for non-top-level enums, and codegen.go line 803 where GenerateTypesForResponses includes those additional types. The issue can be closed as fixed.

[...]

PR #654, opened 2022-07-01
Generate enum types from schemas in paths

### LINKED ISSUE #399: Issue generating enum in paths
I may be doing something wrong, or using it wrong, but I am getting errors when generating the HTTP client with the following openapi spec:

```
# openapi.yaml

openapi: 3.0.0
info:
  description: example
  title: example-api
  version: 0.0.1
paths:
  '/path':
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                type: object
                properties:
                  status:
                    type: string
                    enum:
                      - online
                      - offline
                    example: 'online'
components:
  schemas: {}
```

Using master, I generate this with: `go run github.com/deepmap/oapi-codegen/cmd/oapi-codegen -package=clients -generate=client,types -o http.gen.go openapi.yaml`

And that generates the following Go code:

```
...
type GetPathResponse struct {
        Body         []byte
        HTTPResponse *http.Response
        JSON200      *struct {
                Status *N200Status `json:"status,omitempty"`
        }
}
...
```

However  `N200Status` doesn't exist as a type, it hasn't been generated for some reason. Prior to https://github.com/deepmap/oapi-codegen/pull/241 it used to generate it as:

```
type GetPathResponse struct {
        Body         []byte
        HTTPResponse *http.Response
        JSON200      *struct {
                Status *string `json:"status,omitempty"`
        }
}
```

I think the issue might be because enum types are only generated for `swagger.Components` but this enum is on `swagger.Paths`:  https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/codegen.go#L301-L304

I am able to get around this issue by moving the enum into the components section which will then generate all the types:

```
openapi: 3.0.0
info:
  description: example
  title: example-api
  version: 0.0.1
paths:
  '/path':
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Response'
components:
  schemas:
    Response:
      type: object
      properties:
        status:
          type: string
          enum:
            - online
            - offline
          example: 'online'
```

Just wondering if this expected behaviour and if I should always put everything in the components section?

AlainODea:
If defining types in the schema within the parameters, responses, or requestBody within the paths of the spec, generate enums for them.

Also gather deeply nested enums in component responses or requestBodies.

Fixes #399 #467 #513 #626

### DIFF
--- a/examples/authenticated-api/echo/api/api.gen.go
+++ b/examples/authenticated-api/echo/api/api.gen.go
@@ -45,6 +45,15 @@ type ThingWithID struct {
 	Name string `json:"name"`
 }
 
+// N200 defines model for 200.
+type N200 = []ThingWithID
+
+// N201 defines model for 201.
+type N201 = []ThingWithID
+
+// POST defines model for POST.
+type POST = Thing
+
 // AddThingJSONRequestBody defines body for AddThing for application/json ContentType.
 type AddThingJSONRequestBody = Thing
 
--- a/examples/petstore-expanded/chi/api/petstore.gen.go
+++ b/examples/petstore-expanded/chi/api/petstore.gen.go
@@ -51,6 +51,15 @@ type Pet struct {
 // Empty defines model for .
 type Empty = int32
 
+// N200 defines model for 200.
+type N200 = []Pet
+
+// Default defines model for default.
+type Default = Error
+
+// POST defines model for POST.
+type POST = NewPet
+
 // FindPetsParams defines parameters for FindPets.
 type FindPetsParams struct {
 	// tags to filter by
--- a/examples/petstore-expanded/chi/api/petstore.gen.go
+++ b/examples/petstore-expanded/chi/api/petstore.gen.go
@@ -48,6 +48,9 @@ type Pet struct {
 	Tag *string `json:"tag,omitempty"`
 }
 
+// Empty defines model for .
+type Empty = int32
+
 // FindPetsParams defines parameters for FindPets.
 type FindPetsParams struct {
 	// tags to filter by
--- a/examples/petstore-expanded/echo/api/models/models.gen.go
+++ b/examples/petstore-expanded/echo/api/models/models.gen.go
@@ -33,6 +33,9 @@ type Pet struct {
 	Tag *string `json:"tag,omitempty"`
 }
 
+// Empty defines model for .
+type Empty = int32
+
 // FindPetsParams defines parameters for FindPets.
 type FindPetsParams struct {
 	// tags to filter by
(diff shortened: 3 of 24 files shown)


AlainODea:
In retrospect, this breaks some subtle things that don't appear to have functional impact. internal/test/strict-server/client/client.go generates internal/test/strict-server/client/client.gen.go and it ends up with this strange diff after my change (dropping a useless **Empty** type in place of **Reusableresponse**):

```diff
diff --git a/internal/test/strict-server/client/client.gen.go b/internal/test/strict-server/client/client.gen.go
index 6c02ddf..7ddcd94 100644
--- a/internal/test/strict-server/client/client.gen.go
+++ b/internal/test/strict-server/client/client.gen.go
@@ -22,8 +22,8 @@ type Example struct {
        Value *string `json:"value,omitempty"`
 }
 
-// Reusableresponse defines model for reusableresponse.
-type Reusableresponse = Example
+// Empty defines model for .
+type Empty = int
 
 // MultipleRequestAndResponseTypesTextBody defines parameters for MultipleRequestAndResponseTypes.
 type MultipleRequestAndResponseTypesTextBody = string
```

jamietanna (changes requested):
This doesn't quite solve #626, as in this case we get a:

```go
// N200Health defines model for 200.Health.
type N200Health string
```

Whereas I'd expect a `struct`

AlainODea:
> This doesn't quite solve #626, as in this case we get a:
> 
> ```go
> // N200Health defines model for 200.Health.
> type N200Health string
> ```
> 
> Whereas I'd expect a `struct`

Good catch! I had missed that. I'll take another shot at this and see where I can get.

Thank you for the code review.

jamietanna:
@AlainODea no worries! Mind marking this as a draft?

(I've also spotted that there may be cases where we're not (in the existing code) generating enums from nested objects - I'll try and get an issue raised for it soon)

discordianfish:
Are you still working on this? Would be great to see this fixed.

AlainODea:
> Are you still working on this? Would be great to see this fixed.

I would love to. Unfortunately, my immediate need for this has passed and I'm swamped with other priorities at work.

There are a number of related PRs like Jamie's that take a more methodical approach to expanding support for types outside of schemas.

In the meantime, the best approach might be to manually or automatically transform the OpenAPI specs you need by moving things to schemas.

mromaszewicz:
Thank you for submitting this PR, @AlainODea — but we need to close it because all three cases you targeted (inline enums in path params, requestBodies, and responses) now generate correctly. Three independent fixes landed: commit `14fc88e2` (inline param enums, June 2022), commit `270581d0` (`GenerateTypesForPaths()` for response bodies, Nov 2022), and commit `81b9d952` (the Apr 2026 anonymous-schema-hoisting overhaul that feeds op-derived types into `GenerateEnums()`). A repro with the patterns from your PR generates correct named enum types.

Apologies for the long wait — your PR pointed at a real gap, even if the eventual fix took several rounds and a different shape.

Issue #625, opened 2022-06-10
Enum value collision


sulfax:
In the release notes it says value collision should be handled like this:
// Defines values for Enum1.
const (
	Enum1One   Enum1 = "One"
	Enum1Three Enum1 = "Three"
	Enum1Two   Enum1 = "Two"
)

// Defines values for Enum2.
const (
	Enum2One   Enum2 = "One"
	Enum2Three Enum2 = "Three"
	Enum2Two   Enum2 = "Two"
)

However, in my project it is handled like this
// Defines values for Enum1.
const (
	One   Enum1 = "One"
	Three Enum1 = "Three"
	Two   Enum1 = "Two"
)

// Defines values for Enum2.
const (
	Enum2One   Enum2 = "One"
	Enum2Three Enum2 = "Three"
	Enum2Two   Enum2 = "Two"
)

If colliding, a prefix is appended, but this creates inconsistency in my code. Not even sure if this will be generated the same way each time or if this can end up being swapped at one point.

corani:
I found the same issue when looking at differences between the gorilla-server generated code and the code I was generating from my own (hacked together) version. I think it makes sense to always generate the type prefix to be consistent.

PR #662, opened 2022-07-07
Add option to prefix enum values with their typename

corani:
Add a compatibility option to *always* prefix enum values with their typename.

Previously the following spec:
```yaml
status:
  type: string
  enum: [ success, failed ]
```

would be generated as:
```go
type Status string

const (
    Success Status = "success"
    Failed  Status = "failed"
)
```

**or** in case of conflicts:
```go
type Status string

const (
    StatusSuccess Status = "success"
    StatusFailed  Status = "failed"
)
```

This means if you start out without conflicts but introduce conflicts in a later version the generated code unexpectedly changes from under you and you potentially have to fix your code in many places.

When the new option `compatibility.always-prefix-enum-values` is set to `true`, the latter code is always generated.

Closes: #625

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -111,15 +111,17 @@ type EnumDefinition struct {
 	// ValueWrapper wraps the value. It's used to conditionally apply quotes
 	// around strings.
 	ValueWrapper string
-	// Conflicts is set to true when this enum conflicts with another in
-	// terms of TypeNames
-	Conflicts bool
+	// PrefixTypeName determines if the enum value is prefixed with its TypeName.
+	// This is set to true when this enum conflicts with another in terms of
+	// TypeNames or when explicitly requested via the
+	// `compatibility.always-prefix-enum-values` option.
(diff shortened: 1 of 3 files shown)


pkg/codegen/schema.go
```diff
@@ -111,15 +111,17 @@ type EnumDefinition struct {
...
 	// around strings.
 	ValueWrapper string
-	// Conflicts is set to true when this enum conflicts with another in
-	// terms of TypeNames
-	Conflicts bool
+	// PrefixTypename determines if the enum value is prefixed with its TypeName.
+	// This is set to true when this enum conflicts with another in terms of
+	// TypeNames or when explicitly requested via the
+	// `compatibility.always-prefix-enum-values` option.
+	PrefixTypename bool
```
jamietanna, line 118: Probably makes sense to be:
```suggestion
	PrefixTypeName bool
```

If it's using `TypeName`

corani: Good catch, done!

Issue #657, opened 2022-07-05
Feature: automatic validation of enum values


corani:
Based on #656 a custom unmarshal function can be defined on enum types to validate that the provided enum value is actually allowed by the spec. The generated code could look as follows:

```go
func (v *Status) UnmarshalJSON(bs []byte) error {
    var s string

    if err := json.Unmarshal(bs, &s); err != nil {
        return err
    }

    switch Status(s) {
    case StatusSuccess, StatusFailed:
        *v = Status(s)

        return nil
    default:
        return fmt.Errorf("invalid enum value %q", s)
    }
}
```

corani:
I just noticed that `enum` types don't generate type aliases but new types (at least gorilla-server). That means the custom unmarshal function would already be possible.

corani:
Sorry for the noise, I tagged the wrong issue in my PR. I will take a stab at this issue as well in a separate PR though.

corani:
As discussed in PR #663 it's more appropriate to do the validation in a middleware rather than in generated code. 

One example of such a middleware can be found here: https://github.com/deepmap/oapi-codegen/blob/master/pkg/chi-middleware/oapi_validate.go 

This uses `kin-openapi` to automatically validate the requests based on the OpenAPI specification and returns reasonable error messages.

PR #663, opened 2022-07-07
Add option to validate enum values

corani:
Add a compatibility option to generate a custom unmarshaler that validates the enum values.

When the new option `compatibility.validate-enum-values` is set to `true` code like below is generated for the following specification:

```yaml
status:
  type: string
  enum: [ success, failed ]
```

```go
func (v *Status) UnmarshalJSON(bs []byte) error {
    var s string

    if err := json.Unmarshal(bs, &s); err != nil {
        return err
    }

    switch Status(s) {
    case Success, Failed:
        *v = Status(s)

        return nil
    default:
        return fmt.Errorf("invalid enum value: %v", s)
    }
}
```

Closes: #657

### DIFF
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -623,6 +623,7 @@ func GenerateEnums(t *template.Template, types []TypeDefinition) (string, error)
 			enums = append(enums, EnumDefinition{
 				Schema:       tp.Schema,
 				TypeName:     tp.TypeName,
+				TypeDecl:     tp.Schema.TypeDecl(),
 				ValueWrapper: wrapper,
 			})
 		}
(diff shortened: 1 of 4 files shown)


corani:
Discussion:
I'm not sure whether this should be opt-in or opt-out. Making it opt-out may be unexpected as people may already be doing their own validation and returning custom error messages.

deepmap-marcinr:
I don't think this should be done in the generated code, but rather in middleware. We provide an example validator middleware to show how it's done, using kin-openapi. It will definitely catch the invalid enum and return a BadRequest.

corani:
> I don't think this should be done in the generated code, but rather in middleware. We provide an example validator middleware to show how it's done, using kin-openapi. It will definitely catch the invalid enum and return a BadRequest.

Thanks for the feedback. I've tested this: it indeed catches the invalid enums and provides a reasonable error message.

I'll close this PR and the corresponding issue.

Issue #661, opened 2022-07-06
[generating code, enum types, variable names] Duplicate types from enum convertions


brunoluizkatz:
Hi, we have the following type described into specification.yaml:

```yaml
    Units:
      properties:
        units:
          type: string
          enum:
            - C
            - bar
            - l/h
            - l/min
            - m/s
            - 1/min
            - A
            - kW
            - Pa.s
            - g/cm3
            - kWh
            - "%"
            - h
```

I currently creating the generated code with:
```Go
// go:generate go run github.com/deepmap/oapi-codegen/cmd/oapi-codegen --config=oapi-public-config.yaml specification.yaml
```
with the configuration file:

```yaml
package: public
generate:
  chi-server: true
  models: true
  types: true
  embedded-spec: true
output: models/public/api.gen.go
```
When generating the models and types, the result is a duplicated type:

```Go
// NotificationRuleUnits defines model for NotificationRule.Units.
type NotificationRuleUnits string

// Defines values for NotificationRuleUnits.
const (
	NotificationRuleUnits      NotificationRuleUnits = "%"
	NotificationRuleUnitsA     NotificationRuleUnits = "A"
	NotificationRuleUnitsBar   NotificationRuleUnits = "bar"
	NotificationRuleUnitsC     NotificationRuleUnits = "C"
	NotificationRuleUnitsGcm3  NotificationRuleUnits = "g/cm3"
	NotificationRuleUnitsH     NotificationRuleUnits = "h"
	NotificationRuleUnitsKW    NotificationRuleUnits = "kW"
	NotificationRuleUnitsKWh   NotificationRuleUnits = "kWh"
	NotificationRuleUnitsLh    NotificationRuleUnits = "l/h"
	NotificationRuleUnitsLmin  NotificationRuleUnits = "l/min"
	NotificationRuleUnitsMs    NotificationRuleUnits = "m/s"
	NotificationRuleUnitsN1min NotificationRuleUnits = "1/min"
	NotificationRuleUnitsPaS   NotificationRuleUnits = "Pa.s"
)
```
- Note the `NotificationRuleUnits` variable name are the same as type `NotificationRuleUnits`
- I believe the value for enum `%` are not converted to a word like "Perc" or "Percentage" and are removed from the variable name

PR #664, opened 2022-07-07
fix: percent in enum values

corani:
Translate `%` in enum values to `Percent`. E.g. the following model:

```yaml
funny:
  type: string
  enum: [ '%' ]
```

results in:

```go
const (
    FunnyPercent Funny = "%"
)
```

Closes: #661


corani:
This should fix #661, however it feels a bit like whack-a-mole. There's a long list of other characters that are dropped.

Issue #666, opened 2022-07-07
Why are discriminators not supported for AllOf ?


LelouBil:
Hello, 
I have an OpenAPI Schema that uses inheritance like this.

```yaml
ViewConnectionNoRelationship:
      type: object
      additionalProperties: false
      allOf:
        - $ref: "#/components/schemas/ViewConnectionBase"
        - type: object
          additionalProperties: false
          properties:
            name:
              type: string
            documentation:
              type: string
          required:
            - name
            - documentation

ViewConnectionRelationship:
      type: object
      additionalProperties: false
      allOf:
        - $ref: "#/components/schemas/ViewConnectionBase"
        - type: object
          additionalProperties: false
          properties:
            relationship_id:
              type: string
          required:
            - relationship_id
```
With two schemas inheriting from a top schema.

ViewConncetionBase also has a discriminator like this :
```
discriminator:
        propertyName: connection_type
        mapping:
          connection: "#/components/schemas/ViewConnectionNoRelationship"
          relationship: "#/components/schemas/ViewConnectionRelationship"
```

I searched online and found this is how one should use discriminators for inheritance (based on https://redocly.com/docs/resources/discriminator/#allof-for-inheritance).

And for half of my use-case, generating a java client library, it works and generates correctly the inheritance ONLY if I use the discriminator like this. However, for oapi-codegen it doesn't work, and just refuses to merge schemas with discriminators involved, why so ?

https://github.com/deepmap/oapi-codegen/blob/b11a594aa540a0c4ebd25f247228e226bed15f9f/pkg/codegen/merge_schemas.go#L189-L192

At the very least, the library could just ignore the discriminator, since it can't do polymorphism via inheritance in Golang. If I correctly understand what I found, at first the library was using struct embeddings, but it was then later changed to schema merging.

As of right now (and from what I found), the discriminator on AllOf and OneOf/AnyOf is more of a "hint" for the deserializer to find the correct schema on the first try, so simply ignoring it for now in the case of AllOf should not be a problem.

Currently, the library doesn't even read the discriminator for OneOf and just gives it back to the developer to use, along with methods that error if the JSON schema is not matching, so I don't see why it couldn't just be ignored for AllOf.

Antonboom:
+1

ssamsonov-IS360:
+1

mromaszewicz:
[...]

PR #667, opened 2022-07-07
AllOf : Only stop when merging schemas with multiple discriminators

LelouBil:
This is a fix for #666.

It seems to work fine, based on https://github.com/OAI/OpenAPI-Specification/issues/2165 discriminators should be inherited with AllOf, and it can safely be done when only one parent has one.


furkankavukcu:
What happened to this pr?

LelouBil:
Well I did this PR because I was working on a projet that required it. Now I'm doing other stuff.

I'm leaving it open because it might be useful but the maintainers didn't seem to notice it.

mromaszewicz (approved):

PR #680, opened 2022-07-15
Fixes #679: JSON decoding to types.File

### LINKED ISSUE #679: Invalid processing of string format "binary"
Since the introduction of this commit https://github.com/deepmap/oapi-codegen/commit/2cf7fcf5b26d1a4362e7c300bd65c20f4f6c4298 in master, we cannot send arbitrary data in "binary" string formats:

Here is what happens with this kind of string: `data:bonjours`, with the echo Binder:
```
failed to bind body: code=400, message=illegal base64 data at input byte 4, internal=illegal base64 data at input byte 4
```

It happens because of `UnmarshalJSON` in `types.File`, that attempts to JSON decode the input into a []byte. However, when json decoding into a []byte, Go expects the source to be base64 encoded.

"binary" string format is supposed to be "any sequence of octets", not "base64 encoded data".

The Go behaviour can be seen here: https://go.dev/play/p/uk7iNb8YzAR

florentchauveau:
Fixes #679. Because both the source and the destination are of type []byte, Go expects the
source to be base64 encoded. However, the OpenAPI string format "binary" designates
"any sequence of octets".

Because "binary" is matched with types.File, a special processing is required
to avoid the base64 decoding of json.

### DIFF
--- a/pkg/types/file.go
+++ b/pkg/types/file.go
@@ -34,7 +34,12 @@ func (file *File) MarshalJSON() ([]byte, error) {
 }
 
 func (file *File) UnmarshalJSON(data []byte) error {
-	return json.Unmarshal(data, &file.data)
+	var s string
+	if err := json.Unmarshal(data, &s); err != nil {
+		return err
+	}
+	file.data = []byte(s)
+	return nil
 }
 
 func (file File) Bytes() ([]byte, error) {


florentchauveau:
Fixes #679

PR #682, opened 2022-07-16
Fix x-go-type-import code generation

WesleyMiller1998:
While trying to use x-go-type-import for my own project, I ran into an issue where it wasn't generating the types. I dug into the code gen and realized that for each schema, request, or response it actually overwrites the import map. The intended behavior, from my understanding, would be to add the imports rather than overwrite. Overall, the change is pretty small, but it is evident that the test case and example missed this. If desired, I can update the test spec and example.

### DIFF
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -819,7 +819,6 @@ func GetTypeDefinitionsImports(swagger *openapi3.T, excludeSchemas []string) (ma
 }
 
 func GetSchemaImports(schemas map[string]*openapi3.SchemaRef, excludeSchemas []string) (map[string]goImport, error) {
-	var err error
 	res := map[string]goImport{}
 	excludeSchemasMap := make(map[string]bool)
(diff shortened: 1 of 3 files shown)


VladimirStepanov:
Thank you for your changes, I've missed this :(

VladimirStepanov:
@deepmap-marcinr could you approve and merge it?

VladimirStepanov:
@m68kadse could you approve it plz? it's small, but important changes

VladimirStepanov (approved):

VladimirStepanov:
@jamietanna hello! it's very important fix, could you approve and merge it?

jamietanna:
Sorry, I'm not a committer on the repo! I believe the primary maintainer is on vacation at the moment. If you need this, it's worth pinning to your fork (https://www.jvt.me/posts/2022/07/07/go-mod-fork/)

deepmap-marcinr:
Thank you.

PR #684, opened 2022-07-18
Get security scopes from context

corani:
In my project I'm using a sidecar that handles authentication and
injects a header into the request containing the scopes for which the
user is authorized.

In the application itself I annotate the different endpoints with the
required scopes and match these via a middleware function with the
scopes from the user request.

I found myself repeating the code to get the security scopes from the
context in multiple places. This change generates a simple getter for
this that handles the type cast.

Usage:
```yaml
paths:
  /public/resource:
    get:
  /private/resource:
    get:
      security:
        - AuthInfo: ["read"]
    post:
      security:
        - AuthInfo: ["write"]
components:
  securitySchemes:
    AuthInfo:
      type: apiKey
      in: header
      name: X-Authinfo
```

```go
apiScopes := openapi3.AuthInfoScopesFromContext(req.Context())

if scopes == nil {
    // no authorization needed
} else {
    // match apiScopes with request scopes
}
```

### DIFF
--- a/examples/authenticated-api/echo/api/api.gen.go
+++ b/examples/authenticated-api/echo/api/api.gen.go
@@ -25,6 +25,15 @@ const (
 	BearerAuthScopes = "BearerAuth.Scopes"
 )
 
+func BearerAuthScopesFromContext(ctx context.Context) []string {
+	value := ctx.Value(BearerAuthScopes)
+	if value == nil {
+		return nil
+	}
+
+	return value.([]string)
+}
+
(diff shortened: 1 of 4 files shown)


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the middleware has been extracted to separate repos, and the security handling architecture has changed significantly since this PR was written.

PR #685, opened 2022-07-19
Add example for requestBody external reference

hikhvar:


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the codebase has evolved substantially since 2022 and external ref handling has been reworked. Any example would need to be redone fresh.

PR #688, opened 2022-07-20
Chi-server wrapping middleware logic fix.

juddbaguio:
Current implementation server middlewares to a handler goes like this:

```go
var handler = func(w http.ResponseWriter, r *http.Request) {
	siw.Handler.YourRouteHandler(w, r)
}

	handler = middleware(handler)
}
```
The snippet above tends to execute beginning middlewares at the later than the last ones.

The looping is supposed to be like this:
```go
var handler = func(w http.ResponseWriter, r *http.Request) {
	siw.Handler.YourRouteHandler(w, r)
}

for i := len(siw.HandlerMiddlewares) - 1; i >= 0; i--  {
	h := siw.HandlerMiddlewares[i]
        if h != nil {
          handler = h(handler)
        }
}
```
The solution is referenced from [Ardan Lab's Service repository on wrapping middlewares](https://github.com/ardanlabs/service/blob/master/foundation/web/middleware.go)

### DIFF
--- a/pkg/codegen/templates/chi/chi-middleware.tmpl
+++ b/pkg/codegen/templates/chi/chi-middleware.tmpl
@@ -172,8 +172,11 @@ func (siw *ServerInterfaceWrapper) {{$opid}}(w http.ResponseWriter, r *http.Requ
     siw.Handler.{{.OperationId}}(w, r{{genParamNames .PathParams}}{{if .RequiresParamObject}}, params{{end}})
   })
 
-  for _, middleware := range siw.HandlerMiddlewares {
-    handler = middleware(handler)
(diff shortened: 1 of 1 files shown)


deepmap-marcinr:
Hmm, interesting, so we process them in reverse order. This will reverse processing order for anyone who regenerates their code.

juddbaguio:
Exactly, we need to execute the right order of middlewares, this might also cause confusion to the devs who came from the Node.js background.

uhey22e:
+1

FYI,
Chi reverse the order of middlewares as like as this PR.
https://github.com/go-chi/chi/blob/0316d5a1df8598eceb137f5f77945be56810b564/chain.go#L42

jamietanna:
Thanks for this, looks like we got it finished in #787!

PR #699, opened 2022-07-29
Correct error message for new flag

jamietanna:


jamietanna:
Happy to ignore this in lieu of #700

PR #700, opened 2022-07-29
Add fallback for old configuration style

### LINKED ISSUE #629: Last release was a breaking change, should it have not been a v2 release?
The latest release has changed the cli and is now incompatible with previous invocations. It's an easy fix by adding the `--old-config-style`, but users shouldn't have to do that for a minor release.

jamietanna:
As a means to avoid the breaking changes we introduced in v1.11.0, where
the new-style configuration was made the default, we can tweak our
heuristic for how we work out what configuration is in use by only using
the new-style config if the new `--config` flag is used.

This change in itself is non-breaking, and makes it possible to continue
using the old configuration arguments, as long as you don't provide the
new `--config` flag.

Similar to thoughts in #629, but does not require a revert.

Closes #629.

### DIFF
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -34,12 +34,13 @@ func errExit(format string, args ...interface{}) {
 }
 
 var (
-	flagOutputFile     string
-	flagConfigFile     string
-	flagOldConfigStyle bool
-	flagOutputConfig   bool
-	flagPrintVersion   bool
-	flagPackageName    string
+	isOldConfigStyle bool
+
+	flagOutputFile   string
+	flagConfigFile   string
+	flagOutputConfig bool
+	flagPrintVersion bool
+	flagPackageName  string
 
 	// The options below are deprecated, and they will be removed in a future
 	// release. Please use the new config file format.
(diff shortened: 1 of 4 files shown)


deepmap-marcinr:
So we will break people who updated :)

Let me do a couple of experiments first. I want to see whether we can correctly infer config file structure just by parsing the config.

jamietanna:
Thanks Marcin, but no, no one gets broken 👏🏽

This allows people who are using the `--old-config-style` to simply remove that flag and work as before, folks who've not yet updated to the latest version will be able to continue using the old version, and folks who are using the new `--config` can continue.

There are a few places we use both styles of arguments, hence a bit more comfort in what I'm saying, and happy to hear how your own investigations go to see if there are any gaps!

jamietanna:
Hey @deepmap-marcinr just thought I'd check in to see if you'd had a chance to do any further testing to see if this would be a good option for making the "old config" work?

cmd/oapi-codegen/oapi-codegen.go
```diff
@@ -113,8 +113,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	// fallback to the old config style if we don't receive the new `--config` flag
```
deepmap-marcinr, line 116: This won't work, because `--config` has been around for a long time, with a different format config file.

deepmap-marcinr:
What do you think of this? https://github.com/deepmap/oapi-codegen/pull/747

Issue #669, opened 2022-07-10
Strict server return type using matching interface


deefdragon:
The new strict server looks amazing, but using an interface{} for the return type is not ideal. While its not the best as it would not provide an easy way for the user get the list of types, could an interface be used as the return type instead? This would allow for only specific types which match the interface to be returned, granting some type checking. (the user could create their own type matching the interface, but that would have to be done very intentionally)

This would require a separate error type in the response, but I think it would be better to be more explicit when wanting to returrn an error here anyway.

```go
type PetStoreImpl struct {}
func (*PetStoreImpl) GetPets(ctx context.Context, request GetPetsRequestObject) interface{} {
    var result []Pet
	// Implement me
    return GetPets200JSONResponse(result)
}
```

could become 
```go
type PetStoreImpl struct {}

func (*PetStoreImpl) GetPets(ctx context.Context, request GetPetsRequestObject) (resp IsGetPetsResponse, err error) {
    var result []Pet
	// Implement me
    return GetPets200JSONResponse(result)
}

///in api
type IsGetPetsResponse interface {
    CanGetPetsResponse() bool
}

```

Warboss-rus:
When designing strict server interface I was looking for type checking the response object, but for now go does not offer a great variant implementation. The solution should offer type check for the compiler and should be descriptive for humans. The common community solutions for this problems are:
- A struct of pointers with only one of them initialized. Pretty descriptive, but cumbersome to use, since you have to wrap your response into another struct. Also hard to keep invariants, since user can initialize multiple pointers, or none at all, which we can only validate in runtime.
- An interface with useless methods. Offers good type checking, but not as descriptive because of the duck typing. Making the method public also hints user to implement it in his own types, which will not work :), so its better to make the method private (I think standard library uses this approach somewhere). What I don't like about this method is the need to separate error into separate return value. I like the idea of controllers returning a single response, error or not. It will make it harder to migrate to another solution once go gets proper union types.
- Type constraints introduced in go 1.18 can specify an interface which can only be implemented by concrete types, but for now they only work in generic code and cannot be used as a regular types. The [original generic proposal](https://go.googlesource.com/proposal/+/refs/heads/master/design/43651-type-parameters.md#permitting-constraints-as-ordinary-interface-types) states that using constraints as ordinary interface types may be possible, but AFAIK no work on that has been done.
[...]

PR #701, opened 2022-07-30
Strongly typed responses for strict server

Warboss-rus:
Before this, the methods of strict server would return empty interface, like so:
`FindPets(ctx context.Context, request FindPetsRequestObject) interface{}`
With this PR, methods will return an interface unique to each method and an error, like so:
`FindPets(ctx context.Context, request FindPetsRequestObject) (FindPetsResponseObject, error)`

This method-specific interface contains a single method `Visit%METHOD_NAME%Response`, that will marshal the response into `http.ResponseWriter`:
```
type FindPetsResponseObject interface {
	VisitFindPetsResponse(w http.ResponseWriter) error
}
```

This has several advantages:
- Compile-time type checking of response type. Before, you would only get an error in runtime, if you passed an unsupported response object.
- Complicated response marshalling code moved from sever-specific code into common code, making it less error-prone
- Making it clear for user that he can return errors from strict server methods
- It is now simpler to break the strictness by making your own type that satisfies the interface. Before, you would have to use a middleware to access response writer.

The disadvantages are:
- More code is being generated (additional interfaces and methods)
- Breaking change of interface. Not that big of a deal since there was no official release with strict server. Existing user would have to change their code though.
- Harder to do middleware tricks, like redirects. Before you could return anything, catch it in a middleware and execute your own logic. Now, you would have to make separate structs for each method, or use errors (from my expirience most reusable stuff are errors anyway).

@deepmap-marcinr, can you look at this before you make a public release?

This PR is a result of discussion in #669, which this PR should close.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,17 +271,16 @@ function to bind the form to a struct. All other content types are represented b
 
 To form a response simply return one of the generated structs with corresponding status code and content type. For example,
 to return a status code 200 JSON response for a AddPet use the `AddPet200JSONResponse` struct which will set the correct
-Content-Type header, status code and will marshal the response data. You can also return an `error` interface, that will be
-cause an `Internal Server Error` response. If you return a response that is not supported by this method, you will get an error.
-Unfortunately go does not support union types outside generic code, so we can't type check in compile time.
+Content-Type header, status code and will marshal the response data. You can also return an error, that will
+cause an `Internal Server Error` response.
 
 Short example:
 ```go
 type PetStoreImpl struct {}
-func (*PetStoreImpl) GetPets(ctx context.Context, request GetPetsRequestObject) interface{} {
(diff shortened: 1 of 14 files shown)


deefdragon (approved):
I'm not the most qualified person to be reviewing this as I don't know oapi's ins and outs the best, but it looks rather good.

The only thing that I can see is updating the types L631 in the readme, but the readme honestly needs a consolidated effort to update it all at once given some of the recent changes.

deefdragon:
@deepmap-marcinr any chance we can get a review/merge on this?

I've started implementing it on a fork of some of my more "production" code as a test, and I think its worth adding.

deepmap-marcinr:
This is a breaking api change, but I suspect the usage at this point is still quite low, so probably ok.

Issue #544, opened 2022-03-28
[client] nested array property not generated as slice


jxsl13:
```go
//go:generate go run github.com/deepmap/oapi-codegen/cmd/oapi-codegen --generate=types,skip-prune --package=api -o test.gen.go test.yaml
```

**test.yaml**
```yaml
openapi: 3.0.4
info:
  description: example api
  version: v1
  title: example api
servers:
  - url: "https://example.com"
    description: Production

paths:
  /download:
    get:
      summary: "download"
      description: "download"
      operationId: "getDownload"
      responses:
        200:
          description: "download list"
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/DownloadList"

components:
  schemas:

    DownloadList:
      properties:
        archiveDownloadList:
          type: array
          items:
          $ref: "#/components/schemas/ArchiveDownload"
      required:
        - archiveDownloadList

    ArchiveDownload:
      properties:
        downloadUrl:
          type: string
        sha256Sum:
          type: string
        targetDir:
          type: string
        permissionList:
          type: array
          items:
            $ref: "#/components/schemas/PathPermissionItem"
      required:
        - downloadUrl
        - targetDir
        - sha256Sum
        - permissionList

    PathPermissionItem:
      properties:
        path:
          type: string
        recursive:
          type: boolean
          default: false
```

got code:
```go
// Package api provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/deepmap/oapi-codegen version v1.9.1 DO NOT EDIT.
package api

// ArchiveDownload defines model for ArchiveDownload.
type ArchiveDownload struct {
	DownloadUrl    string               `json:"downloadUrl"`
	PermissionList []PathPermissionItem `json:"permissionList"`
	Sha256Sum      string               `json:"sha256Sum"`
	TargetDir      string               `json:"targetDir"`
}

// DownloadList defines model for DownloadList.
type DownloadList struct {
	ArchiveDownloadList ArchiveDownload `json:"archiveDownloadList"` // <--- no slice
}

// PathPermissionItem defines model for PathPermissionItem.
type PathPermissionItem struct {
	Path      *string `json:"path,omitempty"`
	Recursive *bool   `json:"recursive,omitempty"`
}

```

expected code

```go
// Package api provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/deepmap/oapi-codegen version v1.9.1 DO NOT EDIT.
package api

[...]

Issue #373, opened 2021-05-25
OpenAPI 3.1 support?


mitar:
OpenAPI 3.1 is great, brings full JSON Schema compatibility. It would be great if generator would support it.

eli-kaplan:
Bumping this, OAS 3.1 [was released](https://www.openapis.org/blog/2021/02/18/openapi-specification-3-1-released) approximately seven months ago now and brings a number of very useful features.

sonu27:
This would be great

jamietanna:
So right now you can use a document with an OpenAPI 3.1 version, but we don't yet support new stuff.

There's a great post about what we need to add at https://www.openapis.org/blog/2022/05/11/openapi-v3-1-resources-for-tooling-developers

Is there anything we want to count as out of scope? I'd be tempted to look at whether I can help with the effort to add support

jamietanna:
Related upstream issue that we'd need to wait for before we can support everything https://github.com/getkin/kin-openapi/issues/230

jamietanna:
Per https://github.com/getkin/kin-openapi/issues/230#issuecomment-1284319206 this could also be a good option, but would lead to significant rework - may be worth it, especially as kin-openapi's not yet gotten around to it

amery:
If you end up redesigning oapi-codegen using libopenapi, could you please consider allowing generating code for other languages too? I'm sure people here would be interested in using the same awesome tool to generate non-go client code

jamietanna:
Just to cover this, there's some discussion between @deepmap-marcinr and I about this, and we're considering - for now - to convert an OpenAPI 3.1 spec to an OpenAPI 3.0 spec (via libopenapi) to continue to rely on kin-openapi, which will allow for OpenAPI 3.1 specs to continue to work, but without some of the benefits.

We'd prefer to keep on kin-openapi for the meantime, but also if there's no expected update to 3.1 (and folks do really want 3.1) we can look at shifting to use libopenapi, but it'll require a slightly more significant set of changes than we're willing to do right now, especially as [Marcin is continuing to think about what a v2 could look like](https://github.com/deepmap/oapi-codegen/discussions/440)

jamietanna:
For folks looking for OpenAPI 3.1 support, what features are you hoping for? We are looking at what initial changes may be required (supporting multiple `type`s in a schema), but then wondering what additional things could be added to the backlog, like webhook support

mgabeler-lee-6rs:
[...]

PR #713, opened 2022-08-10
Validate OpenAPI spec before performing codegen

jamietanna:
Although #544 couldn't have been caught by this, we may be able to flag
to users of cases where their OpenAPI is incorrect.

This is a draft because until
https://github.com/deepmap/oapi-codegen/issues/373 is resolved, this may break
valid OpenAPI 3.1 specs.

### DIFF
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


jamietanna:
Let's make it opt-in?

mromaszewicz:
I added something similar recently:
https://github.com/oapi-codegen/oapi-codegen/pull/2435

The PR above adds validation for catching codegen errors early and printing nice error messages, versus failing with compile errors later. Given that we have 3.1 support now, you might want to revisit this (but the same argument now holds for 3.2)

PR #715, opened 2022-08-10
Add name resolver to codegen config

wtertius:


pkg/codegen/configuration.go
```diff
@@ -74,8 +74,9 @@ type OutputOptions struct {
 	ExcludeTags   []string          `yaml:"exclude-tags,omitempty"`   // Exclude operations that have one of these tags. Ignored when empty.
 	UserTemplates map[string]string `yaml:"user-templates,omitempty"` // Override built-in templates from user-provided files
 
-	ExcludeSchemas     []string `yaml:"exclude-schemas,omitempty"`      // Exclude from generation schemas with given names. Ignored when empty.
-	ResponseTypeSuffix string   `yaml:"response-type-suffix,omitempty"` // The suffix used for responses types
+	ExcludeSchemas     []string       `yaml:"exclude-schemas,omitempty"`      // Exclude from generation schemas with given names. Ignored when empty.
+	ResponseTypeSuffix string         `yaml:"response-type-suffix,omitempty"` // The suffix used for responses types
+	NameNormalizer     NameNormalizer `yaml:"-"`                              // Function used to normalize golang names and types
```
jamietanna, line 79: 🤔 I think we're missing a binding here for the YAML tag name?

Also could we please add a test file (see #582 for an example in `internal/test/client`) to make sure the configuration + generation works? This will make sure that we don't break it in the future!

wtertius: > I think we're missing a binding here for the YAML tag name?

I'm not sure, we can serialize reference to function, that's why it's skipped. I can put some string here with preset values, like `default` and `custom`. Let's agree what should be here?

wtertius: Could you please review now?

I've moved the name normalizers inside the library and given ability to configure name normalizer as preset string. Test has been added also.

wtertius:
Moving to new PR: https://github.com/deepmap/oapi-codegen/pull/749

Issue #714, opened 2022-08-10
Add Strict server support for Gorilla Mux


jamietanna:
When we implemented #499, it was started before we had Gorilla Mux supported (#465), therefore support is missing.

We should add support for Gorilla Mux.

Warboss-rus:
It should work fine with chi version, since the difference is only in routing

jamietanna:
Closed by #716

PR #716, opened 2022-08-10
Add support for strict server with Gorilla

DouglasDwyer:
This PR adds an extra template which supports strict server generation with Gorilla mux. The template is very similar to the Chi server template.


Warboss-rus:
Why not just use chi template and generate it for gorilla? There is no router-specific code in strict templates

DouglasDwyer:
Agreed - there is no router-specific code in the Chi server template, so it could work in lieu of something Gorilla-specific. Currently, calling oapi-codegen with the `gorilla` and `strict` options just doesn't generate the strict server implementation at all. Should I replace the reference to the Gorilla template in `operations.go` with a reference to Chi server? If we choose to do this, might it make more sense to rename the strict Chi server to something more neutral (like StrictHTTPServer)?

Warboss-rus:
Yeah, you can rename the template to strict-http and generate it for both chi and gorilla

DouglasDwyer:
Sounds good. I've gone ahead and made the requested changes.

Warboss-rus (approved):
LGTM

jamietanna:
This would close #714 if merged - thanks for raising, quick glance on mobile looks good, but will have a proper review later 🤞🏽

deepmap-marcinr:
It looks like the generated files need to be regenerated for precommit to pass.

PR #717, opened 2022-08-10
Properly create models for allOf schemas with properties additionally defined

DouglasDwyer:
The types generator does not properly create structure types for models of the following format. The properties defined directly on the model are not included in the output struct:

```
SessionIdentity:
  required:
  - websocket_url
  type: object
  allOf:
     - $ref: '#/components/schemas/SessionData'
  properties:
    websocket_url:
      type: string
      description: websocket url
```

This is because models that contain an allOf are constructed by merging the properties of all base types together. However, properties which are defined on the actual model itself are not considered. This PR addresses the problem by including the actual schema instance during allOf merging.

### DIFF
--- a/pkg/codegen/merge_schemas.go
+++ b/pkg/codegen/merge_schemas.go
@@ -9,31 +9,30 @@ import (
 
 // MergeSchemas merges all the fields in the schemas supplied into one giant schema.
 // The idea is that we merge all fields together into one schema.
-func MergeSchemas(allOf []*openapi3.SchemaRef, path []string) (Schema, error) {
+func MergeSchemas(sc *openapi3.Schema, path []string) (Schema, error) {
 	// If someone asked for the old way, for backward compatibility, return the
 	// old style result.
 	if globalState.options.Compatibility.OldMergeSchemas {
-		return mergeSchemas_V1(allOf, path)
+		return mergeSchemas_V1(sc.AllOf, path)
 	}
-	return mergeSchemas(allOf, path)
+	return mergeSchemas(sc, path)
 }
 
-func mergeSchemas(allOf []*openapi3.SchemaRef, path []string) (Schema, error) {
+func mergeSchemas(sc *openapi3.Schema, path []string) (Schema, error) {
+	allOf := sc.AllOf
 	n := len(allOf)
 
-	if n == 1 {
-		return GenerateGoSchema(allOf[0], path)
-	}
+	schema := *sc
+	schema.AllOf = nil
 
-	schema := *allOf[0].Value
-
-	for i := 1; i < n; i++ {
+	for i := 0; i < n; i++ {
 		var err error
 		schema, err = mergeOpenapiSchemas(schema, *allOf[i].Value)
 		if err != nil {
 			return Schema{}, fmt.Errorf("error merging schemas for AllOf: %w", err)
 		}
 	}
+
 	return GenerateGoSchema(openapi3.NewSchemaRef("", &schema), path)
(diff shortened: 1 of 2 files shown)


chanced:
I'm not sure your approach is valid. If anything, I think the properties would end up being those you describe on the base schema, overriding the results of allOf. I could be mistaken there though.

DouglasDwyer:
@chanced Thanks for your feedback. Currently, the issue that we are facing with this is that the generated model only contains the fields of the inherited types (in the case of the example, that would be `SessionData`). The changes in this PR add the properties on the schema itself, as well as all of its `allOf` schemas, to the list of schemas to merge together.

Specifically, when it comes time to generate a "merged" schema containing all properties, I begin by creating a copy of the original schema. I set its allOf property to nil, to avoid an infinite egress. Then I iterate over the original schema's allOfs, merging them together with the copied schema, to create a merged result.

The result is a struct with the allOf schema properties and the schema properties themselves merged together.

chanced:
My point is is that things don't get merged in JSON schema without them being in `oneOf` or `anyOf`. I'm just not 100% certain which would take precedent. I'm fairly certain the base schema's `properties` would though.

To achieve what you are after, use the schema:

```yaml
SessionIdentity:
  required:
  - websocket_url
  type: object
  allOf:
     - $ref: '#/components/schemas/SessionData'
     - type: object
        properties:
          websocket_url:
            type: string
            description: websocket url
```

DouglasDwyer:
@chanced I see - your concern is with regard to the OpenAPI standard itself? Unfortunately, I couldn't find any relevant place in the standard where the behavior for mixing properties and `allOf` is defined. However, both the official Swagger generator and the general OpenAPI generator properly handle this case, by generating a struct with both sets of properties. Plus, according to [this StackOverflow answer](https://stackoverflow.com/questions/49604468/swagger-schema-oneof-anyof-allof-valid-at-the-same-time) and [the associated source](https://github.com/cloudflare/doca/issues/52), the given schema with properties should behave the same as the schema example that you gave.

I agree that placing the properties inside of an `allOf` tag would be a workaround for this issue. However, wouldn't it be better for the behavior to match that of OpenAPI generator? Thanks.

chanced:
Ah, I may be mistaken regarding properties. If so, I apologize.

chanced:
@DouglasDwyer I was completely wrong about the merging of `properties`. I apologize for adding noise to your pull request. 

I meant to mention this earlier and forgot.

Swiftwork:
I wanted to lend some support for this pull-request as I too ran into the same issue. I cloned this pull-request and tried it on my full REST service OpenAPI 3.1 spec and it produced identical results as the previous merged allOf schema.

With this PR Schema
```yaml
User:
  required:
    - id
    - firstName
  allOf:
    - type: object
      properties:
        id:
          type: string
    - $ref: '#/components/schemas/UserProperties'
```
Produces the same result as
```yaml
User:
  allOf:
    - type: object
      properties:
        id:
          type: string
    - $ref: '#/components/schemas/UserProperties'
    - type: object
      required:
        - id
        - firstName
```

genslein:
Hi folks any update here? This is really problematic as this is the recommended approach for achieving this behavior from the OAS community https://github.com/OAI/OpenAPI-Specification/issues/1514#issuecomment-374268421

mromaszewicz:
Thank you for this change, however, it introduces a backward incompatible change to the generated code. #1415 is a more targeted fix. I'm closing this in favor of the other fix.

PR #726, opened 2022-08-19
add option to generate custom server based on user templates

marcos10soares:
The goal of this PR is to add the option to use "user provided" templates without depending on a specific framework, by using the new option `custom-server`.

Objectives:
- when calling oapi-codegen user provides a folder with multiple template files, the number of files or names do not matter
- all the files should get parsed and the result returned as a single file

Non goals
- support multiple file output (I tried this but it changed the current implementation too much) 🥲 

This is useful when you want to create custom/private implementations (in my case, based on httprouter) that are too opinionated to be used by other people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,19 @@ Strict server also has its own middlewares. It can access to both request and re
 as well as raw request\response data. It can be used for logging the parsed request\response objects, transforming go errors into response structs,
 authorization, etc. Note that middlewares are server-specific.
 
+#### Custom server generation
+
+oapi-codegen also supports custom templates as an input to generate server code. This option does not use any of the existing templates, all the templates must be provided by the user.
+The goal of this option is to allow users to experiment with code generation for other frameworks or even avoid frameworks if they want.
+
+The user provided folder with templates can have as many files as desired, they will be parsed and concatenated into a single file.
+
+Example usage:
+
+```bash
+oapi-codegen -generate custom-server -o server.gen.go -old-config-style -package server -templates ./your-template-folder-path examples/petstore-expanded/petstore-expanded.yaml
+```
+
 #### Additional Properties in type definitions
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -305,6 +305,19 @@ Strict server also has its own middlewares. It can access to both request and re
...
+
+oapi-codegen also supports custom templates as an input to generate server code. This option does not use any of the existing templates, all the templates must be provided by the user.
+The goal of this option is to allow users to experiment with code generation for other frameworks or even avoid frameworks if they want.
+
+The user provided folder with templates can have as many files as desired, they will be parsed and concatenated into a single file.
+
+Example usage:
+
+```bash
+oapi-codegen -generate custom-server -o server.gen.go -old-config-style -package server -templates ./your-template-folder-path examples/petstore-expanded/petstore-expanded.yaml
```
jamietanna, line 318: I'd say we should recommend using the new config style for primary examples :pray:

README.md
```diff
@@ -647,6 +660,8 @@ you can specify any combination of those.
  same package to compile.
 - `chi-server`: generate the Chi server boilerplate. This code is dependent on
  that produced by the `types` target.
+- `custom-server`: generate a server file based on user provided templates. This code 
+ is not dependent on the code produced by the `types` target, but could be used it also.
```
jamietanna, line 664: > but could be used it also.

What do you mean, sorry?

jamietanna (commented):
This looks interesting! Mind adding an example into `internal/test` so we can see a) what it looks like and b) have a way of proving that we don't have regressions in the future? May also be worth having one in `examples/petstore-expanded`, too

marcos10soares:
dropping this PR and replacing it with this: https://github.com/deepmap/oapi-codegen/pull/944

PR #722, opened 2022-08-16
Add linting with golangci-lint

jamietanna:

PR #727, opened 2022-08-24
Remove io/ioutil

leonnicolas:
This Commit edits the go templates so that io/ioutil is replaced by io.

This will make oapi-codegen generate code that is compliant with [staticcheck](https://staticcheck.io/).

Since go 1.16 io/ioutil is deprecated. The current version of oapi-codegen
requires go 1.16 to compile. Most users will also use at least go 1.16
to run the generated code because no binaries are distributed, but users
need to compile oapi-codegen themselves.
If not, this is would be a breaking change.

Signed-off-by: leonnicolas <leonloechner@gmx.de>

### DIFF
--- a/examples/authenticated-api/echo/api/api.gen.go
+++ b/examples/authenticated-api/echo/api/api.gen.go
@@ -11,7 +11,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"path"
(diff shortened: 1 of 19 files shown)


jamietanna (approved):
Thanks very much for this!

This covers the generated code, but we've still got some internal code to clean up - that's a good approach to start with, and this will make generated code compliant which is most important :clap:

leonnicolas:
Thanks for the feedback. I also have a work in progress to update the oapi code itself. I can make another PR for it. I guess it would be needed for https://github.com/deepmap/oapi-codegen/pull/722 anyways.

deepmap-marcinr:
Please regenerate everything, it looks like this commit missed some files. See the failed precommit. Happy to merge afterward.

leonnicolas:
> Please regenerate everything, it looks like this commit missed some files. See the failed precommit. Happy to merge afterward.

done, I deleted a new line and forgot to run `make generate`

deepmap-marcinr:
You've still got one failure. I really dislike that `moq` file :)

PR #731, opened 2022-08-28
Update jsonmerge.go

mrkouchaki:
uable to find /v2

### DIFF
--- a/pkg/runtime/jsonmerge.go
+++ b/pkg/runtime/jsonmerge.go
@@ -3,7 +3,7 @@ package runtime
 import (
 	"encoding/json"
 
-	"github.com/apapsch/go-jsonmerge/v2"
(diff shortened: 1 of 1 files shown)


jamietanna:
Hey, not sure why this change is needed - I've got at least one project that's using the jsonmerge changes successfully :thinking:

jxsl13:
there is a v2.0.0 tag but not a /v2 path. Don't know the implications of those for the versioning and import paths.

jamietanna:
Good spot - this may be because the author has pinned the module version path inside go.mod - https://github.com/apapsch/go-jsonmerge/commit/90b01458c2af82553f17c56a431386c57a8a569f

jamietanna:
@mrkouchaki any thoughts on the above? I'll probably move to close this as I don't think it's needed

Issue #734, opened 2022-08-29
Broken code is generated when spec. includes operationId starts with digit


masu-mi:
When `operationId` that starts with digit is set explicitly, I can't get correct go code.
To avoid trouble, we can overwrite the `operationId` with other word but 3GPP's specification have many `opeartionId` fields start with digit. I understand it's not recommended value for operationId, but it's not banned.

> Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.
> ref. https://spec.openapis.org/oas/v3.1.0#operation-object

Because of it, I think it's acceptable to add some prefix string (e.g. `Op`) when `operationId` start with some digit.

## reproduce example

```sh
go run github.com/deepmap/oapi-codegen/cmd/oapi-codegen --config=config.yaml spec.yaml
```

```yaml
## spec.yaml
openapi: 3.0.2
info:
  version: "0.0.1"
paths:
  /3gpp/foo:
    get:
      operationId: 3GPPFoo
```

```yaml
## config.yaml
package: head_digit_of_opeartion_id
generate:
  strict-server: true
output: issue.gen.go
output-options:
  skip-prune: true
```

PR #735, opened 2022-08-29
Fix/handle initial digit of operation

masu-mi:
# Overview
resolve https://github.com/deepmap/oapi-codegen/issues/734

## changes

* Add example spec. files. in `internal/test/issues/issue-head-digit-of-operation-id` directory.
* Change `codegen.GenerateResponseDefinitions()`
* Add generated code with cd `internal/test/issues/issue-head-digit-of-operation-id`; go generate


pkg/codegen/operations.go
```diff
@@ -510,10 +510,12 @@ func OperationDefinitions(swagger *openapi3.T) ([]OperationDefinition, error) {
 					return nil, fmt.Errorf("error generating default OperationID for %s/%s: %s",
 						opName, requestPath, err)
 				}
-				op.OperationID = op.OperationID
 			} else {
 				op.OperationID = ToCamelCase(op.OperationID)
```
deepmap-marcinr, line 514: Please see the `typeNamePrefix` function in utils.go - it can do something like this automatically, combined with `SanitizeGoIdentity`.

PR #743, opened 2022-09-06
Add Import Functionality to Parameter and Header Components, Change Imports to Aliases

WesleyMiller1998:
This PR extracts GetImport from GetImports so that schemas defined in parameter and header components can be added easily. The PR is relatively small and a test has been added for parameter. GetImports has also been refactored to use the new GetImport function. 

There are also some random gofmt formatting changes.


WesleyMiller1998:
Frankly I have no idea how to make 2 forks/PRs at the same time so I just added to this PR.

I made a "minor" change so that import types are now aliases rather than redefined types. I feel as though this is the intended behavior, but I could be mistaken. Under this method, imported types will inherit MarhsalJSON and MarshalText which they were not doing before. I feel like this is a crucial component, but again I could be mistaken. I suppose it would be possible to make this optional through the api config, so feedback would be appreciated.

WesleyMiller1998:
@deepmap-marcinr I believe that the first half of these changes are pretty important. If you'd like to me to split it into a second PR for the second half, I am willing to, but from what I can tell those changes are also important. Will definitely accept feedback on this if you disagree

deepmap-marcinr:
Someone has already fixed imports in parameters. I cherry-picked your change about aliases.

PR #747, opened 2022-09-08
Be more flexible parsing config files.

### LINKED ISSUE #629: Last release was a breaking change, should it have not been a v2 release?

deepmap-marcinr:
Release v1.11.0 introduced a new config file format, and it's default-on, which breaks existing generation upon update. This change attempts to infer the version of the config file and flags from the the structure of the file, or the presence of deprecated flags.

Closes #629

### DIFF
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -60,7 +60,7 @@ type configuration struct {
 	OutputFile string `yaml:"output,omitempty"`
 }
 
-// This structure is deprecated. Please add no more flags here. It is here
+// oldConfiguration is deprecated. Please add no more flags here. It is here
 // for backwards compatibility and it will be removed in the future.
 type oldConfiguration struct {
 	PackageName        string                       `yaml:"package"`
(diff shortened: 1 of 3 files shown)


cmd/oapi-codegen/oapi-codegen.go
```diff
@@ -113,8 +113,73 @@ func main() {
 		os.Exit(1)
 	}
 
+	// We will try to infer whether the user has an old-style config, or a new
+	// style. Start with the command line argument. If it's true, we know it's
+	// old config style.
+	var oldConfigStyle *bool
```
jamietanna, line 119: I'd probably keep this as a `bool` rather than `*bool`, so it's easier to work with? Not given it a go, but I think that'd be possible

jamietanna (commented):
I think this looks pretty good as a solution and alternative to #700 🙌 

I guess you've given it a go with some of the old config files to test with, too?

cmd/oapi-codegen/oapi-codegen.go
deepmap-marcinr: I made it a bool to differentiate between known false and unset. It's hard to know in fallbacks whether something is actually false or zero-value false. Didn't feel like making an optional just for this.

PR #749, opened 2022-09-12
Generate names with initialisms

wtertius:
Adding support of two naming styles:
* new - with initialisms like `MyHTTPHandlerForAPI`
* old - with simple camel case like `MyHttpHandlerForApi`

By the steps of https://github.com/deepmap/oapi-codegen/pull/715.


deepmap-marcinr:
While it's nice to be consistent with initialisms, I generally feel it's not worth changing the code generation behavior for things like this, because lots of people will need to update their code once it breaks, or we need to flagify this fix to be optional.

Jleagle:
Does it not default to the current behaviour, and only a config change will break things, which is expected?

deepmap-marcinr:
Sorry, looked too fast. I resolved a bunch of conflicts (hopefully correctly). Thanks for the fix.

deepmap-marcinr:
Sorry, I had to revert, because some initialisms broke in the existing code, and I had compiler errors. I'm happy to merge this, once it doesn't break existing code (or maybe I merged conflicts improperly).

Issue #751, opened 2022-09-12
broken code at generating alias type of type in external package


masu-mi:
# Overview
Because we handle `.Ref` as simple string in `strict-interface.tmpl`, an output get like a `type ExxxxResponse = ExternalRef0Foo` in the case of generating alias type of type in external package. It's invalid code in golang.

masu-mi:
Thank you for review on my PR. I've confirmed the problem resolved.

PR #752, opened 2022-09-12
Ft/handle external pkg name in response

masu-mi:
# Overview
an example resolving of https://github.com/deepmap/oapi-codegen/issues/751

## changes

* add test `internal/test/issues/issue-removed-external-ref`
* add `UppercaseFirstCharacterWithPkgName ` function for `ucFirstWithPkgName` filter
* use `ucFirstWithPkgName` filter instead of `ucFirst` filter in template


deefdragon:
I would like to note that this caused a break in a pattern I had been using in a pre-release.

Specifically, before this change, it was possible to return the type that was being referenced directly, but now that its embedded, It needs to be wrapped. (or a var created and the individual fields set.)

IE I have to return a `api.Watchers204JSONResponse{}` instead of my common `api.NoContentJSONResponse{}` now.

I feel like neither solution is perfect tho as making common code for the struct server may just be limited by go's type system/may need to be revisited after generic methods/type switches.

PR #753, opened 2022-09-13
feat: add config to emit `required:"true"` field tag

xobotyi:
It is handy for cases when openapi spec is generated from sources and source's types reuses types generated from other openapi spec.

### DIFF
--- a/pkg/codegen/configuration.go
+++ b/pkg/codegen/configuration.go
@@ -76,6 +76,7 @@ type OutputOptions struct {
 
 	ExcludeSchemas     []string `yaml:"exclude-schemas,omitempty"`      // Exclude from generation schemas with given names. Ignored when empty.
 	ResponseTypeSuffix string   `yaml:"response-type-suffix,omitempty"` // The suffix used for responses types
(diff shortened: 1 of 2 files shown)


jamietanna (commented):
Would it be possible to add an example of what this looks like? I can't visualise what this would look like

xobotyi:
```go
type StructuresProject struct {
	CreatedAt  int                   `json:"createdAt" required:"true"`
	CreatorID  string                `json:"creatorID" required:"true"`
	Deleted    bool                  `json:"deleted" required:"true"`
}
```

We use https://github.com/swaggest/openapi-go to generate schemas from the code

xobotyi:
So, basically, we use spec generation from internal types, and generates public client and types from generated spec.
Then, other service tries to reuse type generated from spec, but it generated spec it turns optional field since has no `required:"true"` tag

xobotyi:
As for now we're using patched fork but really want to jump back to main package to avoid mess with merging your updates.

xobotyi:
In the end I've managed to make a hook for spec-generator and it utilizes `x-oapi-codegen-extra-tags` to achieve same functionality.

PR #755, opened 2022-09-15
Migrate deprecated io/ioutil package to io and os packages

mycrEEpy:
Since Go 1.16 it's recommended to migrate away from the deprecated io/ioutil package.

This PR replaces io/ioutil by the replacement functions in the io and os packages, see https://pkg.go.dev/io/ioutil


StevenACoffman:
@deepmap-marcinr This works in our production environment.

Issue #758, opened 2022-09-23
Newline missing after many errors printed by oapi-codegen command


nandedamana:
The `oapi-codegen` command does not print a newline after some error messages. Here is an example:
```
nandakumar@nandakumar-laptop:~/git/oapi-codegen/cmd/oapi-codegen$ oapi-codegen api.yml
configuration error: package name must be specifiednandakumar@nandakumar-laptop:~/git/oapi-codegen/cmd/oapi-codegen$ 
```
Expected:
```
nandakumar@nandakumar-laptop:~/git/oapi-codegen/cmd/oapi-codegen$ oapi-codegen api.yml
configuration error: package name must be specified
nandakumar@nandakumar-laptop:~/git/oapi-codegen/cmd/oapi-codegen$ 
```
But the program does print newline after some other errors:
```
nandakumar@nandakumar-laptop:~/git/oapi-codegen/cmd/oapi-codegen$ oapi-codegen
Please specify a path to a OpenAPI 3.0 spec file
nandakumar@nandakumar-laptop:~/git/oapi-codegen/cmd/oapi-codegen$
```
The problem has been identified and is with the usage of `errExit()`. This function doesn't append a newline, while some calling statements do and some others don't.

nandedamana:
Closed by #759

PR #759, opened 2022-09-23
fix: inconsistency in newline printing on error

nandedamana:
Fixes https://github.com/deepmap/oapi-codegen/issues/758
Problematic lines found manually as well as with `grep 'errExit.*\\n\"' oapi-codegen.go`

### DIFF
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -30,6 +30,7 @@ import (
 
 func errExit(format string, args ...interface{}) {
 	_, _ = fmt.Fprintf(os.Stderr, format, args...)
+	fmt.Println("")
 	os.Exit(1)
 }
 
@@ -237,7 +238,7 @@ func main() {
 
 	code, err := codegen.Generate(swagger, opts.Configuration)
 	if err != nil {
-		errExit("error generating code: %s\n", err)
+		errExit("error generating code: %s", err)
 	}
 
 	if opts.OutputFile != "" {
@@ -353,7 +354,7 @@ func updateOldConfigFromFlags(cfg oldConfiguration) oldConfiguration {
 		var err error
 		cfg.ImportMapping, err = util.ParseCommandlineMap(flagImportMapping)
 		if err != nil {
-			errExit("error parsing import-mapping: %s\n", err)
+			errExit("error parsing import-mapping: %s", err)
 		}
 	}
 	if cfg.ExcludeSchemas == nil {
@@ -411,7 +412,7 @@ func newConfigFromOldConfig(c oldConfiguration) configuration {
 
 	templates, err := loadTemplateOverrides(cfg.TemplatesDir)
 	if err != nil {
-		errExit("error loading template overrides: %s\n", err)
+		errExit("error loading template overrides: %s", err)
 	}
 	opts.OutputOptions.UserTemplates = templates
 
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -28,9 +28,16 @@ import (
 	"github.com/deepmap/oapi-codegen/pkg/util"
 )
 
+func withNewline(s string) string {
+	if len(s) > 1 && s[len(s) - 1] != '\n' {
+		return s + "\n"
+	}
+	
+	return s
+}
+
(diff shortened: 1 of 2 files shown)


cmd/oapi-codegen/oapi-codegen.go
```diff
@@ -30,6 +30,7 @@ import (
 
 func errExit(format string, args ...interface{}) {
 	_, _ = fmt.Fprintf(os.Stderr, format, args...)
+	fmt.Println("")
```
jamietanna, line 33: This unfortunately won't work, as it's not printing to `os.Stderr`

jamietanna, line 33: Possibly, we could do:

```suggestion
	_, _ = fmt.Fprintf(os.Stderr, format + "\n", args...)
```

Then it's definitely going to include the newline, or we make sure that anyone calling this method adds the `\n`, which is probably my preference

nandedamana: True. Sorry for being careless about stdout vs stderr.

nandedamana: Yes. Curious to know why do you prefer caller appending the newline. Wouldn't it result in more unterminated error messages (of course, the patch could result in double-terminated messages if the callers don't know, but it could be solved by appending an 'ln' to the function name, like in Println)?

jamietanna: Looking at the code calling it, yep, there are quite a few places where we're not explicitly adding the `\n` at the end, so agreed we need to do something about it.

I think keeping it as `errExit` probably is fine, although I like the use of `ln`.

Maybe we can add some logic that if it doesn't end with a `\n` we append it, then users of the method can specify it explicitly, or it can be added automagically, so we don't end up with duplicates?

nandedamana: Implemented the automagic addition of newline since checking the last char won't be resource intensive or error-prone in Go compared to C. Please check commit f0460847c0d9417423b4ae64f4fc36825cdf7ba4 .

deepmap-marcinr:
I don't like this fix. Maybe we can just find all the missing `\n`'s an add them. Generally, modifying outputs leads to surprises, even though in this case, there is little risk.

Here are the missing newlines.

```
$ find . -name \*.go | xargs grep -Hni errExit  | grep -v '\\n'
./cmd/oapi-codegen/oapi-codegen.go:30:func errExit(format string, args ...interface{}) {
./cmd/oapi-codegen/oapi-codegen.go:128:			errExit("error reading config file '%s': %v", flagConfigFile, err)
./cmd/oapi-codegen/oapi-codegen.go:145:			errExit("error parsing configuration style as old version or new version: %v", err)
./cmd/oapi-codegen/oapi-codegen.go:186:				errExit("error reading config file '%s': %v", flagConfigFile, err)
./cmd/oapi-codegen/oapi-codegen.go:190:				errExit("error parsing'%s' as YAML: %v", flagConfigFile, err)
./cmd/oapi-codegen/oapi-codegen.go:196:			errExit("error processing flags: %v", err)
./cmd/oapi-codegen/oapi-codegen.go:203:				errExit("error reading config file '%s': %v", flagConfigFile, err)
./cmd/oapi-codegen/oapi-codegen.go:207:				errExit("error parsing'%s' as YAML: %v", flagConfigFile, err)
./cmd/oapi-codegen/oapi-codegen.go:219:		errExit("configuration error: %v", err)
./cmd/oapi-codegen/oapi-codegen.go:226:			errExit("error YAML marshaling configuration: %v", err)
./cmd/oapi-codegen/oapi-codegen.go:245:			errExit("error writing generated code to file: %s", err)
```

deepmap-marcinr:
I've fixed this by adding newlines everywhere.

https://github.com/deepmap/oapi-codegen/commit/5413d015ec4a163e223d47fac22909cd65f244f0

nandedamana:
Agree. I regret accepting the suggestion to automagically add the newline. Such things only make sense in app-level stuff, not in a function to print error messages.

However, I would've preferred adding newline to `errExit()` itself (without any magic), because no caller of `errExit()` would've wanted to explicitly avoid the newline (unlike printf, a second call isn't possible here).

deepmap-marcinr:
Thank you, though, for helping make this project better.

PR #780, opened 2022-10-09
chore: Update dependencies

mycrEEpy:
Update of all dependencies ~~which do not break the test suite~~.

~~Seems like `kin-openapi v0.95.0` has some changes which are breaking the tests, therefore it's not included here.~~

EDIT: Also updated kin-openapi which needed some changes in the middleware test assertions.


d-sauer:
Looking forward to `kin-openapi v.104.0` (I have some changes there that I needed in this project of resolving schemas)

jamietanna (approved):

jamietanna:
Thanks for this - been meaning to get around and enable Dependabot to simplify this

Issue #786, opened 2022-10-13
Middleware is applied last to first


wttw:
I'm using a strict chi server.

ChiServerOptions takes a slice of middleware, []MiddleWareFunc. When these are applied they're chained in the order of the slice, meaning that the middleware executes last to first.

I'd expect them to be executed first to last, as most other Go middleware chains do (by iterating through the slice in reverse order).

ericvolp12:
Looks like `chi`'s backwards chaining logic is written in their `chain` function [here](https://github.com/go-chi/chi/blob/master/chain.go#L36). Should be trivial to update in the [chi-middleware template](https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/templates/chi/chi-middleware.tmpl#L180). I'll throw a quick patch together.

ericvolp12:
This should now be supported in the next release (including [this PR](https://github.com/deepmap/oapi-codegen/issues/687)) using the following options in your config YAML:
```yaml
compatibility:
  apply-chi-middleware-first-to-last: true
  ```

This issue can be closed (but I don't have permissions to close it).

Issue #687, opened 2022-07-20
Chi middleware looping logic execute the beginning middlewares at the later part


juddbaguio:
Current implementation server middlewares to a handler goes like this:

```go
var handler = func(w http.ResponseWriter, r *http.Request) {
	siw.Handler.YourRouteHandler(w, r)
}

for _, middleware := range siw.HandlerMiddlewares {
	handler = middleware(handler)
}
```
The snippet above tends to execute beginning middlewares at the later than the last ones.

The looping is supposed to be like this:
```go
var handler = func(w http.ResponseWriter, r *http.Request) {
	siw.Handler.YourRouteHandler(w, r)
}

for i := len(siw.HandlerMiddlewares) - 1; i >= 0; i--  {
	h := siw.HandlerMiddlewares[i]
        if h != nil {
          handler = h(handler)
        }
}
```
The solution is referenced from [Ardan Lab's Service repository on wrapping middlewares](https://github.com/ardanlabs/service/blob/master/foundation/web/middleware.go)

chanced:
Here is the source from chi:

```go
// chain builds a http.Handler composed of an inline middleware stack and endpoint
// handler in the order they are passed.
func chain(middlewares []func(http.Handler) http.Handler, endpoint http.Handler) http.Handler {
	// Return ahead of time if there aren't any middlewares for the chain
	if len(middlewares) == 0 {
		return endpoint
	}


	// Wrap the end handler with the middleware chain
	h := middlewares[len(middlewares)-1](endpoint)
	for i := len(middlewares) - 2; i >= 0; i-- {
		h = middlewares[i](h)
	}


	return h
}
```
https://github.com/go-chi/chi/blob/9db25e5b381fbd6350583b857e74ecaffe355819/chain.go#L34-L49

ericvolp12:
This should now be supported in the next release (including [this PR](https://github.com/deepmap/oapi-codegen/issues/687)) using the following options in your config YAML:
```yaml
compatibility:
  apply-chi-middleware-first-to-last: true
  ```

This issue can be closed (but I don't have permissions to close it).

mromaszewicz:
This has been fixed via the compatibility.apply-chi-middleware-first-to-last configuration option, which reverses the middleware wrapping order so that middlewares execute in the order they are listed. Closing as resolved.

PR #787, opened 2022-10-14
Chain Chi middlewares in reverse order so they execute first-to-last

ericvolp12:
Resolves reversed-ordering of Chi middleware chaining.

After this change, Chi middlewares will be applied in first-to-last ordering i.e. earlier wired/invoked middlewares will run before later wired middlewares.

This addresses https://github.com/deepmap/oapi-codegen/issues/687 and https://github.com/deepmap/oapi-codegen/issues/786

Not sure how I'd go about testing this since the existing unit tests don't seem to have complex middlewares, could use some advice if we want to include a unit test for this.

Note that this will change behavior for Chi users who wired their middlewares in reverse order to compensate, I think most people at this point have adopted version pinning for code generation since we've had some breaking changes in the past few versions but it's worth noting in the release notes.

### DIFF
--- a/pkg/codegen/templates/chi/chi-middleware.tmpl
+++ b/pkg/codegen/templates/chi/chi-middleware.tmpl
@@ -177,8 +177,8 @@ func (siw *ServerInterfaceWrapper) {{$opid}}(w http.ResponseWriter, r *http.Requ
     siw.Handler.{{.OperationId}}(w, r{{genParamNames .PathParams}}{{if .RequiresParamObject}}, params{{end}})
   })
 
-  for _, middleware := range siw.HandlerMiddlewares {
-    handler = middleware(handler)
+  for i := len(siw.HandlerMiddlewares) -1; i >= 0; i-- {
+    handler = siw.HandlerMiddlewares[i](handler)
   }
 
   handler.ServeHTTP(w, r.WithContext(ctx))
--- a/pkg/codegen/templates/chi/chi-middleware.tmpl
+++ b/pkg/codegen/templates/chi/chi-middleware.tmpl
@@ -7,7 +7,7 @@ type ServerInterfaceWrapper struct {
 
 type MiddlewareFunc func(http.Handler) http.Handler
 
-{{range .Operations}}{{$opid := .OperationId}}
+{{range .}}{{$opid := .OperationId}}
 
 // {{$opid}} operation middleware
 func (siw *ServerInterfaceWrapper) {{$opid}}(w http.ResponseWriter, r *http.Request) {
@@ -177,15 +177,9 @@ func (siw *ServerInterfaceWrapper) {{$opid}}(w http.ResponseWriter, r *http.Requ
     siw.Handler.{{.OperationId}}(w, r{{genParamNames .PathParams}}{{if .RequiresParamObject}}, params{{end}})
   })
 
-  {{if $.Options.Compatibility.ApplyChiMiddlewareFirstToLast}}
   for i := len(siw.HandlerMiddlewares) -1; i >= 0; i-- {
     handler = siw.HandlerMiddlewares[i](handler)
   }
-  {{else}}
-  for _, middleware := range siw.HandlerMiddlewares {
-    handler = middleware(handler)
-  }
-  {{end}}
 
   handler.ServeHTTP(w, r.WithContext(ctx))
 }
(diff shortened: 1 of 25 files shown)


ericvolp12:
Oh, I just realized these issues were already addressed in https://github.com/deepmap/oapi-codegen/pull/688 but it didn't end up being followed-up on.

jamietanna (changes requested):
Hey, thanks for this - I think we'll need to make this an opt-in configurable change, rather than a breaking change.

The breaking change _I think_ you're referring to is when we changed the change to how to invoke the generator?

Although, yes, we could've rolled that out in a less breaking fashion, I think that's more noticeable as a breaking change to us subtly changing underlying behaviour, which may not be clear to consumers, and break a lot of applications.

ericvolp12:
I was more referencing this rollback release after we broke the type signature of some of the generated code for Gin - https://github.com/deepmap/oapi-codegen/releases/tag/v1.8.1

I'll look into turning this into a configuration-driven change though it might be a few days before I can get around to it.

ericvolp12:
Okay JK I got bored and decided to work on this.

The changes as shown now pass a new `TemplateParameters` object into the template engine which include an `Operations` property that is the old `ops` object that used to be the root of most templates (a `[]OperationDefinition`) as well as an `Options` property which is just the whole `Configuration` object.

This allows us to use Configuration values inside the template to conditionally render the template in different ways.

In this case, all I've done is gated the new mechanism for Chi middleware chaining behind the new Compatibility configuration variable `Options.Compatibility.ApplyChiMiddlewareFirstToLast`.

```go
{{if $.Options.Compatibility.ApplyChiMiddlewareFirstToLast}}
for i := len(siw.HandlerMiddlewares) -1; i >= 0; i-- {
  handler = siw.HandlerMiddlewares[i](handler)
}
{{else}}
  handler = middleware(handler)
}
{{end}}
```

To start, we can have this be an opt-in compatibility option but using this new method of template rendering, it should be trivial to make the first-to-last chaining order an opt-out when we decide enough time has passed to establish it as the default.

This will also open up the ability to have template-specific feature flags in configuration that don't require maintaining lots of additional templates but just feature Option gating logic inside the templates instead.

deepmap-marcinr:
That's a lot of work you did, but options are globally stored on `globalstate.options`, precisely because it's a pain to pass them around everywhere, so your change can be done without all the changes to all the templates. All you really need to change is that one chi template and pass it an additional flag in its context, rather than changing all of them.

ericvolp12:
> That's a lot of work you did, but options are globally stored on `globalstate.options`, precisely because it's a pain to pass them around everywhere, so your change can be done without all the changes to all the templates. All you really need to change is that one chi template and pass it an additional flag in its context, rather than changing all of them.

Okay, I didn't see that getting injected as a Template Func, thanks for the pointer!

In other news, I learned what Go Template Funcs are today!

Hopefully this change is a lot simpler and makes more sense, the activation functionality is the same but we're eschewing all my previous changes that were injecting options a second time.

ericvolp12:
I saw this line while Go Generating locally:
```
go generate ./...
go: downloading github.com/matryer/moq v0.2.7
```

Looks like something got updated that the CI is unhappy with there?

deepmap-marcinr:
> 

It looks like a dependency got updated, it's no big deal, you just haven't fetched it before.

ericvolp12:
> It looks like a dependency got updated, it's no big deal, you just haven't fetched it before.

Does this look okay to merge now or is there anything else you'd like me to change?

pkg/codegen/templates/chi/chi-middleware.tmpl
```diff
@@ -177,9 +177,15 @@ func (siw *ServerInterfaceWrapper) {{$opid}}(w http.ResponseWriter, r *http.Requ
     siw.Handler.{{.OperationId}}(w, r{{genParamNames .PathParams}}{{if .RequiresParamObject}}, params{{end}})
   })
 
+  {{if opts.Compatibility.ApplyChiMiddlewareFirstToLast}}
```
deepmap-marcinr, line 180: I must be missing something - how does `opts` get passed in here? When we call GenerateChiServer, we pass in an array of OperationDefinition, which doesn't have opts.

deepmap-marcinr: Oh, nevermind, I see it. This code has changed so much since I wrote it, I get lost sometimes.

deepmap-marcinr:
Yep, I merged it, thank you.

I'm looking to do a new release soon, just waiting on some PR wrapup.

PR #789, opened 2022-10-15
chore: Fix comments/imports

Jleagle:
Fixed some function names with underscores
Sorted imports
Made comments idiomatic
Removed some unnecessary types


deepmap-marcinr:
It looks like you have reformatted generated code, which will be undone when it's regenerated - and that's why there's a precommit failure. Our autogenerated code works, but it's really not suitable for style checking, too many little issues, but it works ok, it's just not compliant with pedantic linters.

I'm happy to merge this, but please undo reformatting of anything with `*.gen.go` in the name.

Jleagle:
Tests now passing, didn't mean to format the generated code.

PR #808, opened 2022-10-23
Add extra JSON type checking

AnnaDodson:
- Following from #802, adding extra check for different JSON content types as a catchall to catch anything with JSON suffix


jamietanna (approved):
Thanks!

Issue #479, opened 2021-11-09
Optional parameter with default value should not be treated as pointers


ilya-direct:
Schema
```
schema:
  type: object
  properties:
      name:
          type: string
      comment:
          type: string
          default: ""
  required:
      - name
```

Generated as:
```
type AddItemJSONBody struct {
        Name    string `json:"name"`
	Comment    *string `json:"comment,omitempty"`
}
```

But it would be useful to be (without a pointer):
```
type AddItemJSONBody struct {
        Name    string `json:"name"`
	Comment    string `json:"comment,omitempty"`
}
```

within `json.unmarshal`  empty `comment` should be assigned to `default` value.

PS pointer values must be checked for  `nil` and that is an additional pain.

MateusFrFreitas:
I'm facing the same issue. Do you have any solution?

veleek:
I'm running into a similar issue.  I've got an optional property that I want to be marshaled as a string same as `Comment` and I was hoping that setting `x-nullable` false would force the removal of the pointer.  

Would that be a reasonable change to make?  If `x-nullable` is false, but the value is not provided, then it should just fall back to the default if provided, or the zero value for the type, which the json marshalling should do automatically.

veleek:
It looks like this old PR also does something similar by adding support for `x-go-nullable` which forces a non-pointer when the value is false (or empty).

scottshotgg:
@deepmap-marcinr 
Is there any way this could be merged/supported?

deepmap-marcinr:
This one is tricky, because the json marshaler is not aware of the fact that there should be a default value when unmarshaling. So, if we make a field a non-pointer type, then you will have no way to tell apart in your own code whether it's been provided but set to the zerovalue or not provided at all.

I would love all fields with defaults to be non-pointers, it's nicer to work with, however, before we have that, we need to have a spec-aware json marshaler. I've got the beginnings of one in an experimental code tree, but IMO, that's a hard requirement before we add undeterminable state.

So far, the choice has been to use uglier models but complete determinism.

scottshotgg:
Thanks for the comment. I'll check out the experimental branch if I can find it. 

For now I've forked and incorporated some of the changes from #146. This works for now; I would much rather work with pointers for simple types, especially since I've got a middleware that validates the JSON object itself before reaching any app code.

dnesting:
I've always made good use of protobuf's accessors as a solution to this problem:

> For either of these field definitions:
>
> ```
[...]

PR #823, opened 2022-10-26
Enable forcing non-pointer types using Nullable

veleek:
Proposed update to GoTypeDef to enable forcing a non pointer type. This is useful in go to avoid using `*string` instead of `string` in most cases where the value is not required, but there's no need to differentiate between `nil` and empty string values.

There may need to be some more semantics around this in order to force the pointer in some cases where `p.Nullable` is true.  We probably need some discussion around this prior to merging.

See also #479

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -93,8 +93,8 @@ func (p Property) GoFieldName() string {
 
 func (p Property) GoTypeDef() string {
 	typeDef := p.Schema.TypeDecl()
-	if !p.Schema.SkipOptionalPointer &&
-		(!p.Required || p.Nullable ||
+	if p.Nullable && !p.Schema.SkipOptionalPointer &&
+		(!p.Required ||
 			(p.ReadOnly && (!p.Required || !globalState.options.Compatibility.DisableRequiredReadOnlyAsPointer)) ||
 			p.WriteOnly) {
 


jamietanna:
Thanks @veleek for this! This may no longer be necessary, especially as with #1404, we'll be introducing a separate type that can be opted-in for better Nullable support :crossed_fingers:

ghen:
May we consider adding this as a default behaviour?
At least have an option in generator output configuration to enable this as a default behaviour.

We have many YAML spec files, and annotating 1000s of properties with `x-go-type-skip-optional-pointer` is a no go.

Technically, the generator should respect a built-in `nullable` spec that it ignores at the moment, even when it is set explicitly.

In other word for the sollowing schema:
```yaml
schema:
  type: object
  properties:
      nonNullable:
          type: string
      nonNullableExplicit:
          type: string
          nullable: false
      nullable:
          type: string
          nullable: true
```

The expected struct:
```go
type JSONBody struct {
        nonNullable          string `json:"nonNullable,omitempty"`
	nonNullableExplicit  string `json:"nonNullableExplicit,omitempty"`
        nullable             *string `json:"nullable,omitempty"`
}
```

Instead it produces pointers everywhere:
```go
type JSONBody struct {
        nonNullable         *string `json:"nonNullable,omitempty"`
	nonNullableExplicit *string `json:"nonNullableExplicit,omitempty"`
        nullable            *string `json:"nullable,omitempty"`
}
```

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the `nullable-type` configuration option now exists. When enabled, nullable fields use `nullable.Nullable[type]` wrapper instead of pointer types. This PR's intent is fully implemented.

PR #825, opened 2022-10-27
Add x-dynamodb field tag

anthonyLock:
This PR adds the x-dynamodb as a option similar to x-oapi-codegen-extra-tags, it adds the dynamodbav tag on which is used by dynamodb
However if you give the x-dynamodb filed omitempty then the same logic for json tags applies to dynamodbav


deepmap-marcinr:
I need to think about how we can do this generically. I'd like to have anyone be able to annotate the spec with their own field tags without having to change the code generator.

stevenh:
What about using plugin modules?

53jk1 (approved):

jamietanna:
I think we're going to close this off as it is - I'll look at making it something that we could define within https://github.com/deepmap/oapi-codegen/pull/851 (or similar) instead

Issue #829, opened 2022-10-28
Gin strict handler uses Bind which causes the request to abort without error


jeroendk:
Gin `Bind` (`MustBind` under the hood) aborts the request immediately and sends a 400 error as plain text to the client. 

```
  if err := ctx.Bind(&body); err != nil {
      ctx.Error(err)
      return
  }
```
This prevents any custom error handler from setting the proper content-type headers. 

My suggestion would be to use `ShouldBind` instead and set the response status code for backward compatibility:

```
if err := ctx.ShouldBind(&body); err != nil {
    ctx.Status(http.StatusBadRequest)
    ctx.Error(err)
    return
}
```

mromaszewicz:
PR #830 (commit 050c4bfe, merged 2022-11-01) was submitted by the same reporter and switched the strict-gin template from ctx.Bind (which calls MustBind and aborts with 400 plain-text) to ctx.ShouldBindJSON, with errors delegated to the configurable RequestErrorHandlerFunc. The current template at pkg/codegen/templates/strict/strict-gin.tmpl:82 confirms ShouldBindJSON is in use. The issue is fully resolved.

Closing as fixed. If you can still reproduce on a recent release, please reopen with the exact spec and version used.

PR #830, opened 2022-10-28
829 fix: shouldBind instead of Bind to prevent abort

jeroendk:
Fix for #829

Use shouldBind instead of (must)Bind to allow custom error response.


deepmap-marcinr:
Please regenerate all generated code to fix precommit, then I'm happy to merge.

PR #840, opened 2022-11-07
Make external ref internalization optional

kkroo:
This disables the call to `InternalizeRefs` by default in `GenerateInlinedSpec`.

Currently, internalizing external references results in broken validator due to https://github.com/getkin/kin-openapi/issues/618 and is unnecessary because of `import-mapping` configuration feature. 

Not sure if the call to `InternalizeRefs` is an artifact, so added a flag to enable in output options but can remove it all together as well per the maintainers' discretion.

### DIFF
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -233,8 +233,12 @@ func Generate(spec *openapi3.T, opts Configuration) (string, error) {
 	}
 
 	var inlinedSpec string
+	var refNameResolver openapi3.RefNameResolver
+	if opts.OutputOptions.InternalizeRefs {
+		refNameResolver = openapi3.DefaultRefNameResolver
+	}
 	if opts.Generate.EmbeddedSpec {
-		inlinedSpec, err = GenerateInlinedSpec(t, importMapping, spec)
+		inlinedSpec, err = GenerateInlinedSpec(t, importMapping, spec, refNameResolver)
 		if err != nil {
 			return "", fmt.Errorf("error generating Go handlers for Paths: %w", err)
 		}
--- a/pkg/codegen/inline.go
+++ b/pkg/codegen/inline.go
@@ -26,9 +26,11 @@ import (
 
 // GenerateInlinedSpec generates a gzipped, base64 encoded JSON representation of the
 // swagger definition, which we embed inside the generated code.
-func GenerateInlinedSpec(t *template.Template, importMapping importMap, swagger *openapi3.T) (string, error) {
+func GenerateInlinedSpec(t *template.Template, importMapping importMap, swagger *openapi3.T, refNameResolver openapi3.RefNameResolver) (string, error) {
 	// ensure that any external file references are embedded into the embedded spec
-	swagger.InternalizeRefs(context.Background(), nil)
+	if refNameResolver != nil {
+		swagger.InternalizeRefs(context.Background(), refNameResolver)
+	}
 	// Marshal to json
 	encoded, err := swagger.MarshalJSON()
 	if err != nil {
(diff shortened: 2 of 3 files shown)


pkg/codegen/codegen.go
```diff
@@ -233,8 +233,12 @@ func Generate(spec *openapi3.T, opts Configuration) (string, error) {
 	}
 
 	var inlinedSpec string
+	var refNameResolver openapi3.RefNameResolver
+	if opts.OutputOptions.InternalizeRefs {
+		refNameResolver = openapi3.DefaultRefNameResolver
+	}
```
stevenh, line 239: Unfortunately this changes the API so isn't compatible and would need a major version increment. See `inline.go` comment for how to solve this.
```suggestion
```

pkg/codegen/codegen.go
```diff
@@ -233,8 +233,12 @@ func Generate(spec *openapi3.T, opts Configuration) (string, error) {
...
 	if opts.Generate.EmbeddedSpec {
-		inlinedSpec, err = GenerateInlinedSpec(t, importMapping, spec)
+		inlinedSpec, err = GenerateInlinedSpec(t, importMapping, spec, refNameResolver)
```
stevenh, line 241: As above
```suggestion
		inlinedSpec, err = GenerateInlinedSpec(t, importMapping, spec)
```

pkg/codegen/inline.go
```diff
@@ -26,9 +26,11 @@ import (
 
 // GenerateInlinedSpec generates a gzipped, base64 encoded JSON representation of the
 // swagger definition, which we embed inside the generated code.
-func GenerateInlinedSpec(t *template.Template, importMapping importMap, swagger *openapi3.T) (string, error) {
+func GenerateInlinedSpec(t *template.Template, importMapping importMap, swagger *openapi3.T, refNameResolver openapi3.RefNameResolver) (string, error) {
```
stevenh, line 29:
```suggestion
func GenerateInlinedSpec(t *template.Template, importMapping importMap, swagger *openapi3.T) (string, error) {
```

pkg/codegen/inline.go
```diff
@@ -26,9 +26,11 @@ import (
...
 	// ensure that any external file references are embedded into the embedded spec
-	swagger.InternalizeRefs(context.Background(), nil)
+	if refNameResolver != nil {
+		swagger.InternalizeRefs(context.Background(), refNameResolver)
+	}
```
stevenh, line 33: While not perfect we can use the `globalState` to achieve the same thing without breaking the API.
```suggestion
	if globalState.OutputOptions.InternalizeRefs {
		swagger.InternalizeRefs(context.Background(), openapi3.DefaultRefNameResolver)
	}
```

stevenh (changes requested):

stevenh (approved):
LGTM

kkroo:
How can I get this merged?

stevenh:
Needs either the owner or one of the collaborators to review, approve and merge.

jamietanna:
Hey @kkroo this was added as part of https://github.com/deepmap/oapi-codegen/pull/771 due to cases where a non-import-mapped specification broke, i.e. one where we're `$ref`ing a `schema` ([example](https://github.com/co-cddo/federated-api-model/blob/335e40bfc1d19d1208d6b7dc5fffed7060af7aae/schemas/v1alpha/openapi.yml)) at which point no validation issues were seen.

I'll try and get a look at this soon, do you have an example spec where this is breaking?

jamietanna:
Hey @kkroo friendly nudge, as I've just seen this again, and I didn't get a reply last time. NP if it's not still a priority for you

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the codebase has been significantly restructured since this was written. While the concept still has merit, this PR is too outdated to merge — any work would need to be done fresh against the current codebase.

Issue #841, opened 2022-11-07
The order of the middlewares is reversed in gorilla/mux implementation


mpls104:
Hi, thank you for adding [support the code generation for gorilla/mux](https://github.com/deepmap/oapi-codegen/issues/465). 

I've tried this feature and noticed that the order of the middlewares passed to `HandlerWithOptions`  are reversed when they applied.

Like the implementation of middleware in gorilla mux, how about changing the order?
https://github.com/gorilla/mux/blob/3cf0d013e53d62a96c096366d300c84489c26dd5/mux.go#L143-L145 

Please let me hear your opinion @jamietanna because you have worked for supporting gorilla/mux a lot!

Here's my implementation of this issue. https://github.com/deepmap/oapi-codegen/pull/842

PR #842, opened 2022-11-07
Reverse the order of the middleware when they applied in gorilla/mux implementation

mpls104:
resolve https://github.com/deepmap/oapi-codegen/issues/841

### DIFF
--- a/examples/petstore-expanded/gorilla/api/cfg.yaml
+++ b/examples/petstore-expanded/gorilla/api/cfg.yaml
@@ -5,4 +5,4 @@ generate:
   embedded-spec: true
 output: petstore.gen.go
 compatibility:
-  apply-gorilla-mux-middleware-first-to-last: true
+  apply-gorilla-middleware-first-to-last: true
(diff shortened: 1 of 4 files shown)


stevenh (approved):
LGTM

jamietanna:
Thanks for this - looks like this was because I copy-pasta'd Chi's server, and was resolved there in #787

jamietanna (changes requested):
Looks good - only requested change is that similar to [the Chi PR](https://github.com/deepmap/oapi-codegen/pull/787), we add a compatibility option to allow folks to opt-in to these?

mpls104:
@jamietanna 
Addressed in 4b9090ff9ee59594b1555dabb7c072925b7389c4

jamietanna (approved):
Looks good, thanks for the changes :raised_hands: 

Only thought I have is whether we should rename the flag:

```suggestion
-apply-gorilla-mux-middleware-first-to-last
+apply-gorilla-middleware-first-to-last
```

Given the configuration to generate gorilla/mux servers is `gorilla-server`?

mpls104:
Understood. Since `GenerateOptions` has already named as `gorilla-server`, let's omit the word `mux`. I'll push commit to fix them.

PR #844, opened 2022-11-10
chore: use switch for types

stevenh:
Use switch statements for types to reduce code needed and increase readability.

Also eliminate errors.New on an fmt.Sprintf, replacing with fmt.Errorf.


stevenh:
Replacing with lint PR

PR #845, opened 2022-11-10
ci: enable golangci-lint and fix errors

stevenh:
Enable golangci-lint fixing the errors that it highlights.

This addresses:
* Missing error checks.
* Duplicate code.
* Direct error comparison
* Error wrapping.
* Test helpers.
* Document comments.
* Missing http Request.Body.Close()
* Switches instead of complex ifs.
* Use strings.ReplaceAll instead of strings.Replace with -1.
* Unnecessary blank lines.
* Duplicate literal strings.
* Unneeded casts.
* Removed inits.

Also:
* Merge ci steps into one so it runs quicker.


jamietanna:
Thanks for this - we'll stick with #722 in the meantime.

There are some things here that may be worth resolving, so we'll pick these up as a follow-up.

stevenh:
Thanks for the consideration, I'll look to bring this back once #722 is merged as there quite a few more fixes in this which are worth getting committed, in particular the error checks.

PR #847, opened 2022-11-11
feat: template for parse response

stevenh:
Use a template for parsing response instead of `genResponseUnmarshal` allowing users to customise or fix quickly without having to wait for a code change and new release of `oapi-codegen`.

This new approach also uses nested switch statements to guarantee correct flow and minimise `strings.Contains` calls and if decoding fails returns a wrapped error describing what decode failed.

This adds a few new template helpers:
* `hasSuffix` - same as `strings.HasSuffix`
* `hasPrefix` - same as `strings.HasPrefix`
* `split` - same as `strings.Split`
* `dict` - adds / creates key value pairs into a string indexed map.
* `get` - get value from a map created by dict.
* `set` - set a value for a key in a map created by dict.
* `sum` - sums all passed values

Example output:
```go
contentType := rsp.Header.Get("Content-Type")
switch {
case strings.Contains(contentType, "json"):
	switch {
	case rsp.StatusCode == 200:
		var dest GenericObject
			return nil, fmt.Errorf(`decode json for "GenericObject": %w`, err)
		}

		resp.JSON200 = &dest
	default:
		var dest GenericObject
			return nil, fmt.Errorf(`decode json for "GenericObject": %w`, err)
		}

		resp.JSONDefault = &dest
	}
case strings.Contains(contentType, "xml"):
	switch {
	case rsp.StatusCode == 200:
		var dest GenericObject
		if err := xml.Unmarshal(bodyBytes, &dest); err != nil {
			return nil, fmt.Errorf(`decode xml for "GenericObject": %w`, err)
		}

		resp.XML200 = &dest
	}
}
```

Also improve stringToGoCommentWithPrefix to:
* Remove leading space on the first line.
* Remove duplicate prefixes when adding, including <prefix> the <prefix>.
* Lower case the first letter of the first word when adding prefix.

Example comment fixes:
```patch
-       // Name Name of the pet
+       // Name of the pet.

-       // Id Unique id of the pet
+       // Id unique id of the pet.
```


deepmap-marcinr:
This is an interesting PR, and kind of in the opposite direction that I was thinking of heading, since I've been hoping to move more code into Go, out of templates, for flexibility.

For example, I'd like to provide overrides for content-types where users can specify for which content type they'd like to generate models. This means hardcoding strings in templates is a no-go, but easily fixed with a "shouldGenerateModels(content-type)" kind of function.

On the upside, allowing an override definitely allows for quicker fixes to issues.

stevenh:
> This is an interesting PR, and kind of in the opposite direction that I was thinking of heading, since I've been hoping to move more code into Go, out of templates, for flexibility.
> 
> For example, I'd like to provide overrides for content-types where users can specify for which content type they'd like to generate models. This means hardcoding strings in templates is a no-go, but easily fixed with a "shouldGenerateModels(content-type)" kind of function.
> 
> On the upside, allowing an override definitely allows for quicker fixes to issues. 

As a user I'd definitely say that less code in go and more in templates / config is a benefit as it allows fixes and customisations without the delay of having to fork the repo, submit a PR and wait for it to be merged and released.

If I understand you correctly it sounds like you would like the ability for the user to set a configuration option that would control the generation of specific content-type code?

53jk1 (approved):

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the user-templates system now allows complete template overrides, which achieves the same customization goal through a different mechanism.

PR #851, opened 2022-11-11
feat(config): add type mapping

stevenh:
Add the ability to override type the default type mapping behaviour so that users can easily provide global control of their go type generation without having to use x-go-type for every occurrence.

### DIFF
--- a/pkg/codegen/configuration.go
+++ b/pkg/codegen/configuration.go
@@ -82,6 +82,32 @@ type OutputOptions struct {
 	ExcludeSchemas     []string `yaml:"exclude-schemas,omitempty"`      // Exclude from generation schemas with given names. Ignored when empty.
 	ResponseTypeSuffix string   `yaml:"response-type-suffix,omitempty"` // The suffix used for responses types
 	ClientTypeName     string   `yaml:"client-type-name,omitempty"`     // Override the default generated client type with the value
+
+	// TypeMappings provides the ability to configure how types are mapped
+	// including overriding the default behaviour.
+	// Entries keys are in the format: <schemaType>:<schemaFormat>.
+	//
+	// An example of a custom type for type: "string", format: "uuid":
+	//   type-mappings:
+	//     string-uuid:
+	//       skip-optional-pointer: true
+	//       define-via-alias: true
+	TypeMappings map[string]TypeMapping `yaml:"type-mappings,omitempty"`
+}
+
+// TypeMapping defines a schema type mapping override.
+type TypeMapping struct {
+	// GoType is the go type that will be used to represent the schema.
+	GoType string `yaml:"go-type"`
+
+	// SkipOptionalPointer should be set t true if we don't need a pointer type
+	// even when they are optional.
+	SkipOptionalPointer bool `yaml:"skip-optional-pointer"`
+
+	// DefineViaAlias if true schema will declare a type via alias
+	// `type Foo = bool`, otherwise the type definition will be a classic
+	// `type Foo bool`
+	DefineViaAlias bool `yaml:"define-via-alias"`
 }
 
 // UpdateDefaults sets reasonable default values for unset fields in Configuration
--- a/pkg/codegen/schema_test.go
+++ b/pkg/codegen/schema_test.go
@@ -1,0 +1,120 @@
+package codegen
+
+import (
+	"testing"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/stretchr/testify/require"
+)
+
+type testSchema struct {
+	schema openapi3.Schema
+	path   []string
+	want   Schema
+	err    string
+}
+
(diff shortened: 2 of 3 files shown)


53jk1 (approved):

jamietanna:
Apologies that we're very late to seeing this - will make sure we get to it for the next+1 release.

If you fancy addressing conflicts that'd be appreciated, but we're happy to take it from here too

stevenh:
Thanks @jamietanna, better late than never 😄

All rebased and conflicts fixed.

jamietanna:
Thank you! Will see if I can get it into the next release, but chances are it'll be next+1

jamietanna:
Woops, I broke that :sweat_smile: 

I've got a few things to tweak on the PR, so will resolve them now :+1:

examples/outputoptions/typemappings/cfg.yaml
```diff
@@ -0,0 +1,13 @@
...
+package: typemappings
+output: gen.go
+generate:
+  models: true
+output-options:
+  # to make sure that all types are generated, even if they're unreferenced
+  skip-prune: true
+  type-mappings:
+    string-email:
+      define-via-alias: true
```
jamietanna, line 11: @stevenh got a good example of what this would look like when generated? I don't see any difference right now :thinking:

pkg/codegen/configuration.go
```diff
@@ -221,8 +222,64 @@ type OutputOptions struct {
...
+// TypeMapping defines a schema type mapping override.
+type TypeMapping struct {
+	// GoType is the Go type that will be used to represent the schema.
+	GoType *string `yaml:"go-type,omitempty"`
+
+	// SkipOptionalPointer should be set to true if we don't need a pointer type even when they are optional.
+	SkipOptionalPointer bool `yaml:"skip-optional-pointer"`
+
+	// DefineViaAlias if true schema will declare a type via alias `type Foo = bool`, otherwise the type definition will be a classic `type Foo bool`
+	DefineViaAlias bool `yaml:"define-via-alias"`
```
jamietanna, line 274: these should be optional too

examples/outputoptions/typemappings/gen_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/oapi-codegen/runtime/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestClient(t *testing.T) {
+	var c Client
+
+	// it is the UUID type, and it is not defined with the pointer
+	assert.Equal(t, types.UUID{}, c.Uuid)
+
```
jamietanna, line 15:
```suggestion
```

pkg/codegen/schema_test.go
```diff
@@ -454,3 +456,119 @@ func TestProperty_GoTypeDef_nullable(t *testing.T) {
...
+			GoType:              ptr("mypkg.UUID"),
+			SkipOptionalPointer: ptr(true),
+		},
+	}
+	defer func() { globalState.options.OutputOptions.TypeMappings = old }()
+
+	tests = make(map[string]*testSchema)
+	addTest("string", "uuid", "mypkg.UUID", "")
+	tests["string-uuid"].want.SkipOptionalPointer = true
+	tests["string-uuid"].want.DefineViaAlias = false
```
jamietanna, line 555: Why do we think `DefineViaAlias` should be `false` here?

(Think I've broken the test, but want to be sure as to _why_ we've done this)

stevenh: Yer with it being optional it needs to be defined in the test as it no longer gets the default value from the type, and being a string this is set to true.

stevenh:
@jamietanna all issues should be resolved.

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because type mapping has been implemented in the codebase through a different architecture than what this PR proposed. The PR is too outdated to merge.

stevenh:
> Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because type mapping has been implemented in the codebase through a different architecture than what this PR proposed. The PR is too outdated to merge.

Thanks for the update @mromaszewicz could you clarify how the new architecture allows the user to override a type without having to add x-go-type to every occurance?

For example the current system uses a uuid type for uuid's but we need them to be strings, for legacy reasons.

mromaszewicz:
Take a look at this config file. I have changed the default type for `integer` and overwrote globally how `date` is handled.
https://github.com/oapi-codegen/oapi-codegen/blob/main/examples/output-options/type-mapping/config.yaml

mromaszewicz:
This is not merged yet, but here are corresponding docs:
https://github.com/mromaszewicz/oapi-codegen/blob/docs/docs/configuration.md#type-mapping

PR #857, opened 2022-11-14
Fix error handler for required query parameters

abdollahpour:
If at the parameters of the API, it has a required field in queries, the generated code breaks since the ErrorHandler does not have any return value. But in the template, it has been used in a condition (expects a boolean return).
This fix simply changes the usage aligned with the other part of the template.


alexstojda:
Also encountering this error and would appreciate this PR being merged 🙏🏼

jamietanna:
FYI @deepmap-marcinr - happy if I merge + release this as a patch release?

jamietanna (approved):

deepmap-marcinr:
@jamietanna  - sure, make a patch release.

jamietanna:
https://github.com/deepmap/oapi-codegen/releases/tag/v1.12.3 🚀

PR #861, opened 2022-11-17
readme.md, fixed demo code

lintaba:
Demo generating code caused issues. Had been able to find 1 solution in a Japanese page. Maybe it will be more useful for beginners in this way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ into objects which match the OpenAPI 3.0 definition. The code generator in this
 directory does a lot of that for you. You would run it like so:
 
     go install github.com/deepmap/oapi-codegen/cmd/oapi-codegen@latest
-    oapi-codegen petstore-expanded.yaml > petstore.gen.go
+    oapi-codegen petstore-expanded.yaml -package packagename -o petstore.gen.go
 
 Let's go through that `petstore.gen.go` file to show you everything which was
 generated.


README.md
```diff
@@ -40,7 +40,7 @@ into objects which match the OpenAPI 3.0 definition. The code generator in this
 directory does a lot of that for you. You would run it like so:
 
     go install github.com/deepmap/oapi-codegen/cmd/oapi-codegen@latest
-    oapi-codegen petstore-expanded.yaml > petstore.gen.go
+    oapi-codegen petstore-expanded.yaml -package packagename -o petstore.gen.go
```
stevenh, line 43: The original example runs fine for me, what error did you see?

lintaba: Error message: **configuration error: package name must be specified**

```bash
$ go version
go version go1.19.3 darwin/amd64
$ go install github.com/deepmap/oapi-codegen/cmd/oapi-codegen@latest
$ ls
specs.yaml
$ oapi-codegen specs.yaml 
configuration error: package name must be specified
$ oapi-codegen -pac
[specs.txt](https://github.com/deepmap/oapi-codegen/files/10079263/specs.txt)
kage packagename specs.yaml >/dev/null
```
a minimal specs.yaml has been attached.

let me know if you need any further info

stevenh: Thanks for the example, I had an older version and looks like the [configuration changes introduced a regression](https://github.com/stevenh/oapi-codegen/commit/6e66df9a62903962a12b6d1a059e6a0ba6699cd7) so there's no default.

I've done a quick PR https://github.com/deepmap/oapi-codegen/pull/873 which fixes this regression as well as improving it for named files, which is the common case.

lintaba: thanks.
Could you send me something about how to test it? (not sure how to install a package from a different branch/repo than the default one, havent done before)

Kejam: I noticed if you used openapi right command will be: 

oapi-codegen -package api -o api/name.go api/nameyaml

Otherwise I cathed: Only one OpenAPI 3.0 spec file is accepted and it must be the last CLI argument

Spent one hour for solved all problem:(

stevenh: > thanks. Could you send me something about how to test it? (not sure how to install a package from a different branch/repo than the default one, havent done before)

The following should be general gist:
```shell
git remote add stevenh https://github.com/stevenh/oapi-codegen.git
git fetch stevenh
git checkout fix/package-detect
go install ./cmd/oapi-codegen
```

kkettinger:
I think it is still relevant, as the example in the readme doesn't work.

bkane-msft:
@deepmap-marcinr , I'm also running into this bug, simply by following what's in the README

stevenh:
What error are you seeing @kkettinger want to make sure its the same thing.

PR #864, opened 2022-11-19
Fix error handling when empty query string passed

SogoKato:
This fixes an error that occurs when using query parameter with gin.

Though siw.ErrorHandler returns no value, the template code tries to use the return value.

```
siw.ErrorHandler(c, fmt.Errorf("Query argument <ParamName> is required, but not found: %s", err), http.StatusBadRequest) (no value) used as valuecompiler[TooManyValues](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#TooManyValues)
```

Change of this p-r fixes the error in my environment.


SogoKato:
duplicate of #857, close this p-r.

jamietanna:
https://github.com/deepmap/oapi-codegen/releases/tag/v1.12.3 🚀

PR #888, opened 2022-12-07
Add docker build on release

insidieux:
- Add Dockerfile
- Add GitHub workflow running on `release published` event


Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    CGO_ENABLED=0
+COPY . /go/src/github.com/deepmap/oapi-codegen
+WORKDIR /go/src/github.com/deepmap/oapi-codegen
+RUN go build -o ./bin/oapi-codegen ./cmd/oapi-codegen/oapi-codegen.go
+
+### Image
+FROM alpine:${ALPINE_VERSION} as image
+COPY --from=build-binary /go/src/github.com/deepmap/oapi-codegen/bin/oapi-codegen /usr/local/bin/oapi-codegen
+RUN chmod +x /usr/local/bin/oapi-codegen
+ENTRYPOINT ["/usr/local/bin/oapi-codegen"]
```
jamietanna, line 17: Is it possible for this to be `FROM scratch`?

jamietanna:
What are your thoughts, @deepmap-marcinr? Not sure if this fits within your thoughts about what we'd want to maintain in this project, but looks like a few folks are interested in it.

Dockerfile
insidieux: Done!

deepmap-marcinr:
What's the reasoning behind this? To `docker run oapi-codegen` instead of using `go install` to get it?

insidieux:
> What's the reasoning behind this? To `docker run oapi-codegen` instead of using `go install` to get it?
For example to use it in CI, as docker image for stage. 
Also i don't need to control and check what the version is currently installed, just use docker image with tag I need.

insidieux:
So? Any updates? 
If you are struggle to prevent providing Dockerfile and image, I can just make my project with dependabot and reference to yours repo)
Imho, project could be fully self-maintained and let users  more convenient service of usage)

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ENV GOOS=linux \
+    GOARCH=amd64 \
+    CGO_ENABLED=0
+COPY . /go/src/github.com/deepmap/oapi-codegen
+WORKDIR /go/src/github.com/deepmap/oapi-codegen
+RUN go build -o ./bin/oapi-codegen ./cmd/oapi-codegen/oapi-codegen.go
+
+### Image
+FROM scratch as image
+COPY --from=build-binary /go/src/github.com/deepmap/oapi-codegen/bin/oapi-codegen /usr/local/bin/oapi-codegen
```
deepmap-marcinr, line 15: We probably want to run as non-root user.

deepmap-marcinr:
How is this going to look like in terms of releases to ghcr or docker hub or whatever? It looks like it's automatic whenever I push a release - but now, I have to worry about exceeding storage or access quotas, because people will complain when this fails.

I'm open to doing something like this, the process with github actions looks simple, but I think maybe I'd need to push it to dockerhub after setting up my own account there. Is there any particular reason to use one over the other? I've only ever consumed from public docker repositories, never produced.

Dockerfile
insidieux: Ok, i will update PR later and trigger you

insidieux:
> How is this going to look like in terms of releases to ghcr or docker hub or whatever? It looks like it's automatic whenever I push a release - but now, I have to worry about exceeding storage or access quotas, because people will complain when this fails.
> 
> I'm open to doing something like this, the process with github actions looks simple, but I think maybe I'd need to push it to dockerhub after setting up my own account there. Is there any particular reason to use one over the other? I've only ever consumed from public docker repositories, never produced.

Depends on you want to use =) I can provide Action for push to Dockerhub, if you want =) And there is cool feature Automated builds in Dockerhub, but it works only for paid accounts.

Dockerfile
insidieux: Done! It will be run as oapi-codegen user

kbakdev (approved):

Mxrk:
Any updates here? Would like to see it happening.

DrBlury:
I'd like to see an update here as well. Having a docker image drastically improves adoption. For some companies/developers it's a no-go if they have to check out a repo or install something in a not so streamline way. Docker helps with that. All the cli tools you want/need can be just inside containers and don't bloat the host system. They can be easily removed as well and using other versions of the software doesn't require and weird procedures. Just changing the version tag.

insidieux:
So? Seems like not only me ask for this feature =) But 6 months passed from first push... 
Any updates? Or I must to create just side project with dependabot automation and custom image?

deepmap-marcinr:
I am happy to merge the Dockerfile and Makefile changes to produce a container image, however, I'm in a tight spot with where to host this docker image.

Deepmap got acquired by a large company which has all kinds of licensing requirements and internal requirements around shipping binaries, because it opens us up to liability in case we accidentally ship a virus or malware or something.

So, could I propose that we commit the build into this repository, but not the action hook which publishes it? Then, maybe someone who doesn't have these constraints can publish an image. I can link to it in the readme, but mention that it's not supported.

jamietanna:
Hey @deepmap-marcinr is it right that we're maybe in a state we may need to close this as a "can't do, legally"?

Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG GO_VERSION=1.18
+ARG ALPINE_VERSION=3.16
```
jamietanna, line 2: These will need a change

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+COPY . /go/src/github.com/deepmap/oapi-codegen
+WORKDIR /go/src/github.com/deepmap/oapi-codegen
+RUN go build -o ./bin/oapi-codegen ./cmd/oapi-codegen/oapi-codegen.go
+
+### User/passwd layer
+FROM alpine:${ALPINE_VERSION} as user-passwd
+RUN addgroup -S oapi-codegen && adduser -S oapi-codegen -G oapi-codegen
+
+### Image
+FROM scratch as image
```
jamietanna, line 18: Alternatively we could use Chainguard's lightweight undistro images

jamietanna (changes requested):
Just awaiting Marcin's thoughts, after which we can look at some follow up changes

Dockerfile
insidieux: Fixed, set go version according to go.mod and up alpine version

Dockerfile
insidieux: To be honest, i've never used Chainguard's  undistro images... 
I made a change: added possibility to change base image when we want on build.

jamietanna:
Hey folks, I've opened this page a couple of times recently to draft a reply.

The short answer is, nothing yet - there's some behind the scenes stuff which hopefully should make this progress, but until then, Marcin's comment above still holds about us not being able to serve the image.

Am eagerly hoping to have an answer for this before long!

jamietanna:
One thing I also want us to make sure we think about with this is how we'll respond to CVEs in the underlying container (i.e. I don't want to have to have folks chasing us for "hey there's a completely unrelated CVE in the container, please fix")

DrBlury:
A FROM SCRATCH image could be used with just the built go binary. Then there are no such CVEs, correct?

jamietanna:
Correct yes, so chances are this proposal would avoid the CVE concerns, as we'd distribute `FROM scratch` 👍🏼

insidieux:
> Correct yes, so chances are this proposal would avoid the CVE concerns, as we'd distribute `FROM scratch` 👍🏼

Already done in my PR )

bweston92:
Any progress on this?

Ideally a stage in my CI (publishing generated Go files as a package) will run `docker run <oapi-codegen> ...` to generate sources rather than having to use the Go toolchain.

jamietanna:
Hey folks, apologies for the delay on the final decision here - we've decided not to proceed with a Docker image for `oapi-codegen` for the following reasons:

1. [The official recommendation for `oapi-codegen`'s installation is as a source-tracked / `go tool`](https://github.com/oapi-codegen/oapi-codegen/?tab=readme-ov-file#install), not as a binary
1. We'd prefer to not have to maintain a container (even `FROM scratch`)
1. We'd prefer consumers use source-tracked dependencies, which make it easier to [pin to commits](https://github.com/oapi-codegen/oapi-codegen/?tab=readme-ov-file#pinning-to-commits)

We welcome a community-maintained container image - but at this point, this isn't something we're going to be maintained upstream

At some point, we may reconsider this, at which point we'll distribute both a container image and a pre-built binary.

Thanks for your patience + understanding!

PR #910, opened 2022-12-30
support for generating protobuf spec

chaseisabelle:
this is a work in progress, and is a bit of a mess right now. attempting to add support for generating protobuf spec via openapi spec, with http gateway. any feedback/suggestions/contributions is greatly appreciated. 

@todo:
- header params
- schema type defs
- auth stuff


deepmap-marcinr:
Our internal tooling is heavily proto based, and I've thought long and hard about supporting protos nicely in oapi-codegen, and I've not come up with a good answer yet. I think you're going to run into a usability wall.

We treat them as binary blobs of `application/x-google-protobuf`. OpenAPI defines the paths and operations, while protos define the models. The value of oapi-codegen, in my opinion, is turning OpenAPI specs into Go models that are nice to use, but we can't really do that in protos because they are opaque to us.

I've gone so far as to write a `protoc` plugin which can generate OpenAPI models from `.proto` files, so that we can at least be aware of JSON structure of protos but this didn't have any valid use cases.

So, what is your intent with this?

chaseisabelle:
i think i submitted this when i was a bit lit. i found a better way to do what i wanted and forgot to close this out. i may have a use case for transpiling openapi specs to protobuf specs, but that's still being worked out.

tl/dr: thanks for taking a look, but i think i'm gonna scrap this one for now.

PR #914, opened 2023-01-04
fix(gin-template): Respect ctx.Abort()

alexstojda:
Respect when c.Abort() is called in a middleware func.


re-cheid:
I think this PR fixed issue #485

PR #917, opened 2023-01-05
Update authenticated example - set token in echo context..

sblackstone:
It was completely unclear to me how to propagate the parsed token back to the echo context so that it can be used in the handlers.

This small PR updates the example so that the property is set...

### DIFF
--- a/examples/authenticated-api/echo/server/jwt_authenticator.go
+++ b/examples/authenticated-api/echo/server/jwt_authenticator.go
@@ -18,6 +18,8 @@ type JWSValidator interface {
 	ValidateJWS(jws string) (jwt.Token, error)
 }
 
+const JWTClaimsContextKey = "jwt_claims"
+
 var (
 	ErrNoAuthHeader      = errors.New("Authorization header is missing")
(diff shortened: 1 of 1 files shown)


examples/authenticated-api/echo/server/jwt_authenticator.go
```diff
@@ -73,6 +74,11 @@ func Authenticate(v JWSValidator, ctx context.Context, input *openapi3filter.Aut
 	if err != nil {
 		return fmt.Errorf("token claims don't match: %w", err)
 	}
+
+	// Set the user property on the echo context so the handler is able to use the token we generate in here.
+	eCtx := middleware.GetEchoContext(ctx)
+	eCtx.Set("user", token)
```
deepmap-marcinr, line 80: This works, however, let's make it nicer code.

Make a global constant, eg, "JWTClaimsContextKey" and named it "jwt_claims" or sometihng. "user" is awfully generic. We do this kind of stuff all over our produciton code, and this recommendation is based on pain experienced.

sblackstone: Done

sblackstone:
Can this be merged?

PR #918, opened 2023-01-06
Add all enum values helper

skirsten:
Hi, what do you think of something like this?

```go
// Defines values for Weather.
const (
	WeatherSunny   Weather = "sunny"
	WeatherRaining Weather = "raining"
)

var AllWeathers = []Weather{
	WeatherSunny,
	WeatherRaining,
}

```

This would be very useful when integrating with other frameworks or libraries (e.g. writing a terraform provider).
It could also be hidden behind a flag to not bloat by default.

Let me know what you think.

### DIFF
--- a/pkg/codegen/templates/constants.tmpl
+++ b/pkg/codegen/templates/constants.tmpl
@@ -12,4 +12,9 @@ const (
   {{$name}} {{$Enum.TypeName}} = {{$Enum.ValueWrapper}}{{$value}}{{$Enum.ValueWrapper -}}
 {{end}}
 )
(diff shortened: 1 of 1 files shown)


deepmap-marcinr:
Sorry to take so long to get this. Yes, I think this would be useful, for sure. The difficulty of changes like this is that they can cause type conflicts. What if there is another type already named `AllWeathers`? I wonder if we can put some functions on the Weather type to avoid this - but then this is ugly because you need an instance of Weather to query all weathers.

skirsten:
I would still really like this but don't have the time to implement any additional logic to check for duplicates etc.
We just extended the constants.tmpl via user-templates for now.

PR #929, opened 2023-01-20
feat: filter by operation ids

johanneswuerbach:
When generating server mocks or clients we often only need specific operations to be included. Right now we workaround this issue by having unique operation tags, but this has various side-effects.

Add the capability to include or exclude specific operations by their operation id inline with the existing support to filter by tags.


johanneswuerbach:
I would also appreciate a look here @deepmap-marcinr and sorry about the spam :-(

johanneswuerbach:
@deepmap-marcinr is this something, which would be acceptable for this repository? Happy to further elaborate on our use-case.

jamietanna:
Hey @johanneswuerbach thanks for raising! Would you mind adding an example of the usage under `internal/test`, just to show what it would look like + ensure we've got a test to catch regressions in the future

Once done, I'd say we can merge :grin:

johanneswuerbach:
Thanks for the review @jamietanna , integration tests added 👍

johanneswuerbach:
@jamietanna anything else I can support here? Thank you 🙇

johanneswuerbach:
@jamietanna I know you are busy, but I wondered if there is any chance to land this :-)

jamietanna (approved):
Thank you for this!

Issue #712, opened 2022-08-09
The decriminator is set only if the "mapping" field is present


Antonboom:
Related to https://github.com/deepmap/oapi-codegen/pull/471

Hello!

`mapping` field is useful for explicit mapping:
> In scenarios where the value of the discriminator field does not match the schema name or implicit mapping is not possible, an optional mapping definition MAY be used

But by default I expected the generated code to set the descriminator field by schema name:
> The expectation now is that a property with name petType MUST be present in the response payload, **and the value will correspond to the name of a schema** defined in the OAS document.

```yml
MyResponseType:
  oneOf:
  - $ref: '#/components/schemas/Cat'
  - $ref: '#/components/schemas/Dog'
  - $ref: '#/components/schemas/Lizard'
  discriminator:
    propertyName: petType
```

But now it doesn't work like that. Thanks.

Antonboom:
@jamietanna @deepmap-marcinr 
FYI

PR #932, opened 2023-01-22
OneOf: Implicit mapping

Antonboom:
Closes https://github.com/deepmap/oapi-codegen/issues/712

---

From https://swagger.io/specification/#discriminator-object.

In scenarios where the value of the discriminator field does not match the schema name or implicit mapping is not possible, an optional mapping definition MAY be used:

```yaml
MyResponseType:
  oneOf:
  - $ref: '#/components/schemas/Cat'
  - $ref: '#/components/schemas/Dog'
  - $ref: '#/components/schemas/Lizard'
  - $ref: 'https://gigantic-server.com/schemas/Monster/schema.json'
  discriminator:
    propertyName: petType
    mapping:
      dog: '#/components/schemas/Dog'
      monster: 'https://gigantic-server.com/schemas/Monster/schema.json'
```

Here the discriminator value of dog will map to the schema #/components/schemas/Dog, rather than the default (implicit) value of Dog. If the discriminator value does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail. Mapping keys MUST be string values, but tooling MAY convert response values to strings for comparison.

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -1,6 +1,7 @@
 package codegen
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


MysterySuperhero:
+1

53jk1 (approved):

pkg/codegen/schema.go
```diff
@@ -759,15 +760,26 @@ func generateUnion(outSchema *Schema, elements openapi3.SchemaRefs, discriminato
...
 			for k, v := range discriminator.Mapping {
 				if v == element.Ref {
 					outSchema.Discriminator.Mapping[k] = elementSchema.GoType
+					mapped = true
 					break
 				}
 			}
+			// Implicit mapping.
+			if !mapped {
+				outSchema.Discriminator.Mapping[elementSchema.GoType] = elementSchema.GoType
```
Warboss-rus, line 774: I don't think elementSchema.GoType should be used as a key here. It should be schema name (GoType is camel-cased and have a prefix)

Warboss-rus (changes requested):

pkg/codegen/schema.go
Antonboom: @Warboss-rus thank u!
It is my typo. Of course the schema name should be here. Added tests and fixed 👍

Warboss-rus (approved):
LGTM

Antonboom:
@deepmap-marcinr @jamietanna

Antonboom:
Could it be merged, guys?

@Warboss-rus 
@deepmap-marcinr 
@dannymidnight 
@jamietanna 
@chrusty 
@pseudo-su 
@johnstableford 
@weitzj

deepmap-marcinr:
This looks like a very good change, and doesn't appear to break any generated, but changes to oneOf/anyOf are terrifying to me since they are so riddled with special cases.

Antonboom:
@deepmap-marcinr when do you plan to make the new tag? :)

PR #937, opened 2023-01-24
Release Github Action

franklinkim:
Adds automated release creation (including binaries) on tag push using github action and  goreleaser.

**NOTE:**

You would need to add your `PERSONAL_GITHUB_TOKEN` as a new Repository Secret (Settings > Secrets and Variables > Actions).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,32 @@
+name: Release
+on:
+  push:
+    tags:
+      - v*.*.*
+env:
+  GOFLAGS: -mod=readonly
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v3
+        with:
+          # Required for the changelog to work correctly.
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v4
+        with:
+          distribution: goreleaser
+          version: latest
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+          go-version-file: 'go.mod'
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v4
+        with:
+          distribution: goreleaser
+          version: latest
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.PERSONAL_GITHUB_TOKEN }}
```
53jk1, line 32: r u sure it will work?

53jk1 (changes requested):

.github/workflows/release.yml
franklinkim: if you define add a/your github token as a secret to your repo, yes. 

you can see it in action on my fork: https://github.com/franklinkim/oapi-codegen/releases

insidieux: @53jk1 @franklinkim 
You can use predefined variable, passed by Github Actions, so you don't need to create new variable by yourself.
Check example [here](https://github.com/deepmap/oapi-codegen/pull/888/files#diff-82014328c5245884891cc33fbab1ee6d5fc9d2c192b8712b0a222297720172a3R28) in related issue/PR.

franklinkim: Give it a try but I had to add a different one since `secrets.GITHUB_TOKEN` seemed to be missing permissions to create the release

jamietanna:
Thanks for this - at this point, we're working out whether we _do_ want to pre-build binaries, as [our official recommendation](https://github.com/oapi-codegen/oapi-codegen/?tab=readme-ov-file#install) is to use the source distribution with `go run`.

We'll try and have a look at it again at some point, but unfortunately this is low on our priority list.

Also, for full disclosure, something we're _considering but have not at all decided on_ is whether this may be a paid feature, to help [make the project more sustainable](https://github.com/oapi-codegen/oapi-codegen/discussions/1606) if companies require a pre-built binary distribution

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the project uses a different release workflow (`release-drafter.yml`), and the official recommendation is `go run` rather than pre-built binaries.

PR #941, opened 2023-01-26
Fix allOf merges using external refs

### LINKED ISSUE #940: Using allOf with $ref to json schema errors with "unsupported reference"
The generator fails to generate a spec when the schemas use a `$ref` that points to an external JSON schema. This was previously working in v1.11.0. Now broken in v1.12.x.

Sample spec file

```yml
openapi: 3.0.0
x-stoplight:
  id: hvun3hw84a2v8
info:
  title: ref-bug
  version: '1.0'
  description: yay
  contact:
    name: Example
servers:
  - url: 'http://localhost:3000'
paths:
  '/users/{userId}':
    parameters:
      - schema:
          type: integer
        name: userId
        in: path
        required: true
        description: Id of an existing user.
    get:
      summary: Get User Info by User ID
      tags: []
      responses:
        '200':
          description: User Found
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/User'
      operationId: get-users-userId
      description: Retrieve the information of the user with the matching user ID.
components:
  schemas:
    User:
      allOf:
        - $ref: 'https://edgio.github.io/hyperion/schema/node.json'
        - type: object
          additionalProperties: false
          properties:
            name:
              type: string
              description: User name
          required:
            - id
      title: user
      description: Represents a `user`
```

Results in Error message:
> `error generating code: error generating type definitions: error generating Go types for component schemas: error converting Schema User to Go type: error merging schemas: unsupported reference: https://edgio.github.io/hyperion/schema/node.json`

octomad:
Fixes #940.

I've added to the test schema to illustrate the situation and to ensure that the code coverage is there for this particular scenario.

### DIFF
--- a/pkg/codegen/merge_schemas.go
+++ b/pkg/codegen/merge_schemas.go
@@ -53,10 +53,10 @@ func valueWithPropagatedRef(ref *openapi3.SchemaRef) (openapi3.Schema, error) {
 	}
 
 	pathParts := strings.Split(ref.Ref, "#")
-	if len(pathParts) != 2 {
+	if len(pathParts) < 1 || len(pathParts) > 2 {
 		return openapi3.Schema{}, fmt.Errorf("unsupported reference: %s", ref.Ref)
 	}
-	remoteComponent, _ := pathParts[0], pathParts[1]
+	remoteComponent := pathParts[0]
 
 	// remote ref
 	schema := *ref.Value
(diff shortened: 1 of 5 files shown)


pkg/codegen/merge_schemas.go
```diff
@@ -53,10 +53,10 @@ func valueWithPropagatedRef(ref *openapi3.SchemaRef) (openapi3.Schema, error) {
 	}
 
 	pathParts := strings.Split(ref.Ref, "#")
-	if len(pathParts) != 2 {
```
octomad, line 56: This appears to be an artificial limitation for `pathParts` to have exactly 2 parts.

PR #944, opened 2023-01-26
Feat: Add support for custom templates

marcos10soares:
The goal of this PR is to add support for custom user templates without using any specific framework. 

The user specifies a template folder in the config yaml and enables the custom-templates flag. 

The generated code should only use the provided custom user templates plus the types generation.


marcos10soares:
hey @jamietanna @deepmap-marcinr ! what do I need to do to get this PR approved? from the reactions I can see that at least some people are interested in it, also it doesn't impact the project in any way, it's just adding a basic a feature that shouldn't require more development... would love to hear from you! Thanks! 😃

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the `output-options.user-templates` configuration already supports custom template folders, allowing template overrides from files, URLs, or inline. This PR's goal is already met.

PR #947, opened 2023-01-28
Add per-handler middleware support for Gin

festinuz:
Here's my attempt at providing a way to define per-handler middlewares with codegen.

This is in-spirit continuation of abandoned PR #462 and related issue #461, and i cherry-picked code from previous PR.

Key differences:
1. Registered handler-specific middlewares are added to `ServerInterface` and thus are enforced (instead of silent failure)
2. It's for Gin instead of Chi


festinuz:
@deepmap-marcinr, i'd love to hear your thoughts on this (or from any other maintainer)

Looks OK? What's missing? It's kind of hard to know myself as there is no "contribute guidelines" or something similar

One thing that bothers me personally is that this new feature wont be available on all supported servers (Chi, Echo, etc), and i don't feel qualified and/or motivated to add it to all of them. A possible solution for this is to add some kind of "supported features table" in README.md and mark this thing as supported in Gin

README.md
```diff
@@ -662,6 +675,53 @@ which help you to use the various OpenAPI 3 Authentication mechanism.
     type ObjectCategory int
     ```
 
+- `x-oapi-codegen-middlewares`: specifies a list of middlewares. These can be middlewares that are
```
festinuz, line 678: What is conventiof for naming these? Maybe a simpler name, such as `x-middlewares`, is enough?

pkg/codegen/operations.go
```diff
@@ -222,6 +222,24 @@ type OperationDefinition struct {
 	Method              string                  // GET, POST, DELETE, etc.
 	Path                string                  // The Swagger path for the operation, like /resource/{id}
 	Spec                *openapi3.Operation
+	Middlewares         []string // Sent as part of x-oapi-codegen-middlewares
+}
+
+type Data []OperationDefinition
```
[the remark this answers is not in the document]
festinuz: So i went ahead and changed it to `OperationDefinitionsData`, which makes sense to me

README.md
```diff
@@ -546,17 +546,30 @@ which help you to use the various OpenAPI 3 Authentication mechanism.
 
 ## Extensions
 
+| **Extensions**               |    **net/http**    |      **Echo**      |       **Chi**      |       **Gin**      |
```
festinuz, line 549: I've decided to include this table to indicate which extensions are supported for each server.

My reasoning is this: It's easier to contribute to one server implementation than several (as no one is using all of them at once and has a know-how required), and a table like this allows to it explicitly

Also, missing server extensions are a "good first issue", because you can use original feature PR as a reference on what needs to be done.

obrhoff:
Since this PR is also closed. Did you find another way to archive this?

festinuz:
> Since this PR is also closed. Did you find another way to archive this?

Nope, since this PR was ignored i just forked the project and re-wrote a bunch of stuff i needed for my project, such as per-handler middlewares, default values for models,  proper error responses in case of unhandled errors (this package returned empty HTTP 200 at the time, dont know if this has changed)

Sadly it's far from production-ready for external use, and i don't have plans to work on it, so it's not a public repo

sakaijunsoccer:
I also want this feature, but is there a better way to do it?

watercraft:
What I'm looking for is a way to op-in per path on the middleware.  I could do this by having the middleware inspect the swagger for the op-in annotation, perhaps by reviewing how validate does it's job.  I'm coming from the ancient go-raml project that had a securedBy tag to inject something like middleware for authentication.  Note, OpenAPI 3 has appropriate annotations they are just not used by oapi-codegen.

watercraft:
Found my solution using github.com/oapi-codegen/nethttp-middleware OapiRequestValidatorWithOptions AuthenticationFunc.  The function is not called when you configure and empty array on the security tag for your method.  This way I can document which methods require authentication, implement it with a custom function, and know that at least that aspect of the documentation matches the implementation.

PR #952, opened 2023-02-06
Fix pointers equality in TypeDefinitionsEquivalent function

spinillos:
`OAPISchema` is a pointer and using `==` to check the equality of the pointers is always false. The correct way is doing it using `DeepEqual`.


spinillos:
@deepmap-marcinr Sorry for the delay updating this PR. Looks like that it doesn't have more conflicts.

Issue #943, opened 2023-01-26
Generated client code cannot convert UUIDs to string when used as path parameters.


optimality:
I have an API that defines a path parameter as such:

```yaml
id:
      type: string
      format: uuid
```

The generated server code takes a `uuid.UUID` as a parameter, correctly.  The generated client code also accepts a `uuid.UUID` as a parameter as well, again correctly.

But when I actually call the generated client function and pass a UUID in, I get `unsupported type uuid.UUID`.  I believe this is because `StyleParamWithLocation` has no casing for `uuid.UUID`, it just forwards it to `stylePrimitive`, which forwards to `primitiveToString`, which also has no case for `uuid.UUID`.

The offending code is:
```go
switch kind {
	case reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Int:
		output = strconv.FormatInt(v.Int(), 10)
	case reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uint:
		output = strconv.FormatUint(v.Uint(), 10)
	case reflect.Float64:
		output = strconv.FormatFloat(v.Float(), 'f', -1, 64)
	case reflect.Float32:
		output = strconv.FormatFloat(v.Float(), 'f', -1, 32)
	case reflect.Bool:
		if v.Bool() {
			output = "true"
		} else {
			output = "false"
		}
	case reflect.String:
		output = v.String()
	default:
		return "", fmt.Errorf("unsupported type %s", reflect.TypeOf(value).String())
	}
```

It would be trivial to add
```go
case uuid.UUID:
    output = v.String()
```

but I'm not sure you want to introduce a dependency on `uuid.UUID` at this point.

optimality:
Thank you!

deepmap-marcinr:
Thank you for fixing my crappy code :)

PR #956, opened 2023-02-09
fix generated client code cannot convert UUIDs to string when used as…

benyaminbeyzaie:
… path parameters.

fixes https://github.com/deepmap/oapi-codegen/issues/943


jxsl13:
it might make sense to assert the encoding.TextMarshaller after the JSONMarshaller interface instead of the concrete uuid.UUID type.

johnshiver:
should likely update docs and tests

PR #968, opened 2023-02-18
Add ability to try to get templates from files or remotes

### LINKED ISSUE #905: Load user-templates from URL
I have a number of smaller services that I generate API for (ideally) using the same custom user templates, all in different repositories. To prevent drift between my customized templates in different services, I would like to be able to set the template with a URL as opposed to the raw template file.

Example:
```yaml
output: api.gen.go
package: api
output-options:
  skip-prune: true
  user-templates:
    client-with-responses.tmpl: https://raw.githubusercontent.com/deefdragon/oapi-templates/main/templates/client-with-responses.tmpl
```

deefdragon:
Resolves #905

This is currently a draft as I have yet to fully test it.

I think it may be worth doing some minor changes to how templates are loaded in in this PR as well. Id like to allow users to load in more templates than the ones created in this library. I have wanted to create a few helper templates, but its been hard to manage them as I only have a limited number of files and it gets a bit unwieldy.


deefdragon:
I modified the code for loading in defined templates to load in all files the user specifies. I also added documentation as to the new loading, as well as some tests to make sure everything works as expected.

deefdragon:
@deepmap-marcinr is there anything you would need done or changed to get this merged in?

deepmap-marcinr:
I generally do not like fetching specs or templates dynamically because it creates a moving target for an API - each time you regenerate your code, it could change because your upstream specs could have changed. IMO, it's much better to download and commit the specs you are using locally, for predictability. I guarantee you, that if people use this, the next requests coming will be to find some way to track versions or changes.

This change is isolated, so it'd not be an issue to merge it, but I'm not sure if it's necessary.

deefdragon:
If someone was working on a more monolithic project or in a mono-repo, I think worrying about that is fair. The issue however, and why I brought it up originally, is micro-services.

I work on a micro-services architecture with each service in its own git repository. I have about a dozen services, and currently, every time I need to upgrade their templates to add tracking or logging etc., I have to go in and copy-past and then commit for each service. This is a non-trivial amount of work that is only going to grow as I add services. 

I think you are right that its probably not the _best_ practice, but software design is all about balancing trade-offs like speed to write, ability to maintain, ability to scale, security, feature set etc.. To me, being able to guarantee that all my services are working off the same, most up to date templates, (and thus tooling configurations), is more important than the potential API drift (under some circumstances, the API drift is the goal to get said security updates). But I also control all my templates directly.

I've updated the readme with examples of how to use raw.githubusercontent.com URLs etc., and I included a warning encouraging users to use tags or hashes instead of branches to prevent API drift.

deefdragon:
@deepmap-marcinr merged in master to make sure I was up to date and fixed the linting issues. Sorry about the delay on that.

PR #982, opened 2023-02-27
Fix README.md example usage

hjblom:
Currently the example in README.md is not working:

```text
➜  oapi oapi-codegen petstore-expanded.yaml > petstore.gen.go
configuration error: package name must be specified
```

This can be fixed by providing the required tag for -package


stevenh:
This actually isn't needed, the cause was the lack of release after the fix for this went in.

PR #983, opened 2023-02-28
Update schema.go

aclwitt:
Add nested property support for properties; otherwise TypeNames will collide

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -349,7 +349,21 @@ func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
 
 			// We've got an object with some properties.
 			for _, pName := range SortedSchemaKeys(schema.Properties) {
+				description := ""
 				p := schema.Properties[pName]
+				if p.Value != nil {
+					description = p.Value.Description
+					// Get x-go-name for property, if it exists
+					if extGoFieldName, ok := p.Value.Extensions[extGoName]; ok {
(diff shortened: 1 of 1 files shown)


aclwitt:
@deepmap-marcinr I think you touched the property-x-go-name code most recently. I didn't dig too deep, so feel free to lmk if this is all wrong :)

mromaszewicz:
I rebased your change to the latest `main`, and noticed that your change affects code generation in the large components test. Looking into it now.

mromaszewicz:
I tried to get this working correctly, and this change isn't compatible with the way that properties are traversed and stored.

The CI failure above is because, by poking into a schema to rename the type, the name based matching for `required: [...]` breaks.

However, the more fundamental problem is that `x-go-name` is typically defined as an extended property on the object that a field refers to. You've got a string with enum values, and in this particular case, you want that `x-go-name` to refer to the enum's name, not the name of the string property itself. It creates an undecidable problem - what does the `x-go-name` affect, the string field, or the enum within? We'd need a separate extension just to rename the enum value within a schema.

I'm closing this because it breaks lots of other things. Your example works because it's simple, and only a single test case in the giant components.yaml breaks, but this will break other schemas. I'm glad a test caught it.

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because `x-go-name` extension is already supported on nested enum properties, and the enum handling code processes enums for all properties including nested ones.

PR #992, opened 2023-03-07
fix959

Rmarken5:
- issue958
- String manipulation
- Adjust logic to provide external refs


deepmap-marcinr:
I had to revert this because the codegen was crashing.

Rmarken5:
> I had to revert this because the codegen was crashing.

Would you mind providing the json that codegen is running against that's causing the crash so I can take another look?

deepmap-marcinr:
Oh, it was the issue 959 unit test, when you run (go generate ./...) at the top of the module, or just in that issue directory. It's crashing in resolving external references because the SchemaRef for something is nil. I've not had a chance to debug further.

It's easy to repro, just revery my revert and try to regenerate. It's probably some conflict with an earlier commit from someone.

-- Marcin

Issue #993, opened 2023-03-08
Strict interface with reserved Go keywords as property yield invalid generated code


reinkrul:
When generating a strict (Echo) interface for an API that contains properties that are reserved Go keywords, the generated code is invalid.

YAML spec:

```yaml
openapi: 3.0.0

info:
  title: api
  description: sample api
  version: "0.0.0"

paths:
  /endpoint/{type}:
    get:
      parameters:
        - in: path
          name: type
          required: true
          schema:
            type: string
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: string
```

YAML config:

```yaml
package: pr
generate:
  echo-server: true
  models: true
  strict-server: true
```

Generated code:
```go
type GetEndpointTypeRequestObject struct {
	Type string `json:"type"`
}

func (sh *strictHandler) GetEndpointType(ctx echo.Context, pType string) error {
	var request GetEndpointTypeRequestObject

	request.PType = pType // compilation error, should've been request.Type = pType
	// etc
}
```

(see https://github.com/reinkrul/oapi-codegen/tree/reserved-keywords-bug/pr)

Probably related to https://github.com/deepmap/oapi-codegen/issues/30

PR #999, opened 2023-03-10
Fix incorrect fieldname reserved keywords in path parameters (strict server)

reinkrul:
Fixes https://github.com/deepmap/oapi-codegen/issues/993

When generating a strict server interface for an operation that contains a path parameter (e.g. `type`), which is named after a Go keyword, the generated code is invalid.

It generated:
```go
request.PType = pType
```

Which should've been:
```go
request.Type = pType
```

This PR fixes that.


deepmap-marcinr:
Thank you.

PR #1005, opened 2023-03-15
Fix generate equals  structs

l3xx:


deepmap-marcinr:
This is going to hide problems. Since we rename identifiers to become valid Go types, we occasionally create a conflict, and this code will hide duplicates versus creating an error; meaning that some types are missing. That's no good as well.

I think I prefer the generated code breaking due to conflict because it indicates a problem.

jamietanna:
@deepmap-marcinr would you recommend closing this PR, or is there something else we can do with it?

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because this approach was rejected in review — it would hide real problems rather than surface them as errors.

Issue #958, opened 2023-02-10
Client With Responses cant reference external responses


amammay:
In the client-with-responses template if you have have 

pkg b
```yaml
components:
  responses:
    200:
      description: "OK"
      content:
        'application/json':
          schema:
            $ref: '#/components/schemas/Document'
  schemas:
    Document:
      type: object
      properties:
        fields:
          type: object
          additionalProperties:
            $ref: '#/components/schemas/Value'
    Value:
      type: object
      properties:
        stringValue:
          type: string
        arrayValue:
          $ref: '#/components/schemas/ArrayValue'
    ArrayValue:
      type: array
      items:
        $ref: '#/components/schemas/Value'
```

and in pkg a
```yaml
paths:
  /example:
    get:
      operationId: exampleGet
      responses:
        200:
          $ref: "./pkgb/spec.yaml#/components/responses/200"
```       

the codegen output doesnt include external refs

<img width="894" alt="image" src="https://user-images.githubusercontent.com/17421192/218197241-6efc12d7-4037-42a7-ac5a-9eda40423c61.png">

I will open a PR with a broken test to showcase an exact replica

amammay:
pr that shows issue -> https://github.com/deepmap/oapi-codegen/pull/959

Rmarken5:
Proposed solution: #992

Rmarken5:
Issue is with the bump to `kin-openapi v0.112.0 => V0.114.0`.

Schema struct is `loaded` with `Component` as `nil` in new version.
In root schema yaml, I added a component property and schema is loaded properly into memory.

Generated models now have correct package.

[fixed with](https://github.com/deepmap/oapi-codegen/pull/1013  )

jamietanna:
Hey folks, mind re-reviewing this in light of the v2.1.0 release, which made some significant changes (:crossed_fingers: and the fix for this) to external references?

PR #1013, opened 2023-03-27
issue 958 prop

Rmarken5:
- revert revert
- Fix spec

fixes [ISSUE 958](https://github.com/deepmap/oapi-codegen/issues/958)


jamietanna:
Thanks for raising! We may not be able to get this into the release that's coming over the next few days but will try and get to it as a patch release not long after!

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the code has been significantly restructured since this PR was written, and it's no longer possible to determine its applicability.

PR #1033, opened 2023-04-25
Improve logic for x-omitempty

veleek:
Currently x-omitempty only enables you to force the _absence_ of `omitempty`.  This PR changes the logic so that the existance of `x-omitempty` will force omit empty to be whatever the user specifies.

The logic around omit empty is pretty confusing here though.  For example, why is omitEmpty excluded by default if it's nullable (e.g. if `p.Nullable` then `omitEmpty = false`).

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -638,17 +638,20 @@ func GenFieldsFromProperties(props []Property) []string {
 
 		field += fmt.Sprintf("    %s %s", goFieldName, p.GoTypeDef())
 
+		omitEmpty := !p.Nullable &&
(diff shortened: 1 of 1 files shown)


ndbeals:
I agree with the better logic to allow explicitly setting `x-omitempty: true` to force an `omitempty` tag, regardless of whether `nullable: true` is set.

> why is omitEmpty excluded by default if it's nullable
Because if the `omitempty` tag _is_ set when nullable is true then marshaling the struct to json where a nullable field is a pointer and that pointer is `nil` will cause the marshaled JSON to drop that field, where it should be `"field": null`.

deepmap-marcinr:
Thank you.

PR #1036, opened 2023-04-26
feat: add force decode as string

vyegres:
The float and decimal data types in programming languages can have different precision and formatting, which can lead to loss of accuracy or errors when encoding and decoding in the JSON format. Therefore, it is recommended to encode float and decimal values as strings in a format that ensures the preservation of the value's accuracy and format.

For example, in the JSON format, a floating-point number can be encoded as a string in the format "3.14159265359" instead of the number 3.14159265359 to avoid loss of accuracy. Similarly, a decimal value can be encoded as a string in the format "0.12345678901234567890" to preserve precision and formatting.


mromaszewicz:
Thank you for submitting this PR, @vyegres — but we need to close it because it has been superseded by #2223, which delivered the more general `output-options.type-mapping` feature. That covers float/decimal-as-string encoding (and any other primitive remapping) via config rather than a dedicated extension. There's a worked example at `examples/output-options/type-mapping/`.

Apologies for the long wait, and thanks for the contribution.

PR #1058, opened 2023-05-23
Small Features: YAML struct tags, Exclude Depreciated Filter

dave-gray101:
I have a few small tweaks I'm testing out locally - created a PR to track them in case they are useful to others.

- Adds YAML struct tags identical to the JSON ones already generated. 
- Adds a new filter `exclude-depreciated` that does what it says on the tin and excludes depreciated operations from the generated code.

Please let me know if there's a better way to accomplish these, or if there's any improvements recommended!

(In particular, I'm not sure if I'm expected to have ran `make generate` before pushing up this PR - if not, the `*.gen.go` files can be removed from this PR easily)

### DIFF
--- a/examples/authenticated-api/echo/api/api.gen.go
+++ b/examples/authenticated-api/echo/api/api.gen.go
@@ -27,21 +27,21 @@ const (
 // Error defines model for Error.
 type Error struct {
 	// Code Error code
-	Code int32 `json:"code"`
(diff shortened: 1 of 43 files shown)


dave-gray101:
Depreciating this PR - I've cleaned up my struct tag support and am working on a few more tweaks.

PR #1067, opened 2023-05-29
return 404 when no matching route is found

### LINKED ISSUE #1066: [middleware] Request validator should return 404 not 400 if route is unable to be found
This line should return `return http.StatusNotFound, err // We failed to find a matching route for the request.`

https://github.com/deepmap/oapi-codegen/blob/14548c7e7bbee8d8fbd7e706f9b6f9e3381b44fb/pkg/chi-middleware/oapi_validate.go#L75-L81

Thoughts?

sonu27:
Closes #1066

### DIFF
--- a/pkg/chi-middleware/oapi_validate.go
+++ b/pkg/chi-middleware/oapi_validate.go
@@ -77,7 +77,7 @@ func validateRequest(r *http.Request, router routers.Router, options *Options) (
 	// Find route
 	route, pathParams, err := router.FindRoute(r)
 	if err != nil {
-		return http.StatusBadRequest, err // We failed to find a matching route for the request.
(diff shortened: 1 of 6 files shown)


jxsl13:
Should be 405. 404 is resource not found contrary to method not allowed/ not registered.

sonu27:
Currently, it's returning 400 which is not correct.
If you hit most servers on the web where a route does not exist e.g. https://google.com/asdfadsfasdfasdf you will get 404.
405 is if the route exists but method is different.

jxsl13:
fair point

sonu27:
Just updated other middlewares too

sonu27:
@deepmap-marcinr What do you think?

PR #1071, opened 2023-05-31
add links to source in the README.md file

BigBoulard:


jamietanna (commented):
I'll review this before the next release (this week) - looks like a number of changes introduded by a Markdown autoformatter, which generally I'd not be super pro, but looks like these are all updates I'd want to make if I were writing the document, so thanks 👏 

Which Markdown formatter did you use? Or were these all by hand?

PR #1083, opened 2023-06-04
Shorter and more understandable interface function names

adrianpk:


adrianpk:
Cancelled pull request.

PR #1090, opened 2023-06-06
Chi: generate Unimplemented implementation

lelvisl:
Update chi-interface.tmpl to generate an implementation for ServerInterface that always returns a `http.StatusNotImplemented`.

The Unimplemented server can be used like this (assuming the code is generated in package spec):

```
type MyAPIServer struct {
  spec.Unimplemented
}
```

Using Unimplemented is optional, but might help when prototyping APIs or in case the OpenAPI spec is managed in a separate repo and mutliple developers.

grpc-go uses a similar strategy. For example see the [examples/route_guide/server/server.go](https://github.com/grpc/grpc-go/blob/master/examples/route_guide/server/server.go#L57) file.

ref: https://github.com/deepmap/oapi-codegen/pull/631


lelvisl:
@deepmap-marcinr please look at this pr =)

PR #1092, opened 2023-06-07
Fix 1091

### LINKED ISSUE #1091: type name is conflicting, if x-go-type-name's value and generated type name is same
- related to #25 

something like followings

```console
oapi-codegen --config tools/oapi-conf/type.yaml common_types.yaml
```
yaml

```yaml
openapi: 3.0.3
info:
  title: api
  version: 0.0.0
paths:
  /_value:
    get:
      operationId: value
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/value'
          description: OK
components:
  schemas:
    value:
      type: object
      properties:
        name:
          type: string
      x-go-type-name: Value
```

conf

```yaml
package: oapitype
generate:
  models: true
compatibility:
  always-prefix-enum-values: true
```

version

```
v1.13.0
```

## more context

We will sharing type definitions between X and Y services, so we will fixing the name of generated type explicitly.

- x/server.go from X/main.yaml
- y/server.go from Y/main.yaml
- oapitype/types.go from common-types.yaml (from X/main.yaml and Y/main.yaml)

## use `x-go-name` ?

It is incorrect.

```yaml
components:
  schemas:
    error:
      description: error response body
      type: object
      properties:
        createdAt:
          $ref: "#/components/schemas/datetime"
        updatedAt:
          $ref: "#/components/schemas/datetime"
    datetime:
      type: string
      format: date-time
      x-go-name: MyDateTime
```

is generating like followings, but we don't need MyDateTime field.

```go
// MyDateTime defines model for datetime.
type MyDateTime = time.Time

// Error error response body
type Error struct {
	MyDateTime *MyDateTime `json:"createdAt,omitempty"`
	MyDateTime *MyDateTime `json:"updatedAt,omitempty"`
}
```

podhmo:
- fix #1091


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because alias conflict handling now exists via the `DefineViaAlias` field and `DisableTypeAliasesForType` configuration option. The implementation uses a different approach but addresses the same underlying issue.

PR #1107, opened 2023-06-13
fix(validation): skip email validation when the types.Email is nullab…

henripqt:
# WHAT:

- skip email validation when unmarshaling a `*types.Email` with a nil value
- add more tests & a sentinel error

# WHY:

Currently validation is performed when unmarshaling empty emails even when it shouldn't (using a pointer type is akin to a nullable JSON field and therefore shouldn't be validated). This results in being unable to retrieve data using the generated client if the target API returns a null email in the payload even when this one isn't required.

# HOW:

- skip email validation when unmarshaling a `*types.Email` with a nil value

### DIFF
--- a/pkg/types/email.go
+++ b/pkg/types/email.go
@@ -5,12 +5,19 @@ import (
 	"errors"
 )
 
+// ErrValidationEmail is the sentinel error returned when an email fails validation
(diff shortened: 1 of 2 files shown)


kayn1:
👌 👍

PR #1124, opened 2023-06-26
support staticcheck st1005

Hyuga-Tsukui:
I want to resolve an issue where generated files are becoming noise due to being the target of `error strings should not be capitalized (ST1005)go-staticcheck.`

Changing the template like in this PR would solve the problem, but this approach may not be preferable because it could affect the generated code of users who are already using it.

### DIFF
--- a/examples/petstore-expanded/strict/api/petstore-server.gen.go
+++ b/examples/petstore-expanded/strict/api/petstore-server.gen.go
@@ -459,7 +459,7 @@ func (sh *strictHandler) FindPets(w http.ResponseWriter, r *http.Request, params
 			sh.options.ResponseErrorHandlerFunc(w, r, err)
 		}
 	} else if response != nil {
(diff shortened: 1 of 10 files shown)


jamietanna:
Thanks! I've re-raised this in #1205 as I wasn't able to push to your branch. Feel free to pull the changes into your branch, otherwise I'll merge #1205 next week - you'll still get authorship info, but won't get a "PR accepted", which I'm sure you'd like!

Hyuga-Tsukui:
Thank you! This will reduce the noise!

jamietanna:
All good - would you like to update this PR with the changes from #1205, or happy for me to merge #1205 as-is?

Hyuga-Tsukui:
Thank you for your thoughtful suggestion. I've incorporated the changes from #1205 into my #1124. If there are no issues, I'd appreciate it if you could merge #1124.

jamietanna (approved):
Thanks for this 👏🏽

Issue #1130, opened 2023-06-29
bug: generated server code for chi is not working


sarathsp06:
When the response is of type `text/plain` generated server code is trying to convert to []byte while the response object is a struct 
```yml
ForbiddenError:
      description: Access token is valid but does not have the required scope
      content:
        text/plain:
          schema:
            type: string
            example: Forbidden
```
generated code
```go
type GetAccounts403TextResponse struct{ ForbiddenErrorTextResponse }

func (response GetAccounts403TextResponse) VisitGetAccountsResponse(w http.ResponseWriter) error {
	w.Header().Set("Content-Type", "text/plain")
	w.WriteHeader(403)

	_, err := w.Write([]byte(response))
	return err
}

```

`_, err := w.Write([]byte(response))` fails at compile time

jamietanna:
Hey, can you please share a minimal example? This doesn't seem quite right - I can see a `text/plain` response of `type: string`, so am not sure where we're getting this struct generated from?

It'd be super helpful to see whether this is a `strict` server or just `chi`.

sarathsp06:
@jamietanna could you please check

PR #1132, opened 2023-06-29
fix(template): fix strict-interface template 

sarathsp06:
fix strict-interface template  to make sure text response content type uses string type.

This is in reference to #1130 . The code change is probably not semantically correct, happy to make appropriate changes

### DIFF
--- a/pkg/codegen/templates/strict/strict-interface.tmpl
+++ b/pkg/codegen/templates/strict/strict-interface.tmpl
@@ -38,7 +38,9 @@
 
         {{range .Contents}}
             {{$receiverTypeName := printf "%s%s%s%s" $opid $statusCode .NameTagOrContentType "Response"}}
-            {{if and $fixedStatusCode $isRef -}}
+            {{if eq .NameTag "Text" -}}
+                type {{$receiverTypeName}} string
+            {{else if and $fixedStatusCode $isRef -}}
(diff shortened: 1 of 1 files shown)


tarampampam:
@sarathsp06 @deepmap-marcinr Looks like this PR broke the echo strict server:

```yaml
  /healthz:
    get:
      tags: [Service]
      summary: Liveness probe
      description: Is the app alive or dead?
      operationId: livenessProbe
      responses:
        '200':
          description: Ok
          content: {text/plain: {example: OK}}
          headers: {X-Request-Id: {schema: {$ref: '#/components/schemas/RequestID'}, description: Request ID}}
```

image

<details>
  <summary>Text</summary>

```go
type LivenessProbe200TextResponse string

func (response LivenessProbe200TextResponse) VisitLivenessProbeResponse(w http.ResponseWriter) error {
	w.Header().Set("Content-Type", "text/plain")
	w.Header().Set("X-Request-Id", fmt.Sprint(response.Headers.XRequestId))
	w.WriteHeader(200)

	_, err := w.Write([]byte(response.Body))
	return err
}
```

</details>

<details>
  <summary>Config</summary>

```yaml
package: openapi

generate:
  echo-server: true
  strict-server: true
  models: true
  embedded-spec: true
```

</details>

jamietanna:
This is being reverted as part of https://github.com/oapi-codegen/oapi-codegen/pull/1773

PR #1134, opened 2023-06-30
[Gorilla] update the `MiddlewareFunc` type for compatibility

namco1992:
This PR ported the fix https://github.com/deepmap/oapi-codegen/pull/578 to gorilla/mux, so the gorilla/mux users can have a consistent behaviour too.


namco1992:
@jamietanna @deepmap-marcinr I suppose this change will be ok since the last one #578 doesn't seem to have that much of a repercussion. Thanks!

jamietanna:
@namco1992 mind addressing the conflicts? After that - following the case we've already done this in a somewhat-but-not-that-bad breaking change, I'll merge it

Issue #1121, opened 2023-06-20
use of AnyOf generates struct type names starting with a number


Krulknul:
Trying to compile my schema using this command:
`oapi-codegen -package generated openapi/api.yaml > api.gen.go`
I'm on version `v1.13.0`
openapi: 3.0.0

This error is given:
```
error generating code: error formatting Go code
: generated.go:263:16: '_' must separate successive digits (and 1 more errors)
````

I've tried to create a minimal example where this issue occurs:

This is a simple schema with cats, dogs and rats.
[api.yaml.txt](https://github.com/deepmap/oapi-codegen/files/11807506/api.was.yaml.txt)

This is what's written to stderr:
[api.gen.go.txt](https://github.com/deepmap/oapi-codegen/files/11807527/api.gen.go.txt)

You can see the error at the top and bottom of the go file. 
At line 263 and 320, invalid Go is generated.

This works up until version 1.12.1, but after that, it is broken.

daleef-rahman:
The issue seems to be in the client code only. Looks like client is generated by default option in `v1.13.0`. You can try skipping the client code generation by using the command 
`oapi-codegen -package generated -generate types,server,spec api.was.yaml > api.gen.new2.go`

Alternatively, if you want to generate the client as well, you can do a small refactoring in the spec file. That should also work.
[api.was.yaml.txt](https://github.com/deepmap/oapi-codegen/files/11939176/api.was.yaml.txt)

I will try to investigate and fix the root cause as my first contribution to this repo 🙂

jvgutierrez:
I'm hitting this issue trying to generate the client code for USDA Food Central API (https://app.swaggerhub.com/apis/fdcnal/food-data_central_api/1.0.1). Is feasible to work around this issue without modifying the spec?

alejo-rivera:
> I'm hitting this issue trying to generate the client code for USDA Food Central API (https://app.swaggerhub.com/apis/fdcnal/food-data_central_api/1.0.1). Is feasible to work around this issue without modifying the spec?

Found myself here for the same reason. I cloned his fork from #1178 and got a workable client. I look forward to see this finished and merged though!

vasyl-haievyi:
Thank you @alejo-rivera  for the suggestion. I also was able to generate a valid client while using the fix. I have issued this issue while working with Hapio APIs - https://docs.hapio.io/hapio-api-definition/Hapio-API.v1.yaml

```go
...
var dest struct {
[...]

PR #1136, opened 2023-07-05
Fix: Use of AnyOf inline generates struct type names starting with a number

daleef-rahman:
https://github.com/deepmap/oapi-codegen/issues/1121 reports that use of AnyOf inline generates invalid go code for client (struct type names starting with a number). 
This PR fixes 2 issues:
1. When status code is the first element of path, PathToTypeName function will generate type name starting with a number
2. When a property have UnionElements and we use PathToTypeName to generate a reference type, we are not creating the corresponding struct for it

Closes https://github.com/deepmap/oapi-codegen/issues/1121

### DIFF
--- a/internal/test/strict-server/chi/server.gen.go
+++ b/internal/test/strict-server/chi/server.gen.go
@@ -951,6 +951,20 @@ type UnionExample200ResponseHeaders struct {
 	Header2 int
 }
 
+type UnionExample200ApplicationAlternativePlusJSONResponse struct {
+	Body    Example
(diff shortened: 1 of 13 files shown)


deepmap-marcinr:
I'm happy to merge this.

Please rebase, then run `go generate ./...` from the top level, and fix that linter error.

daleef-rahman:
> I'm happy to merge this.
> 
> Please rebase, then run `go generate ./...` from the top level, and fix that linter error.

Thank you for your message. Will get this done today.

PR #1143, opened 2023-07-11
fix error handler signature

hnarimiya:
Hello,
I have modified the ErrorHandler method signature.

Changed to receive err instead of receiving error message as string.

This is because an error message cannot set an appropriate error object to a response such as api when handling multi errors.

This is an existing breaking change.

However, it seems that there are many cases where the update work does not take that much and there are more merits.

Could you please consider including it in the next minor version?

### DIFF
--- a/pkg/chi-middleware/oapi_validate.go
+++ b/pkg/chi-middleware/oapi_validate.go
@@ -18,7 +18,7 @@ import (
 )
 
 // ErrorHandler is called when there is an error in validation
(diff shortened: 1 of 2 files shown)


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because error handlers already receive both the `error` object and `http.Request`. The current signature is: `RequestErrorHandlerFunc func(w http.ResponseWriter, r *http.Request, err error)`.

Issue #1144, opened 2023-07-11
Return error in fiber-middleware.ErrorHandler 


illiafox:
It seems that the whole [fiber-middleware](https://github.com/deepmap/oapi-codegen/tree/master/pkg/fiber-middleware) package was fully copied from [gin-middleware](https://github.com/deepmap/oapi-codegen/tree/master/pkg/gin-middleware). 

Unlike gin, fiber explicitly returns errors in handlers and does not have neither [`ctx.Abort()`](https://pkg.go.dev/github.com/gin-gonic/gin#Context.Abortl) or [`ctx.Error()`](https://pkg.go.dev/github.com/gin-gonic/gin#Context.Error) method

```go
// fiber.go

// Handler defines a function to serve HTTP requests.
type Handler = func(*Ctx) error
``` 

[fibermiddleware.Options](https://pkg.go.dev/github.com/deepmap/oapi-codegen/pkg/fiber-middleware@v1.13.0#Options) has [ErrorHandler](https://pkg.go.dev/github.com/deepmap/oapi-codegen/pkg/fiber-middleware@v1.13.0#ErrorHandler) field, but it does not return error as it was designed for [`gin.Context.Abort()`](https://pkg.go.dev/github.com/gin-gonic/gin#Context.Abortl) or [`gin.Context.Error()`](https://pkg.go.dev/github.com/gin-gonic/gin#Context.Error) method call inside.
 
https://github.com/deepmap/oapi-codegen/blob/ad62d73d8d22aa251d06b490b259790919899eab/pkg/fiber-middleware/oapi_validate.go#L49-L50

https://github.com/deepmap/oapi-codegen/blob/ad62d73d8d22aa251d06b490b259790919899eab/pkg/fiber-middleware/oapi_validate.go#L75-L85

`ErrorHandler` should 

1. Pass error instead of string so that developers could use `errors.As/Is`.
2. Return error in order to handle it by fiber middlewares

Code example:

```go

type ErrorHandler func(c *fiber.Ctx, err error, statusCode int) error

...

err := ValidateRequestFromContext(c, router, options)
if err != nil {
	if options != nil && options.ErrorHandler != nil {
		return options.ErrorHandler(c, err, http.StatusBadRequest)
	} else {
		return fiber.NewError(http.StatusBadRequest, err.Error())
	}
}
```

PR #1145, opened 2023-07-11
Change ErrorHandler signature

illiafox:
Similar to #1143, but for fiber
Original issue: #1144 

Due to differences in gin and fiber error handling, `ErrorHandler` option was modified in order to support fiber. Message type was changed to support `errors.Is/As`

```go
// ErrorHandler is called when there is an error in validation
type ErrorHandler func(c *fiber.Ctx, message error, statusCode int) error
```

Usage example: 

```go
fibermiddleware.OapiRequestValidatorWithOptions(swagger, &fibermiddleware.Options{
	ErrorHandler: func(c *fiber.Ctx, message error, statusCode int) error {
                return c.Status(statusCode).JSON(fiber.Map{
			"ok":    false,
			"error": message.Error(),
		}) // .JSON error is returned instead of being logged manually. Fiber middlewares could handle them
	},
})
```

### DIFF
--- a/pkg/fiber-middleware/oapi_validate.go
+++ b/pkg/fiber-middleware/oapi_validate.go
@@ -106,7 +106,7 @@ func ValidateRequestFromContext(c *fiber.Ctx, router routers.Router, options *Op
 		default:
 			// This should never happen today, but if our upstream code changes,
 			// we don't want to crash the server, so handle the unexpected error.
-			return fmt.Errorf("error validating route: %s", err.Error())
+			return fmt.Errorf("error validating route: %w", err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


illiafox:
Since we return error explicitly, all errors wrappings `fmt.Errorf("something happened: %s", err)` must  be replaced with `fmt.Errorf("something happened: %w", err)` in order to use `errors.Is/As`

```go
switch e := err.(type) {
	case *openapi3filter.RequestError:
		// We've got a bad request
		// Split up the verbose error by lines and return the first one
		// openapi errors seem to be multi-line with a decent message on the first
		errorLines := strings.Split(e.Error(), "\n")
		return fmt.Errorf("error in openapi3filter.RequestError: %s", errorLines[0])
	case *openapi3filter.SecurityRequirementsError:
		return fmt.Errorf("error in openapi3filter.SecurityRequirementsError: %w", e)
	default:
		// This should never happen today, but if our upstream code changes,
		// we don't want to crash the server, so handle the unexpected error.
		return fmt.Errorf("error validating request: %w", err)
	}
}
```
 So that we can determine in ErrorHandler whether error is BadRequest or Unauthorized 

```go
fibermiddleware.OapiRequestValidatorWithOptions(&specCopy, &fibermiddleware.Options{
	ErrorHandler: func(c *fiber.Ctx, message error, statusCode int) error {
		var securityRequirementsError *openapi3filter.SecurityRequirementsError
		if errors.As(message, &securityRequirementsError) {
			return api.SendUnauthorizedError(c, securityRequirementsError)
		}
		var requestError *openapi3filter.RequestError
		if errors.As(message, &requestError) {
			return api.SendBadRequestException(c, requestError)
		}
			
		return api.SendInternalError(c, message)
	},
```

jamietanna:
Thank you for this, and sorry it's taken so long for us to get to it - as part of https://github.com/oapi-codegen/nethttp-middleware/pull/35 we're looking at what a "better" OpenAPI validation middleware's error handling would look like and once that's merged, we'll look at pulling that into each of the separate middleware packages as needed

PR #1149, opened 2023-07-17
Load from stdin

const-tmp:
Allows pipelines e.g.
```
swagger-cli bundle -rt yaml openapi.yaml | oapi-codegen \
        -generate types,client,server,strict-server,spec \
        -package api \
        -o api/test.gen.go \
        -
```
Depends on https://github.com/getkin/kin-openapi/commit/b81a83eceea58ec4030e33c045385af38eaa3ddd


pkg/codegen/templates/strict/strict-interface.tmpl
```diff
@@ -63,6 +63,12 @@
                 }
             {{end}}
 
+            {{if eq .NameTag "JSON" | and $fixedStatusCode $isRef -}}
```
deepmap-marcinr, line 66: This is not related to the commit message. Perhaps split it out into a separate PR?

PR #1151, opened 2023-07-21
Support encoding.TextUnmarshaler to bind query params

yseto:
Path parameters now supports encoding.TextUnmarshaler interface.
https://github.com/deepmap/oapi-codegen/pull/404

Query parameters do not.

I changed to call UnmarshalText instead of  reflect.Set if the destination type implements  encoding.TextUnmarshaler


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because TextUnmarshaler support was implemented in 2021 via commit `dbe1736` ("Support standard encoding.TextUnmarshaler interface (#404)"), which predates this PR.

PR #1152, opened 2023-07-26
Fix: Correct `SilenceServersWarning` condition

gmtstephane:
This PR reverse the condition to logs the Warning so that the option `SilenceServersWarning: true` actually silence the warning


gmtstephane:
fix #1153

PR #1154, opened 2023-07-27
chore:ignore vendor

tylitianrui:
chore:ignore vendor


jamietanna:
Hey, not sure why this is needed, but I don't see it adding any value here

PR #1156, opened 2023-07-30
Resolve strict-server mode failure when handling multiple binary request bodies (e.g., jpeg or png)

neilli-sable:
I am using the following OpenAPI definition:

```yaml
openapi: 3.0.3
info:
  title: my api
  version: 1.0.0
paths:
  /image:
    put:
      summary: put image
      operationId: PutImage
      requestBody:
        content:
          image/jpeg:
            schema:
              type: string
              format: binary
          image/png:
            schema:
              type: string
              format: binary
      responses:
        200:
          description: succeed
          content:
            application/json:
              schema:
                type: object
                properties:
                  result:
                    type: string
                    description: result
```

This OpenAPI definition seems to be syntactically correct, and there should be no issues with code generation using the default settings.

However, when I enable strict-server mode, the code generated is as follows:
```go
type PutImageRequestObject struct {
	Body io.Reader
	Body io.Reader
}
```
Since the same field name "Body" is used twice, this is invalid syntax in Go.

To address this, I have created a fix. In the contentType switch, when determining Body as Binary, I store only the first occurrence in the bodyDefinitions to resolve this issue.


deepmap-marcinr:
Can you run this at the top level of the repo?

```
go generate ./...
go test ./...
```

It'll tell us for sure if nothing is broken, and please include any updated files in your commit.

neilli-sable:
Thank you for your response.

After rebasing my changes onto the latest master branch, I run the following commands:
```sh
go generate ./...
go test ./...
```

I would appreciate it if you could review it again.

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the strict HTTP template now properly handles multiple bodies with content-type-based conditional routing.

PR #1167, opened 2023-08-01
Added back the Mapping of JsonName 

dewrich:


jamietanna (changes requested):
Before I review this, can we please remove the changes in package import path I see in a few places? I can see `deepmap` has been replaced with your org.

PR #1174, opened 2023-08-03
Allow setting globalState.spec without Generate

derbylock:
We use oapi-codegen as a library.
E.g. we use public functions (e.g. codegen.OperationDefinitions()) inside our tool to make our specific additional code generation logic.
Currently the globalState.spec variable is used in refPathToGoType() and is set only in the Generate() function, so we have a panic when we use public functions without the Generate() invocation.

I suppose it is nice to have an ability to set the globalState.spec variable without the Generate() function invocation.


jamietanna:
Looks interesting - I'll get a review shortly. I notice that GitHub Actions hasn't triggered on this PR, which is odd :thinking:

PR #1176, opened 2023-08-03
Adds SAML and SCIM to All Caps List

mcrwfrd:
Adds `SAML` and `SCIM` to the list of known acronyms that will remain all uppercase instead of being converted to CamelCase. 

SAML is an XML standard that allows secure web domains to exchange user authentication and authorization data. SCIM is a standard for automating the exchange of user identity information between identity domains, or IT systems.

Both SAML and SCIM are fairly common acronyms used in many code-bases today.

### DIFF
--- a/pkg/codegen/utils.go
+++ b/pkg/codegen/utils.go
@@ -168,8 +168,8 @@ func ToCamelCaseWithInitialism(str string) string {
 func replaceInitialism(s string) string {
 	// These strings do not apply CamelCase
 	// Do not do CamelCase when these characters match when the preceding character is lowercase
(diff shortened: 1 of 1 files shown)


jxsl13:
imo thos ehole problem seems to jave beem tackled incorrectly: https://github.com/fatih/camelcase

jamietanna:
@jxsl13 by "this whole problem" do you mean in terms of the way the existing code handles this?

jxsl13:
by the whole problem I mean that having a hardcoded list of exceptions might be not the best way to tackle this but to use libraries like the one above for properly splitting words

PR #1187, opened 2023-08-08
Generate types from components/securitySchemes to be used as a key in context.WithValue

nek023:
## Summary

Generate types from `components/securitySchemes` to be used as a key in `context.WithValue`.

## Background

For example, prepare the following API schema and configuration.

schemas.yaml

```yaml
paths:
  /example:
    get:
      summary: Example
      operationId: example
      security:
        - BearerAuth: []

components:
  securitySchemes:
    BearerAuth:
      type: http
      scheme: bearer
```

config.yaml

```yaml
package: schemas
generate:
  chi-server: true
  models: true
output: schemas.gen.go
output-options:
  skip-prune: true
```

If you run oapi-codegen, the following code is generated.

```go
const (
	BearerAuthScopes = "BearerAuth.Scopes"
)

func (siw *ServerInterfaceWrapper) Example(w http.ResponseWriter, r *http.Request) {
	ctx := r.Context()

	ctx = context.WithValue(ctx, BearerAuthScopes, []string{})

	handler := http.Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		siw.Handler.Example(w, r)
	}))

		handler = middleware(handler)
	}

	handler.ServeHTTP(w, r.WithContext(ctx))
}
```

The constant `BearerAuthScopes` is generated as a key for `context.WithValue`, but its type is a `string` and there is a risk of collision between packages.

> The provided key must be comparable and should not be of type string or any other built-in type to avoid collisions between packages using context. Users of WithValue should define their own types for keys.

https://pkg.go.dev/context#WithValue

If you use [staticcheck](https://staticcheck.dev/), it will also warn this.

https://staticcheck.dev/docs/checks/#SA1029

This can be avoided by defining a type as shown below.

```go
type bearerAuthContextKey string

const (
	BearerAuthScopes bearerAuthContextKey = "BearerAuth.Scopes"
)
```

In this PR, I have implemented to generate types from `components/securitySchemes` to be used as keys for `context.WithValue`, as shown above.

## Considerations 

While the implementation has been done to avoid breaking the existing code as much as possible, if users are directly using the underlying string (in this case, `"BearerAuth.Scopes`), they will no longer be able to extract values from the `context`.

However, this is considered to be a rare case.

```go
type contextKey string

const exampleContextKey contextKey = "example"

func main() {
	ctx := context.Background()
	ctx = context.WithValue(ctx, exampleContextKey, "hello")

	val := ctx.Value(exampleContextKey)
	fmt.Println(val) // => hello

	val = ctx.Value("example")
	fmt.Println(val) // => nil
}
```

### DIFF
--- a/pkg/codegen/templates/echo/v5/echo-wrappers.tmpl
+++ b/pkg/codegen/templates/echo/v5/echo-wrappers.tmpl
@@ -26,7 +26,7 @@ func (w *ServerInterfaceWrapper) {{.OperationId}} (ctx *echo.Context) error {
 {{end}}
 
 {{range .SecurityDefinitions}}
-    ctx.Set({{.ProviderName | sanitizeGoIdentity | ucFirst}}Scopes, {{toStringArray .Scopes}})
+    ctx.Set(string({{.ProviderName | sanitizeGoIdentity | ucFirst}}Scopes), {{toStringArray .Scopes}})
 {{end}}
 
 {{if .RequiresParamObject}}
--- a/pkg/codegen/templates/gin/gin-wrappers.tmpl
+++ b/pkg/codegen/templates/gin/gin-wrappers.tmpl
@@ -40,7 +40,7 @@ func (siw *ServerInterfaceWrapper) {{$opid}}(c *gin.Context) {
   {{end}}
 
 {{range .SecurityDefinitions}}
-  c.Set(({{.ProviderName | sanitizeGoIdentity | ucFirst}}Scopes), {{toStringArray .Scopes}})
+  c.Set(string({{.ProviderName | sanitizeGoIdentity | ucFirst}}Scopes), {{toStringArray .Scopes}})
 {{end}}
 
   {{if .RequiresParamObject}}
--- a/pkg/codegen/templates/iris/iris-middleware.tmpl
+++ b/pkg/codegen/templates/iris/iris-middleware.tmpl
@@ -35,7 +35,7 @@ func (w *ServerInterfaceWrapper) {{.OperationId}} (ctx iris.Context) {
 {{end}}
 
 {{range .SecurityDefinitions}}
-    ctx.Set({{.ProviderName | sanitizeGoIdentity | ucFirst}}Scopes, {{toStringArray .Scopes}})
+    ctx.Set(string({{.ProviderName | sanitizeGoIdentity | ucFirst}}Scopes), {{toStringArray .Scopes}})
 {{end}}
 
 {{if .RequiresParamObject}}


internal/test/schemas/schemas.gen.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	Access_tokenScopes = "access_token.Scopes"
```
jamietanna, line 28: I realise this is existing code, but would be nice to use a more Go-like name here :thinking: Maybe I'll raise a follow-up for this

internal/test/client/client.gen.go
```diff
@@ -24,6 +24,9 @@ type SchemaObject struct {
 	Role      string `json:"role"`
 }
 
+// openIdContextKey is the context key for OpenId security scheme
+type openIdContextKey string
```
jamietanna, line 28: would we get any benefits from

```suggestion
type openIdContextKey = string
```

I feel this would reduce the limited impact of a breaking change, but not sure if it's worth it for any other reason.

internal/test/schemas/schemas.gen.go
nek023: I think so, too.
Changing it would alter the existing variable name, leading to a breaking change, which is why I didn't address it in this PR.
I would appreciate it if you could raise a follow-up for this.

Maybe the following changes are needed:

```diff
diff --git a/pkg/codegen/codegen.go b/pkg/codegen/codegen.go
index 31978fe..1b0f478 100644
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -639,7 +639,7 @@ func GenerateTypesForSecuritySchemes(t *template.Template, schemes map[string]*o
 
 	for _, schemeName := range SortedSecuritySchemeKeys(schemes) {
 		// Generate a type to be used as a key in context.WithValue
-		goTypeName := LowercaseFirstCharacter(SchemaNameToTypeName(schemeName)) + "ContextKey"
+		goTypeName := LowercaseFirstCharacter(ToCamelCase(SchemaNameToTypeName(schemeName))) + "ContextKey"
 		goType := Schema{
 			GoType:      "string",
 			Description: fmt.Sprintf("is the context key for %s security scheme", schemeName),
diff --git a/pkg/codegen/templates/constants.tmpl b/pkg/codegen/templates/constants.tmpl
index ca9fcb9..6704452 100644
--- a/pkg/codegen/templates/constants.tmpl
+++ b/pkg/codegen/templates/constants.tmpl
@@ -1,7 +1,7 @@
 {{- if gt (len .SecuritySchemeProviderNames) 0 }}
 const (
 {{range $ProviderName := .SecuritySchemeProviderNames}}
-    {{- $ProviderName | sanitizeGoIdentity | ucFirst}}Scopes {{$ProviderName | sanitizeGoIdentity | lcFirst}}ContextKey = "{{$ProviderName}}.Scopes"
+    {{- $ProviderName | sanitizeGoIdentity | camelCase | ucFirst}}Scopes {{$ProviderName | sanitizeGoIdentity | camelCase | lcFirst}}ContextKey = "{{$ProviderName}}.Scopes"
 {{end}}
 )
 {{end}}
```

internal/test/client/client.gen.go
nek023: Yes, that approach would avoid a breaking change.
However, by using a type alias, it ends up being the same as using the string type, so it doesn't change the situation.

```go
package main

import (
	"context"
	"fmt"
)

type keyType1 = string // type alias
type keyType2 = string

const key1 keyType1 = "KEY"
const key2 keyType2 = "KEY"

func main() {
	ctx := context.Background()
	ctx = context.WithValue(ctx, key1, "VALUE")

	// key1 conflicts with key2
	fmt.Println(ctx.Value(key1)) // => VALUE
	fmt.Println(ctx.Value(key2)) // => VALUE (want: nil)
}
```

GRbit:
@jamietanna any updates on this? I would really love seeing this merged. Not like I'm the one who makes decision here, yet I think it's good to go.

mromaszewicz:
You're right, this is an oversight in my original code and we should fix it. I'm pulling this branch and fixing it up for the current state of the code before approving and merging. I think the small risk of breakage (that is easy to fix) is worth having correctness.

mromaszewicz:
@greptileai

mromaszewicz:
@greptileai

PR #1188, opened 2023-08-08
Added function to set date format

IvanProdaiko94:
I would like to have a way to define the date format from my own code.

### DIFF
--- a/pkg/types/date.go
+++ b/pkg/types/date.go
@@ -5,7 +5,11 @@ import (
 	"time"
 )
 
-const DateFormat = "2006-01-02"
+var DateFormat = "2006-01-02"
+
+func SetDateFormat(format string) {
+	DateFormat = format
+}
 
 type Date struct {
 	time.Time


alexwestside:
This looks like what I needed! Without that I must to write custom UnmarshalJSON for correct date converting.

IvanProdaiko94:
@jxsl13 it was constant before

jxsl13:
my bad

pkg/types/date.go
```diff
@@ -5,7 +5,11 @@ import (
 	"time"
 )
 
-const DateFormat = "2006-01-02"
+var DateFormat = "2006-01-02"
```
jamietanna, line 8: With this now as a `var` it'd not need a setter. I'd be considering making this unexported, and then provide a method as `DateFormat()` that returns the current value, so it's _less_ of a breaking change.

Alternatively, we just make this a `var` and no setter.

jamietanna (commented):
Although this is a solution, I'm unsure if this is the right one, given it'd change _every_ date type in use with the OpenAPI spec :thinking:

mromaszewicz:
Thank you for submitting this PR, @IvanProdaiko94 — but we need to close it because the file it targets (`pkg/types/date.go`) has been moved to a separate repo: deprecated in commit `5d1cceae` (Aug 2023) and removed in commit `bccdafdf` (Oct 2023) when types graduated to `github.com/oapi-codegen/runtime`. If you'd still like a configurable date format, please open a fresh PR against that repo. The maintainer also flagged a design concern (a global mutable date format affecting all `Date` instances) that would be worth addressing in any future iteration.

Thanks for the contribution.

PR #1190, opened 2023-08-09
Add x-order support follow up from #627 to keep order of fields

filintod:
An update on PR https://github.com/deepmap/oapi-codegen/pull/627 to be mergeable

------------------------
From that PR:

Fix for https://github.com/deepmap/oapi-codegen/issues/458.
Can be used when you need to set the order of fields in the model:
```
    NewPet:
      type: object
      required:
        - name
      properties:
        name:
          type: string
          x-order: 1
        tag:
          type: string
          x-order: 2
```


stevenh:
@filintod did you see [my comment](https://github.com/deepmap/oapi-codegen/pull/627#issuecomment-1312459697) which could make it much simpler for users, avoiding the cruft of having to set `x-order: <number>` on every element, instead having just a single `x-ordered: true` at the properties level.

I think this would be better approach for this feature.

filintod:
That is a good suggestion to add, but not sure possible so easily. By the time we get to the point here in the code, we have no knowledge of the struct, and golang parsing of maps is usually not deterministic, one could sort alphabetically though.  There is a new yaml [module](https://github.com/goccy/go-yaml) that does a nicer job of preserving order but that would be a bigger change here and would have to be tested/approved. Another option is to add the `x-order` extension automatically at the point we convert from golang struct to openapi format, but that is outside this package.

stevenh:
Thanks @filintod had a little look and you're right the information has already been lost :(

filintod:
@jamietanna could you take a look on this one please?

jamietanna (commented):
Couple of tweaks needed - I'll push them up shortly :+1: Thanks for the work on this, and apologies it's taken so long!

jamietanna:
Last thing to do is to add an `internal/test`, which I'll do shortly, then will merge

PR #1191, opened 2023-08-11
[Codegen] Update lint module for tests

eugenio-bolt:
* Replace `github.com/golangci/lint-1` with `golang.org/x/lint`
* Run `go mod tidy`
* This change is required for users of `oapi-codegen` that use `bazel`, as they can encounter errors when dependencies do not have a `go.mod` file, as is the case with `github.com/golangci/lint-1`


jamietanna:
Thanks @eugenio-bolt! I see that the version we're replacing is an older fork of the /x/ repo, so very much makes sense to do anyway!

Issue #970, opened 2023-02-19
Using generated union type as response returning empty json 


idleway:
I have request that returning in response type **Event** defined like:
```
    event:
      oneOf:
        - $ref: '#/components/schemas/eventTypeOnetimeEditableField'
        - $ref: '#/components/schemas/eventTypeRepeatableEditableField'
```
Generated types are:
```
type Event struct {
	union json.RawMessage
}
func (t Event) MarshalJSON() ([]byte, error) {
	b, err := t.union.MarshalJSON()
	return b, err
}

type PostDescribeEventById200JSONResponse Event
```
The problem is  **PostDescribeEventById200JSONResponse** on json marshalling returning empty json "{}". This happens because **PostDescribeEventById200JSONResponse** is not using MarshalJSON() of underlying struct **Event**.

Using type alias instead of underlying type solving my problem:
```
type PostDescribeEventById200JSONResponse = Event
```

miro-ko:
I can confirm that this is an issue indeed.

I fixed it by adding `MarshalJSON` method to (using your example) `PostDescribeEventById200JSONResponse`:
```go
func (c PostDescribeEventById200JSONResponse) MarshalJSON() ([]byte, error) {
	return Event(c).MarshalJSON()
}
```

gerhardwagner:
did you solve this with a template @idleway ?
@miro-ko did you add this manually to the generated file or do you have a nice way ?
or is there any lead how to fix this issue long term ?

I would create a PR if I had a starting point for a solution

srjinatl:
I have similar issue - in my case the field is just one of the fields that are contained in a containing response object.  The generated code for the strict server interface does not marshall that imbedded union as part of the code that gets generated for the 201 response creation.

gerhardwagner:
I spend some time digging around and this is what I found:

in the `strict-interface.tmpl`
https://github.com/deepmap/oapi-codegen/blob/ad62d73d8d22aa251d06b490b259790919899eab/pkg/codegen/templates/strict/strict-interface.tmpl#L44
we can see that the `=` for aliasing is bound to IsRef which is a function on the `Schema`
https://github.com/deepmap/oapi-codegen/blob/ad62d73d8d22aa251d06b490b259790919899eab/pkg/codegen/schema.go#L42-L44

digging deeper and following my response object I also found this in `GenerateGoSchema`:
[...]

PR #1231, opened 2023-09-03
Use type alias for strict response types

dan-j:
See #970...

@gerhardwagner suggested a fix which also works for me, so I've created a quick PR. I'm not familiar with the actual codebase so not sure if tests are going to pass, or how to add tests to get this merged.

I don't have a lot of time to go digging so any hints are appreciated and I am happy to contribute tests.


gerhardwagner:
would like to see this merged 🙃

dan-j:
Just added a comment to the issue #970 which might prevent this from being merged as-is as it could break other people's implementations

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because strict response types already conditionally use type aliases — aliases are used when the schema is a `$ref` and not an external ref.

Issue #1236, opened 2023-09-07
Incorrect JSON with read/write only field


ShouheiNishi:
In generated codes, read only field is pointer type.
If schema does not have additional property, this read only field with nil value is not exist in marshaled JSON.
But schema has additional property, this field with nil value is marshaled as JSON null.
If this field is not nullable, this is incorrect behavior.

Same issue was exist for write only field.

ShouheiNishi:
Test code is https://github.com/ShouheiNishi/oapi-codegen/commit/f17b2018416714273e2c4bf41e67eb536faccd3b.

```
$ (cd internal/test/issues ; go generate issue-1236/doc.go)
$ (cd internal/test/issues ; go test -v issue-1236/issue_test.go)
=== RUN   TestIssue1236
    issue_test.go:33:
                Error Trace:    /home/nishi/free5gc-deb/openapi5g/oapi-codegen/internal/test/issues/issue-1236/issue_test.go:33
                Error:          Not equal:
                                expected: map[string]interface {}{"Extra":0, "Nullable":interface {}(nil), "Required":0}
                                actual  : map[string]interface {}{"Extra":0, "Nullable":interface {}(nil), "ReadOnly":interface {}(nil), "Required":0, "WriteOnly":interface {}(nil)}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,5 +1,7 @@
                                -(map[string]interface {}) (len=3) {
                                +(map[string]interface {}) (len=5) {
                                  (string) (len=5) "Extra": (float64) 0,
                                  (string) (len=8) "Nullable": (interface {}) <nil>,
                                - (string) (len=8) "Required": (float64) 0
                                + (string) (len=8) "ReadOnly": (interface {}) <nil>,
                                + (string) (len=8) "Required": (float64) 0,
                                + (string) (len=9) "WriteOnly": (interface {}) <nil>
                                 }
                Test:           TestIssue1236
--- FAIL: TestIssue1236 (0.00s)
FAIL
FAIL    command-line-arguments  0.007s
FAIL
$
```

mromaszewicz:
[...]

PR #1238, opened 2023-09-07
Add nil check for read/write only property in generated MarshalJSON()

ShouheiNishi:
Add nil check for read/write only property in generated MarshalJSON()
This PR is fix for #1236.


ShouheiNishi:
I've created a new PR #1303 with a fix for this PR, so I'm closing it.

Issue #1142, opened 2023-07-10
v1.13.0 introduces lots of transitive dependencies to client library


pgier:
I'm only using oapi-codegen to generate a client library against an openapi spec file.  I noticed that when I changed my `go.mod` to use v1.13.0 instead of v1.12.4, there are lots of transitive dependencies added.  There seems to be a lot of stuff added that relate to server code and maybe should not be necessary for a client library.  

```
 go 1.20

-require github.com/deepmap/oapi-codegen v1.12.4
+require github.com/deepmap/oapi-codegen v1.13.0

 require (
        github.com/apapsch/go-jsonmerge/v2 v2.0.0 // indirect
+       github.com/bytedance/sonic v1.9.1 // indirect
+       github.com/chenzhuoyu/base64x v0.0.0-20221115062448-fe3a3abad311 // indirect
+       github.com/gabriel-vasile/mimetype v1.4.2 // indirect
+       github.com/gin-contrib/sse v0.1.0 // indirect
+       github.com/gin-gonic/gin v1.9.1 // indirect
+       github.com/go-playground/locales v0.14.1 // indirect
+       github.com/go-playground/universal-translator v0.18.1 // indirect
+       github.com/go-playground/validator/v10 v10.14.0 // indirect
+       github.com/goccy/go-json v0.10.2 // indirect
        github.com/google/uuid v1.3.0 // indirect
-       github.com/stretchr/testify v1.8.3 // indirect
+       github.com/json-iterator/go v1.1.12 // indirect
+       github.com/klauspost/cpuid/v2 v2.2.4 // indirect
+       github.com/labstack/echo/v4 v4.10.2 // indirect
+       github.com/labstack/gommon v0.4.0 // indirect
+       github.com/leodido/go-urn v1.2.4 // indirect
+       github.com/mattn/go-colorable v0.1.13 // indirect
+       github.com/mattn/go-isatty v0.0.19 // indirect
+       github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+       github.com/modern-go/reflect2 v1.0.2 // indirect
+       github.com/pelletier/go-toml/v2 v2.0.8 // indirect
+       github.com/twitchyliquid64/golang-asm v0.15.1 // indirect
+       github.com/ugorji/go/codec v1.2.11 // indirect
+       github.com/valyala/bytebufferpool v1.0.0 // indirect
+       github.com/valyala/fasttemplate v1.2.2 // indirect
[...]

PR #1243, opened 2023-09-08
Remove Iris dependency

ShouheiNishi:
This PR is to remove dependency of Iris module in "github.com/deepmap/oapi-codegen" and "github.com/oapi-codegen/runtime".

* Remove directory "pkg/runtime" and use module "github.com/oapi-codegen/runtime" instead.
* Remove directory "pkg/iris-middleware" and use module "github.com/oapi-codegen/iris-middleware" instead.
* Split package "github.com/oapi-codegen/runtime/strictmiddleware/iris" to new module "github.com/oapi-codegen/iris-runtime"


jamietanna:
Thanks for this, but see comment in https://github.com/deepmap/oapi-codegen/issues/1142#issuecomment-1713493020 - we'll do this at some point, but it's probably not worth keeping this PR up to date as we'll not be doing it for a few months at least.

PR #1246, opened 2023-09-08
Make arrays use concrete types, not aliases.

jkj:
Currently when you create an array type using something like:

```yaml
components:
  schemas:
    my_item:
      type: object
      properties:
        name:
          type: string
        age:
          type: integer
    example:
      type: array
      items:
        $ref: '#/components/schemas/my_item'
```
the generator produces a type alias for the array:
```go
// Example defines model for example.
type Example = []MyItem
```

This causes problems, particularly for defining methods. This patch ensures that when arrays are generated, they use a concrete type and not a type alias. For example:
```go
// Example defines model for example.
type Example []MyItem
```

The following example indicates why aliases are bad:
```go
type item struct {
        name string
        age int
}

type Alias = []item
type Concrete []item

// func (*Alias) wont_compile() {
//     invalid receiver type *[]item
// }

func (*Concrete) will_compile() {
}
```

This is especially problematic if you need to create methods for things like a JSON `Scan()` method or an SQL `Value()` method for the type.


zsolt-p:
> need to create methods for things like a JSON Scan() method or an SQL Value() method for the type.

I have this same exact use case, too.

my workaround is:
- add the schema to my spec anyway (I do this so I can still `$ref` the type, etc)
- `-exclude-schemas=mySchema` when I generate
- implement the concrete type by hand outside the genned code within the same package, where I can implement its methods, etc

Issue #1221, opened 2023-08-29
"make test" is failed with go version 1.21.0


ShouheiNishi:
Using go version 1.21.0, "make test" is failed.

```
$ git status
HEAD detached at deepmap/master
nothing to commit, working tree clean
$ git log | head
commit 5d1cceae130876e666916b06ef4801419810bd80
Author: Jamie Tanna <jamie.tanna@elastic.co>
Date:   Mon Aug 28 17:35:00 2023 +0100

    Deprecate `pkg/runtime` and `pkg/types`

    As we're now using the oapi-codegen/runtime package for this, we should
    mark the existing functionality as deprecated.

    This also ignores any internally flagged deprecation warnings, as we
$ go version
go version go1.21.0 linux/amd64
$ make test
git ls-files go.mod '**/*go.mod' -z | xargs -0 -I{} bash -xc 'cd $(dirname {}) && go test -cover ./...'
++ dirname examples/go.mod
+ cd examples
+ go test -cover ./...
go: updates to go.mod needed; to update it:
        go mod tidy
++ dirname go.mod
+ cd .
+ go test -cover ./...
go: updates to go.mod needed; to update it:
        go mod tidy
++ dirname internal/test/go.mod
+ cd internal/test
+ go test -cover ./...
go: updates to go.mod needed; to update it:
        go mod tidy
make: *** [Makefile:28: test] Error 123
$
```

jamietanna:
Thanks for flagging this - it looks like it's not `go test` but `go mod tidy` / `make tidy` that's failing - as we're building with Go 1.20 right now, we're unable to reproduce this right now, but can try and look at it soon

jamietanna:
I can see the same with running Go 1.21 too, which wants the following diff:

```diff
diff --git examples/go.mod examples/go.mod                                                                                                
index dd8a415..36c09a5 100644                                                                                                             
--- examples/go.mod                                                                                                                       
+++ examples/go.mod                                                                                                                       
@@ -1,6 +1,8 @@                                                                                                                           
 module github.com/deepmap/oapi-codegen/examples                                                                                          
                                                                                                                                          
-go 1.20                                                                                                                                  
+go 1.21                                                                                                                                  
+                                                                                                                                         
+toolchain go1.21.0 
```

jamietanna:
Through discussion in the Gopher Slack, this is due to Go 1.21 spotting that some (transitive) dependencies require Go 1.21, so this is trying to explicitly target it.

[...]

PR #1251, opened 2023-09-11
Build project with Go 1.20 and Go 1.21

jamietanna:
As part of the upgrade to Go 1.21, the Go toolchain now requires the
`go` directive to match the maximum Go version in use in dependencies.

This leads to any transitive dependency on this library to result in a
requirement of the consuming project moving to Go 1.21.

Although in a lot of cases this may not be as problematic, it forces
consumers to migrate, which we don't need to do in this case.

As noted in #1221, dependencies on Fiber and Iris have caused Go
toolchain changes due to be required, which then requires consumers to
do the same.

By making sure we build across both the current venison and the next
version of Go - taking care to only support the two major Go versions in
use by the Go team - we can hopefully pick up on issues like this
sooner, albeit #1221 was raised super speedily!

This is a slightly more complex setup than just specifying the version
manually, but to avoid the risk of not updating it in all the places, we
can at a minimum use the `go.mod` version for the current targeted
version.

---

Will merge this as-is before #1249, to make sure we have a known "bad" build on `master`, then we can have #1249 fix it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,17 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      # perform matrix testing to give us an earlier insight into issues with different versions of supported major versions of Go
+      matrix:
+        # strategy is used to allow us to pin to a specific Go version, or use the version available in our `go.mod`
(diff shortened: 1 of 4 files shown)


jamietanna:
Confirmed that #1249 is solved by this, so merging as-is to show the state of `master`.

Issue #1250, opened 2023-09-11
Strict server with additionalProperties


ShouheiNishi:
The strict server cannot return JSON object with additionalProperties.
It seems that wrapper for strict server ignore custom MarshalJSON().

We do not create the minimal test code for this issue yet.

ShouheiNishi:
Test code is https://github.com/ShouheiNishi/oapi-codegen/commit/19373ab96086a7d8c6cec20059c918fb2e42db54.

alehed:
We are also being bitten by this issue.

LucasMaciuga:
I also experienced this issue. As a workaround I figured out that if you wrap the schema with the "additionalProerties" tag into a another schema like:
```yaml
---
components:
  schemas:
    DataWrapper: # using this will marshal/unmarshal the additionalProperties
      type: object
      properties:
        data:
          $ref: "#/components/schemas/WithAdditionalProperies"
      example:
        data:
          fixedProperty: "Hello World"
          additionalPropertyFoo: "Bar"
          key: "value"
          unknown: "known"
    WithAdditionalProperies: # using this will NOT marshal/unmarshal the additionalProperties
      type: object
      properties:
        fixedProperty:
         type: string
      additionalProperty: true ## or {}
      example:
        fixedProperty: "Hello World"
        additionalPropertyFoo: "Bar"
        key: "value"
        unknown: "known"
```
it is working.

FilipVranesevicKNEX:
I found a root cause for this issue. Here is an example:

```
paths:
  /test/path1:
    post:
      description: desc
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              additionalProperties:
                type: string
              properties:
                p1:
                  type: string

      responses:
        "200":
          description: 200 response
```

this spec generates following code (skipping not important bits):
```
...

// PostTestPath1JSONBody defines parameters for PostTestPath1.
type PostTestPath1JSONBody struct {
	P1                   *string           `json:"p1,omitempty"`
	AdditionalProperties map[string]string `json:"-"`
}

// PostTestPath1JSONRequestBody defines body for PostTestPath1 for application/json ContentType.
type PostTestPath1JSONRequestBody PostTestPath1JSONBody

...

// Override default JSON handling for PostTestPath1JSONBody to handle AdditionalProperties
func (a *PostTestPath1JSONBody) UnmarshalJSON(b []byte) error {

...

// PostTestPath1 operation middleware
[...]

PR #1203, opened 2023-08-23
Fix external reference propagation

ShouheiNishi:
Fix for #1202

PR #1253, opened 2023-09-11
Fix/strict server additional properties

ShouheiNishi:
Add code to convert objects containing additionalProperties in strict server to the original type before encoding to correctly JSON-encode them.

This PR is fix for #1250 
And this PR requires #1203, otherwise some test codes are broken by reference error.


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because AdditionalProperties handling was improved via commit `9aa9d5c` ("Allow merging of schemas with additionalProperties"). The codegen now has robust support for this.

PR #1260, opened 2023-09-13
Add http.Request parameter to the error handler

dan20130911:
The reason for this change is the HTTP request has information that the error handler can use.
In our project we use the path of the HTTP request in the error handler.

### DIFF
--- a/pkg/chi-middleware/oapi_validate.go
+++ b/pkg/chi-middleware/oapi_validate.go
@@ -17,7 +17,7 @@ import (
 )
 
 // ErrorHandler is called when there is an error in validation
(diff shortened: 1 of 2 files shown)


jamietanna:
Thanks for raising this - I think this is one we've looked at before, but right now this is a breaking change to our users, so we'd need to put this behind a configurable option in `output-options` - is that something you'd be happy doing?

jamietanna (changes requested):
On the face of it, seems good, but see my comment re breaking change

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because error handlers already include the `http.Request` parameter in their signature.

PR #1294, opened 2023-09-29
Respect charsets in JSON content types

christiannicola:
This commit fixes an issue with the codegeneration - when specifying a charset in a JSON content type, the resulting request / response bodies are missing the JSON serializer.

The charset is usually redundant, since JSON should usually be encoded using UTF-8 (see https://www.rfc-editor.org/rfc/rfc8259#section-8.1), however there are some code generators out there that do not honor this (for example: https://github.com/OpenAPITools/openapi-generator/issues/12797)

### DIFF
--- a/internal/test/strict-server/chi/server.gen.go
+++ b/internal/test/strict-server/chi/server.gen.go
@@ -982,6 +982,20 @@ func (response UnionExample200JSONResponse) VisitUnionExampleResponse(w http.Res
 	return json.NewEncoder(w).Encode(response.Body.union)
 }
 
+type UnionExample200ApplicationJSONCharsetUTF8Response struct {
+	Body    Example
+	Headers UnionExample200ResponseHeaders
+}
+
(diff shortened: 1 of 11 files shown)


jamietanna:
Just to confirm, why are we needing to add this? Because some code generators default to adding the charset so we want to make sure this aligns?

Or because you're integrating this project with a generator that adds the charset and therefore can't interact with this?

This may make sense (for some of this) to be an opt-in feature in a compatability option, but let's see

christiannicola:
@jamietanna The code generation does not correctly handle a content type like `application/json; charset=utf8` - The generated source code returns structs without the JSON serializer.

That alone would be an improvement. 

Defining a `charset` for JSON request / response bodies is usually redundant, since JSON should be encoded using UTF-8 anyway, but it is a valid approach for defining JSON payloads within an OpenAPI schema.

jamietanna:
Gotcha, thanks for confirming!

jamietanna:
Hey @christiannicola apologies for the delay on this one, this will now be handled correctly due to https://github.com/deepmap/oapi-codegen/pull/1385 :rocket:

PR #1295, opened 2023-09-29
add option x-go-allof-embed-refs

lahabana:
This is useful when working with extending APIs

I'm opening this as draft as it's missing:

- [ ] Docs
- [ ] More tests

This is useful for us as we have apis that expand APIs already defined externally. This enables us to avoid a lot of mapping code.

I didn't want to extend `old-merge-schemas` as this seems likely to be removed in the future.
We handle the original issue by just refusing the generate go code which would result in a non openAPI compliant (which was the original motivation for not using embedded objects ref: https://github.com/deepmap/oapi-codegen/issues/531)

Signed-off-by: Charly Molter <charly.molter@konghq.com>


lahabana:
There seems to be more recent options

Issue #1263, opened 2023-09-14
Wrong type when using x-go-type-skip-optional-pointer


sonu27:
When using `x-go-type-skip-optional-pointer` on an optional header, invalid code gets generated which cannot compile.

```yaml
openapi: 3.0.3
info:
  title: Form API
  version: '1.0'
paths:
  /api/v1/forms/contact:
    post:
      operationId: sendContactForm
      parameters:
        - in: header
          name: X-Real-IP
          required: false
          schema:
            type: string
            x-go-type-skip-optional-pointer: true # the problem
      responses:
        '204':
          description: Contact form sent
      tags:
        - form

```

This results in:
```go
// ------------- Optional header parameter "X-Real-IP" -------------
	if valueList, found := headers[http.CanonicalHeaderKey("X-Real-IP")]; found {
		var XRealIP string
		n := len(valueList)
		if n != 1 {
			siw.ErrorHandlerFunc(w, r, &TooManyValuesForParamError{ParamName: "X-Real-IP", Count: n})
			return
		}

		err = runtime.BindStyledParameterWithLocation("simple", false, "X-Real-IP", runtime.ParamLocationHeader, valueList[0], &XRealIP)
		if err != nil {
			siw.ErrorHandlerFunc(w, r, &InvalidParamFormatError{ParamName: "X-Real-IP", Err: err})
			return
		}

		params.XRealIP = &XRealIP // the problem, should be without the pointer
		// the fix: params.XRealIP = XRealIP

	}
```

marcomayer:
Stumbled into the same issue with:

```
        - name: pct
          in: query
          required: false
          schema:
            type: integer
            default: 0
            x-go-type-skip-optional-pointer: true
            minimum: 0
            maximum: 100

```
Code:
`	err = runtime.BindQueryParameter("form", true, false, "pct", r.URL.Query(), &params.Pct)`

Getting:
`panic: reflect: call of reflect.Value.IsNil on int Value
`
Maybe I misunderstood how/when x-go-type-skip-optional-pointer should be used?

emiago:
This should be fixed, or at least avoid panic and crashing services! Do your nil checks for reflect!

mromaszewicz:
[...]

PR #1300, opened 2023-10-04
Ignore x-go-type-skip-optional-pointer in parameters

ShouheiNishi:


ShouheiNishi:
To resolve issue #1263, this PR make that fields in request parameter struct are always pointer, even if x-go-type-skip-optional-pointer is true.

sonu27:
@deepmap-marcinr Could this be looked at please?

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the codebase has evolved significantly since this was written. Any investigation of this edge case would need to be done fresh against the current parameter handling code.

Issue #954, opened 2023-02-07
Error on header value of empty string when it isn't required


renom:
1. Header param isn't required in openapi spec.
2. Header value is set but the value is empty string.
3. Generated code reports an error `Invalid format for parameter <param_name>: parameter '<param_name>' is empty, can't bind its value` because of [this code](https://github.com/deepmap/oapi-codegen/blob/master/pkg/runtime/bindparam.go#L45) in `BindStyledParameterWithLocation`.

delca85:
We are also hitting this bug.
I don't think this is the expected behavior: if an header has type string, and it's not required, it could be empty without raising any binding error.

delca85:
Couldn't this be fixed simply checking if `dest` here has type `string`?

[This](https://github.com/deepmap/oapi-codegen/blob/0cfaaa77a7d2a6baa495469b706cc04b1463387b/pkg/runtime/bindparam.go#L45) condition would become:

``` if value == "" && reflect.TypeOf(dest).Kind() != reflect.String```

johnnyaug:
Running into this as well.

renom:
What approach should be used to fix this?
Should we pass `required` to `BindStyledParameterWithLocation` like it's passed to `BindQueryParameter`?

renom:
I've created draft pull requests to fix this issue.

renom:
What are further steps to have them merged?

jamietanna:
Apologies for the delay with this, we're looking to make a slight refactor to https://github.com/oapi-codegen/runtime/pull/19 to make this a non-breaking change, then hopefully will release the changes in the next few days. Then we'll need the code in #1315 / https://github.com/deepmap/oapi-codegen/pull/1353 to be updated accordingly.

PR #1315, opened 2023-10-11
Pass .Required to BindStyledParameterWithLocation and BindStyledParameter

renom:
Fixes https://github.com/deepmap/oapi-codegen/issues/954
Works along with oapi-codegen/runtime/pull/14


renom:
It's okay, I can update the code. I have a few questions though. How do I know if the code is updated right? Will it be enough to update runtime to v1.1.0 and check that cmd/oapi-codegen builds fine?

jamietanna:
Thank you! 

Yes, exactly - once bumping `runtime`, as long as the new code compiles, that will at least give us confidence that it's wired in correctly.

We may want to add an integration test (in https://github.com/deepmap/oapi-codegen/blob/master/internal/test/parameters/parameters_test.go) to validate that the underlying issue in #954 is resolved

renom:
I updated `runtime` to `v1.1.0`, the code builds fine on my laptop. Can it be merged now?

pkg/codegen/templates/fiber/fiber-middleware.tmpl
```diff
@@ -27,7 +27,7 @@ func (siw *ServerInterfaceWrapper) {{$opid}}(c *fiber.Ctx) error {
   }
   {{end}}
   {{if .IsStyled}}
-  err = runtime.BindStyledParameter("{{.Style}}",{{.Explode}}, "{{.ParamName}}", c.Params("{{.ParamName}}"), &{{$varName}})
+  err = runtime.BindStyledParameter("{{.Style}}", {{.Explode}}, {{.Required}}, "{{.ParamName}}", c.Params("{{.ParamName}}"), &{{$varName}})
```
jamietanna, line 30: I think this will fail now, as we've introduced [a new function `BindStyledParameterWithOptions`](https://github.com/oapi-codegen/runtime/blob/v1.1.0/bindparam.go#L71) - mind refactoring these templates to use that new function now?

jamietanna (commented):
Thank you 👏🏽 well also need to run `make generate` and then commit the changes

renom:
Ready to be merged.

jamietanna:
Thank you @renom 👏 cc @itpavelkozlov

PR #1359, opened 2023-11-22
Use non-deprecated function names for oapi-codegen/runtime

cimitan:
Hi, I noticed linting was failing in a project of mine and that was due to deprecated packages being referenced. I assume this could be a leftover / oversight from v2.0.0

I did not run tests to make sure other bits are not being broken by this change, but it looked safe to me at first.

Thanks


jamietanna:
Thanks very much @cimitan! Yeah this was due to the oapi-codegen/runtime library's https://github.com/oapi-codegen/runtime/releases/tag/v1.1.0 release, and we'd not come back to update here. Thanks for the fix, I've also added the other initialism we updated, and regenerated all the code.

PR #1366, opened 2023-12-01
replace deprecated function, regenerate code base

w32blaster:
the `runtime.JsonMerge()` was deprecated, the `runtime.JSONMerge()` should be used instead. The `union.tmpl` still uses the old function which makes a linter angry.

This is a very small and simple PR where a new function is being used in this template, plus the code was re-generated.

Thank you!


jamietanna:
Thank you @w32blaster, apologies I didn't see this before https://github.com/deepmap/oapi-codegen/pull/1359 which I've now merged, which should address this :bow:

PR #1368, opened 2023-12-01
Fix /pets endpoint in examples returning null when there's no pet

DavesBorges:
When no pet is found the examples pet store implementations are returning `null` as the http response body as opposed to empty array (`[]`).

Since the zero value of a slice is nil the pets slice should be created using make so that if no pet matches the query we can return an initialized slice. Otherwise the json package will encode the slice as nil which breaks the api contract.

```diff
func (p *PetStore) FindPets(...){

-    var result []Pet
+   var result []Pet = make([]Pet, 0)
 
    // Code for populating the result
 
    // Return or write the result to the response
}
```

The go json encoder and decoder doesn't complain about the nil value, so when consuming the api using the go standard library (as the tests do) everything is fine but if we would use clients written with languages/tools (javascript for example) they would fail to parse the response body. That's why I  added additional checks to the examples api's tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,7 +369,7 @@ Short example:
 ```go
 type PetStoreImpl struct {}
 func (*PetStoreImpl) GetPets(ctx context.Context, request GetPetsRequestObject) (GetPetsResponseObject, error) {
-    var result []Pet
+    var result []Pet = make([]Pet, 0)
 	// Implement me
     return GetPets200JSONResponse(result), nil
 }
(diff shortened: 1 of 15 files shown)


DavesBorges:
One could argue that the generated strict server should correctly return `[]` if the response being returned from the handler isn't nil when an array is expected but then we would have to start doing the same thing for child elements deep down the JSON tree. 
So I think it's better not to change the generated strict server but instead add a tip/info to the Readme documenting the issue

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the approach (initializing empty slices with `make()` to return `[]` instead of `null`) was questioned by maintainers, and the examples have changed since then.

PR #1374, opened 2023-12-05
Add support for json mediatype with version

obouchet:
## Description

When json media type has a version provided (like `appication/json;version=v1`), the generator ignores it and doesn't produce helper structs in response object (very useful for the parsing of the http response into a struct).

This PR adds the support for this type of json mediatype definition.

### DIFF
--- a/pkg/util/isjson.go
+++ b/pkg/util/isjson.go
@@ -3,5 +3,5 @@ package util
 import "strings"
 
 func IsMediaTypeJson(mediaType string) bool {
-	return mediaType == "application/json" || strings.HasSuffix(mediaType, "+json")
+	return strings.HasPrefix(mediaType, "application/json") || strings.HasSuffix(mediaType, "+json")
 }
--- a/pkg/util/isjson_test.go
+++ b/pkg/util/isjson_test.go
@@ -32,6 +32,11 @@ func TestIsMediaTypeJson(t *testing.T) {
 			want:      true,
 		},
 		{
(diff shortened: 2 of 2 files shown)


jamietanna:
Hmm, this is an interesting one - and one I've dealt with at a previous company - because _technically_ this is non-standard behaviour - https://www.iana.org/assignments/media-types/application/json notes that it is not possible to specify a `version` parameter to the media type.

Not completely against it, but thinking that I'd be tempted to do more strict validation (namely via https://gitlab.com/jamietanna/content-negotiation-go/-/blob/main/mediatype.go?ref_type=heads) to make sure it truly is a JSON-compatible thing.

jamietanna:
Hey @obouchet thanks for this! I've raised https://github.com/deepmap/oapi-codegen/pull/1386 which doesn't go quite as far as I'd initially suggested, but uses the standard library's `mime.ParseMediaType` to make sure that we get the params separated from the media type itself.

Would you like me to add you as a `Co-authored-by` in the commit, as you deserve some props for the change, but I wanted to ask if you're happy with it, first.

obouchet:
Hi @jamietanna!

Thank you so much for this. I like your solution and it's a good step to a full fledge solution if you like to go that far.
Very nice of you to offer me being the co-author of this but I did nothing than just ask for the change and propose a quick and dirty solution for it...
You can stay the sole author for this for sure!

Thank you again!

PR #1376, opened 2023-12-06
Update echo params to be idiomatic

natehart:
Note To Reviewer: I'm aware that you [don't usually want to change variable names in generated code](https://github.com/deepmap/oapi-codegen#contributing), but this particular change should not affect existing code, as interfaces don't need to have the same parameter _names_ as their implementations.

## Change Description

Echo uses `c` as the `echo.Context` parameter name, differentiating it from the widely used `ctx` for a `context.Context`.

This commit updates the generated code to use the `c` convention to bring it in line with Echo's common practice and make generated code more readable.

This change will not break existing code.  Existing `ServerInterface` do not need to have the same parameter names, so all existing code that uses `ctx echo.Context` will be able to ignore the change (or update when it's convenient).

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,13 +127,13 @@ Echo server.
 ```go
 type ServerInterface interface {
     //  (GET /pets)
-    FindPets(ctx echo.Context, params FindPetsParams) error
+    FindPets(c echo.Context, params FindPetsParams) error
     //  (POST /pets)
-    AddPet(ctx echo.Context) error
+    AddPet(c echo.Context) error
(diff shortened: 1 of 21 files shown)


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the templates have been reworked since this PR was written, and renaming the Echo context parameter from `ctx` to `c` is a cosmetic change that would break existing user code for no functional benefit.

PR #1383, opened 2023-12-12
fix body defenitions generation

Egor-mn:


jamietanna:
Would you mind confirming that the changes that are being generated by this make sense? :thinking:

PR #1385, opened 2023-12-13
Fix: Handle code generation for all multipart content-types

### LINKED ISSUE #1381: Content-type "multipart/related" is not worked
When "Content-type" of request or response body is "multipart/related", multipart operation in strict server is not worked correctly.

ShouheiNishi:
In the current code, functions http.Request.MultipartReader() and multipart.Writer.FormDataContentType() are used for the processing of the mutlipart message, but this function is only supported when contentType is "multipart/form-data".
When contentType is specified in OpenAPI definition other than "multipart/form-data", code that compatible with other contentTypes is generated without using these functions.

This PR fix #1381.


jamietanna (approved):
Thanks @ShouheiNishi :raised_hands:

PR #1386, opened 2023-12-13
Support media type parameters for `IsMediaTypeJson`

jamietanna:
Although not _technically_ a valid extension to JSON according to [the
media type definition], it's got a number of services using it for a `v`
or `version` parameter, so we should support this as a valid option.

This makes sure that we parse it as a valid media type, performing the
same checks as we did before by ignoring the parameters.

[the media type definition]: https://www.iana.org/assignments/media-types/application/json

---

As an alternative to https://github.com/deepmap/oapi-codegen/pull/1374

### DIFF
--- a/pkg/util/isjson.go
+++ b/pkg/util/isjson.go
@@ -1,7 +1,14 @@
 package util
 
-import "strings"
+import (
+	"mime"
+	"strings"
+)
 
 func IsMediaTypeJson(mediaType string) bool {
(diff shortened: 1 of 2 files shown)


jamietanna:
Awaiting response from https://github.com/deepmap/oapi-codegen/pull/1374#issuecomment-1853826252

obouchet (approved):

Issue #1378, opened 2023-12-07
Missing package name on type from common API spec when referencing a path operation from another spec in generated code


markuswustenberg:
We're hitting a bug where generated code doesn't compile when using API specs referencing each other. This may or may not be related to #695, so I'm opening a separate issue just to be sure.

The setup is this (also available at https://github.com/maragudk/oapi-codegen-bug):

- An `api.yaml` spec references a path in a `common.yaml` spec.
- In `common.yaml`, a path `/ping` references a schema `PingResponse` in the same file.

When the spec is generated for `api.yaml`, `PingResponse` doesn't have the `externalRef0` package name prefix, and so the generated code can't compile.

This is with v2 of oapi-codegen.

**api.yaml**

```yaml
version: "3.0.0"

paths:
  /ping:
    $ref: "common.yaml#/paths/~1ping"
```

**config.yaml**

```yaml
package: api
generate:
  gin-server: true
  embedded-spec: true
  strict-server: true
  models: true
import-mapping:
  common.yaml: github.com/maragudk/oapi-codegen-bug/oapi/common
output: api.gen.go
```

**common.yaml**

```yaml
version: "3.0.0"

paths:
  /ping:
    get:
      responses:
        200:
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/PingResponse"

components:
  schemas:
    PingResponse:
      type: object
      properties:
        message:
          type: string
```

**config.yaml**

```yaml
package: common
generate:
  gin-server: true
  embedded-spec: true
  strict-server: true
  models: true
output: common.gen.go
```

markuswustenberg:
I also saw this problem inside functions with statements such as `var t YourType` when using path parameters.

PR #1387, opened 2023-12-18
Fix: Refer to external refs correctly in strict interfaces

jamietanna:
The current approach doesn't often work when generating types that are
in external packages, unless that package is also a strict server's
definition, which isn't always the case.

---

Noticed while working on https://github.com/deepmap/oapi-codegen/issues/1378

### DIFF
--- a/pkg/codegen/templates/strict/strict-fiber-interface.tmpl
+++ b/pkg/codegen/templates/strict/strict-fiber-interface.tmpl
@@ -48,7 +48,7 @@
                 type {{$receiverTypeName}} struct{ {{$ref}}{{.NameTagOrContentType}}Response }
                 {{end}}
             {{else if and (not $hasHeaders) ($fixedStatusCode) (.IsSupported) -}}
-                type {{$receiverTypeName}} {{if eq .NameTag "Multipart"}}func(writer *multipart.Writer)error{{else if .IsSupported}}{{if .Schema.IsRef}}={{end}} {{.Schema.TypeDecl}}{{else}}io.Reader{{end}}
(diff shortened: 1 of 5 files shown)


pkg/codegen/templates/strict/strict-fiber-interface.tmpl
```diff
@@ -42,11 +42,13 @@
...
                 {{ if and (not $hasHeaders) ($fixedStatusCode) (.IsSupported) (eq .NameTag "Multipart") -}}
                 type {{$receiverTypeName}} {{$ref}}{{.NameTagOrContentType}}Response
+                {{else if $isExternalRef -}}
+                type {{$receiverTypeName}} struct { {{$ref}} }
                 {{else -}}
                 type {{$receiverTypeName}} struct{ {{$ref}}{{.NameTagOrContentType}}Response }
                 {{end}}
             {{else if and (not $hasHeaders) ($fixedStatusCode) (.IsSupported) -}}
-                type {{$receiverTypeName}} {{if eq .NameTag "Multipart"}}func(writer *multipart.Writer)error{{else if .IsSupported}}{{if .Schema.IsRef}}={{end}} {{.Schema.TypeDecl}}{{else}}io.Reader{{end}}
+                type {{$receiverTypeName}} {{if eq .NameTag "Multipart"}}func(writer *multipart.Writer)error{{else if .IsSupported}}{{if and .Schema.IsRef (not .Schema.IsExternalRef)}}={{end}} {{.Schema.TypeDecl}}{{else}}io.Reader{{end}}
```
jamietanna, line 51: :stop_sign: this doesn't work

PR #1389, opened 2023-12-18
Fix: Ensure external refs are propagated to generated code

### LINKED ISSUE #1378: Missing package name on type from common API spec when referencing a path operation from another spec in generated code

jamietanna:
As noted in #1378, there are cases where a complex set of `$ref`s
between multiple files can lead to broken generated code, which does not
correctly import the package that has been prepared for the external
reference.

We can handle this by looking up any references, where there is a `.Ref`
passed into the type, and then iterate through relevant children.

This requires we handle the updating in-place for these by using a bit
of pointer + indexing fun.

This also adds a relevant test case to validate the fix.

Closes #1378.

---

**Note** that this won't fix every case - we'll review it in https://github.com/deepmap/oapi-codegen/issues/1440


pkg/codegen/operations.go
```diff
@@ -817,6 +821,19 @@ func GenerateResponseDefinitions(operationID string, responses map[string]*opena
 				NameTag:     tag,
 				Schema:      contentSchema,
 			}
+
+			// TODO finish comment
```
jamietanna, line 825: TODO

pkg/codegen/operations.go
```diff
@@ -817,6 +821,19 @@ func GenerateResponseDefinitions(operationID string, responses map[string]*opena
...
+			parts := strings.SplitN(parentRef, "#", 2)
+			if parentRef != "" && len(parts) == 2 {
+				// TODO finish comment
```
jamietanna, line 828: TODO

markuswustenberg:
I'm getting the following error when trying this out:

```
error generating code: error formatting Go code <generated code here> filename.go:5746:70: expected 'IDENT', found 'struct' (and 7 more errors)
```

pkg/codegen/operations.go
```diff
@@ -817,6 +821,19 @@ func GenerateResponseDefinitions(operationID string, responses map[string]*opena
...
+				// if a locally-defined, but `$ref`'d parent, then we need to make sure that we make this an external ref
```
jamietanna, line 829: rephrase

pkg/codegen/schema.go
```diff
@@ -43,6 +43,13 @@ func (s Schema) IsRef() bool {
 	return s.RefType != ""
 }
 
+func (s Schema) IsExternalRef() bool {
+	if !s.IsRef() {
+		return false
+	}
+	return strings.Contains(s.RefType, ".")
+}
```
jamietanna, line 51: unused :thinking:

jamietanna: Ah, it is used, but in templates

internal/test/issues/issue-1378/bionicle/generate.go
```diff
@@ -0,0 +1,3 @@
+package bionicle
+
+//go:generate go run github.com/deepmap/oapi-codegen/v2/cmd/oapi-codegen --config=config.yaml ../bionicle.yaml
```
markuswustenberg, line 3: Will this work without a Go workspace set up in the environment? Is `go generate` smart enough to know this is the same repo, and run the local version?

internal/test/issues/issue-1378/common/config.yaml
```diff
@@ -0,0 +1,9 @@
+package: common
+generate:
+  gorilla-server: true
+  embedded-spec: true
+  strict-server: true
+  models: true
+output-options:
```
markuswustenberg, line 7: Probably not for this PR, but it would make sense to not prune code that is used in dependent APIs by default, instead of having to set this explicitly.

pkg/codegen/externalref.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ensureExternalRefsInRequestBodyDefinitions ensures that when an externalRef (`$ref` that points to a file that isn't the current spec) is encountered, we make sure we update our underlying `RefType` to make sure that we point to that type.
+// This only happens if we have a non-empty `ref` passed in, and that `ref` isn't pointing to something in our file
+// NOTE that the pointer here allows us to pass in a reference and edit in-place
+func ensureExternalRefsInRequestBodyDefinitions(defs *[]RequestBodyDefinition, ref string) {
```
markuswustenberg, line 11: I think it would be easier to understand if the param was `defs []*RequestBodyDefinition`, since the underlying def could be changed directly in that case, and you wouldn't need the `(*defs)[i] = rbd`. Or is there something I'm missing?

markuswustenberg (approved):
In my limited understanding of the internals of this project, LGTM. :D

pkg/codegen/externalref.go
jamietanna: That's a fair shout.

I went for this solution, as I could just pass it in from `operations.go`:

```go
			ensureExternalRefsInRequestBodyDefinitions(&bodyDefinitions, pathItem.Ref)
```

Whereas if I went with that solution, I'd need to create a new slice that had a pointer to each value within `bodyDefinitions` :thinking:

internal/test/issues/issue-1378/bionicle/generate.go
jamietanna: Short answer is yep!

Via https://www.jvt.me/posts/2022/06/15/go-tools-dependency-management/ and how that works.

As this is a submodule, we need to do https://github.com/deepmap/oapi-codegen/blob/5cdb5595bd38525b0a8f19cd322b91c0fb34b4d6/internal/test/go.mod#L1-L5 too

internal/test/issues/issue-1378/common/config.yaml
jamietanna: Raised #1442 to cover this

PR #1390, opened 2023-12-18
Add OperationID in client request header

chrisngyn:
I utilized this library to create a client from an OpenAPI specification, and now I aim to develop an interceptor for monitoring outgoing HTTP requests using Prometheus. 
However, the request path varies based on path parameters, making it impractical to use it as a label value for Prometheus metrics. 
As a solution, I propose adding an additional header (`Operation-Id`) to each HTTP request. This adjustment simplifies the process of writing the monitoring interceptor.

It just a naive solution. Feel free to suggest me the better solutions.

### DIFF
--- a/examples/authenticated-api/echo/api/api.gen.go
+++ b/examples/authenticated-api/echo/api/api.gen.go
@@ -189,6 +189,8 @@ func NewListThingsRequest(server string) (*http.Request, error) {
 		return nil, err
 	}
 
+	req.Header.Add("Operation-Id", "ListThings")
+
 	return req, nil
 }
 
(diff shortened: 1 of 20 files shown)


chrisngyn:
Almost file changes are the result when I ran `make generate`. I just only a one line at `pkg/codegen/templates/client.tmpl`
```
req.Header.Add("Operation-Id", "{{$opid}}")
```

jamietanna:
Hey @chrisngyn thanks for raising this.

Unfortunately this isn't something that we'd like to make the default.

You may be able to do this with a `RequestEditorFn` that's added into the client - it'll still require the OperationID, so maybe adding a field into the `ctx` that can be retrieved by a `RequestEditorFn` would work.

If you were to tweak this and make it an opt-in flag (see #1404 for where we've added an opt-in flag) then that'd be great.

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the maintainer suggested an opt-in flag approach and the redesign was never followed up on. The code has changed significantly since then.

Issue #695, opened 2022-07-27
Generated client code response object does not correctly reference a response from a separate yaml file


jace1717:
Summary: the response object in a the generated client code does not honor the $ref of the of response, if the $ref is referring to content in a different spec file. 

Setup:

**PackageA/config.yaml**
```
package: packageA
generate:
  models: true
  embedded-spec: true
  client: true
output-options:
  skip-prune: true
import-mapping:
  ../packageB/spec.yaml: github.com/deepmap/oapi-codegen/internal/test/externalref/packageB
skip-fmt: true
output: externalref.gen.go
```
**PackageA/spec.yaml**
```
paths:
  /api/test:
    post:
      operationId: TestApi
      summary: test
      description: test
      responses:
        '200':
          description: 'success'
          $ref: ../packageB/spec.yaml#/components/responses/ResponseC
```
**PackageB/spec.yaml**
```
components:
  schemas:
    ObjectB:
      type: object
      properties:
        name:
          type: boolean
  responses:
    ResponseC:
      description: ResponseC
      content:
        application/json:
          schema:
            $ref: '#/components/schemas/ObjectB'
```

**Generated client code for packageA**
`//go:generate go run github.com/deepmap/oapi-codegen/cmd/oapi-codegen --config=config.yaml spec.yaml`
```
type TestApiResponse struct {
	Body         []byte
	HTTPResponse *http.Response
	JSON200      *ObjectB
}
```

**Expected:**
```
type TestApiResponse struct {
	Body         []byte
	HTTPResponse *http.Response
	JSON200      *externalRef0.ObjectB
}
```

I've also attached a modified version of the `internal\test\externalref` code that demonstrates this issue. 
[externalref.zip](https://github.com/deepmap/oapi-codegen/files/9201439/externalref.zip)

jamietanna:
This _might_ be related to https://github.com/deepmap/oapi-codegen/issues/626

jamietanna:
Hey @jace1717 this issue _seems_ to be resolved now - would you mind confirming? I've raised #1391 which takes the code you've provided and moves it to its own test directory, can you confirm if I've done anything wrong, or if it looks like it is resolved?

mromaszewicz:
[...]

PR #1391, opened 2023-12-18
Add test code for #695

jamietanna:
As part of #695.


internal/test/issues/issue-695/packageA/spec.yaml
```diff
@@ -0,0 +1,57 @@
...
+#         name:
+#           type: string
+#   responses:
+#     ResponseC:
+#       description: ResponseC
+#       content:
+#         application/json:
+#           schema:
+#             $ref: '#/components/schemas/ObjectB'
+
```
jamietanna, line 57: may be unnecessary?

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because external reference handling was comprehensively fixed in commit `ad5eada`, which likely supersedes this test case.

PR #1400, opened 2023-12-21
Test matrix for response of strict server

ShouheiNishi:


jamietanna:
Thanks @ShouheiNishi :raised_hands: I'll try and gt to this soon

PR #1404, opened 2023-12-21
Allow generating `nullable.Nullable` for nullable properties

sonasingh46:
This PR adds capability to use [Nullable](https://github.com/oapi-codegen/nullable) types via opt in flag for Nullable types

Ref Issue: https://github.com/deepmap/oapi-codegen/issues/1039


examples/go.mod
```diff
@@ -2,7 +2,10 @@ module github.com/deepmap/oapi-codegen/v2/examples
 
 go 1.20
 
-replace github.com/deepmap/oapi-codegen/v2 => ../
+replace (
+	github.com/deepmap/oapi-codegen/v2 => ../
+	github.com/oapi-codegen/runtime v1.1.0 => ../../../sonasingh46/runtime
```
jamietanna, line 7: FYI this can be done locally using https://go.dev/doc/tutorial/workspaces so we don't require changes in the `go.mod`s

pkg/codegen/configuration.go
```diff
@@ -35,6 +35,7 @@ type GenerateOptions struct {
 	Client        bool `yaml:"client,omitempty"`         // Client specifies whether to generate client boilerplate
 	Models        bool `yaml:"models,omitempty"`         // Models specifies whether to generate type definitions
 	EmbeddedSpec  bool `yaml:"embedded-spec,omitempty"`  // Whether to embed the swagger spec in the generated code
+	OptionalType  bool `yaml:"optional-type,omitempty"`  // Whether to generate nullable type for nullable fields
```
jamietanna, line 38:
```suggestion
	NullableType  bool `yaml:"nullable-type,omitempty"`  // Whether to generate nullable type for nullable fields
```

pkg/codegen/schema.go
```diff
@@ -104,6 +104,27 @@ func (p Property) GoTypeDef() string {
 	return typeDef
 }
 
+// GoTypeDefWithOptionalType generates type based on Optional type define here
+// https://github.com/oapi-codegen/runtime/pull/26/files/ ToDo: Change the PR to
+// actual file in repo once it is merged
+func (p Property) GoTypeDefWithOptionalType() string {
```
jamietanna, line 110: this can actually be collapsed into the existing method. We already have [existing code](https://github.com/deepmap/oapi-codegen/blob/839a661de960d4b11725c46c5a300d66642b0bcc/pkg/codegen/schema.go#L99) that uses global state in code generation, so I'm happy us adding another case in there, instead of introducing a new method

examples/go.mod
sonasingh46: Thanks. I will try that. I readyy never used go workspace

internal/test/go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/kataras/iris/v12 v12.2.6-0.20230908161203-24ba4e8933b9
 	github.com/labstack/echo/v4 v4.11.3
+	github.com/oapi-codegen/nullable v1.0.0
```
jamietanna, line 16: I've pushed a new release (docs changes only) so worth updating too:

```suggestion
	github.com/oapi-codegen/nullable v1.0.1
```

pkg/codegen/schema.go
```diff
@@ -94,13 +94,17 @@ func (p Property) GoFieldName() string {
...
+		return "nullable.Nullable[" + typeDef + "]"
+	}
 	if !p.Schema.SkipOptionalPointer &&
 		(!p.Required || p.Nullable ||
 			(p.ReadOnly && (!p.Required || !globalState.options.Compatibility.DisableRequiredReadOnlyAsPointer)) ||
 			p.WriteOnly) {
 
 		typeDef = "*" + typeDef
 	}
+
```
jamietanna, line 107: needed?

pkg/codegen/schema_test.go
jamietanna, line 1: Did we say we'd add this file first as a prefactor PR?

sonasingh46: Here I have raised a PR. Will rebase this to 1423 once 1423 gets merged
https://github.com/deepmap/oapi-codegen/pull/1423

pkg/codegen/schema.go
sonasingh46: No. Wondering if this gets caught by the linter? I remember running `make lint`. 
I will remove it anyways.

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,298 @@
...
+		},
+		Popularity: &popularity,
+		SearchBox:  &searchBox,
+	}
+
+	expectedJSON := []byte(`{"breakout_group":{"name":"test-bk"},"department":{"alias_name":"dep-alias-name","name":"dep-name"},"home_country":"US","name":"test-name","organization":{"alias_name":"org-alias-name","name":"org-name"},"organization_location":{"country_name":"US","state_name":"California"},"popularity":10,"search_box":5}`)
+
+	gotJSON, err := json.Marshal(updateReq)
+	require.NoError(t, err)
+	fmt.Println(string(gotJSON))
```
jamietanna, line 90:
```suggestion
```

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,298 @@
...
+		},
+		Popularity: &popularity,
+		SearchBox:  &searchBox,
+	}
+
+	expectedJSON2 := []byte(`{"breakout_group":{"name":"test-bk"},"department":{"alias_name":"dep-alias-name","name":"dep-name"},"name":"test-name","organization":{"alias_name":"org-alias-name","name":"org-name"},"popularity":10,"search_box":5}`)
+
+	gotJSON2, err := json.Marshal(updateReq2)
+	require.NoError(t, err)
+	fmt.Println(string(gotJSON2))
```
jamietanna, line 116:
```suggestion
```

pkg/codegen/schema.go
```diff
@@ -681,10 +685,15 @@ func GenFieldsFromProperties(props []Property) []string {
 
 		field += fmt.Sprintf("    %s %s", goFieldName, p.GoTypeDef())
 
-		omitEmpty := !p.Nullable &&
-			(!p.Required || p.ReadOnly || p.WriteOnly) &&
+		commonConditions := (!p.Required || p.ReadOnly || p.WriteOnly) &&
```
jamietanna, line 688: Not sure I like this name - maybe `shouldOmitEmpty`?

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,298 @@
...
+			Client:       true,
+			Models:       true,
+			EmbeddedSpec: true,
+			NullableType: true,
+		},
+	}
+
+	_, err = codegen.Generate(swagger, opts)
+	require.NoError(t, err)
+}
```
jamietanna, line 39: Was this just to check that generating the code works? If so, this shouldn't be needed - by having the `go:generate`s in this package, that's good enough, unless there's something else you're trying to test?

internal/test/issues/issue-1039/spec.yaml
jamietanna, line 1: Is this the minimal spec we could use? If we can trim it down it'd be appreciated, so it's clear for long-term maintenance what's actually important about this

sonasingh46: Update the spec.

internal/test/issues/issue-1039/issue_test.go
sonasingh46: Yes. Removed it.

pkg/codegen/configuration.go
```diff
@@ -35,6 +35,7 @@ type GenerateOptions struct {
 	Client        bool `yaml:"client,omitempty"`         // Client specifies whether to generate client boilerplate
 	Models        bool `yaml:"models,omitempty"`         // Models specifies whether to generate type definitions
 	EmbeddedSpec  bool `yaml:"embedded-spec,omitempty"`  // Whether to embed the swagger spec in the generated code
+	NullableType  bool `yaml:"nullable-type,omitempty"`  // Whether to generate nullable type for nullable fields
```
jamietanna, line 38: Sorry, would this be able to move to the `OutputOptions`? That's a better place for it

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+//go:embed spec.yaml
+var spec []byte
+
+func ToStringPointer(str string) *string {
+	return &str
+}
```
jamietanna, line 19: nit: but doesn't need to be exported:
```suggestion
func toStringPointer(str string) *string {
	return &str
}
```

jamietanna, line 19: alternatively with generics:
```suggestion
func ptr[T any](v T) *T {
	return &v
}
```

pkg/codegen/schema_test.go
```diff
@@ -213,3 +213,243 @@ func TestProperty_GoTypeDef(t *testing.T) {
 		})
 	}
 }
+
+func TestPropertyNullable_GoTypeDef(t *testing.T) {
```
jamietanna, line 217: This test should be collapsed in with the above, no? At the very least:

```suggestion
func TestProperty_GoTypeDef_nullable(t *testing.T) {
```

To indicate what it's testing

jamietanna (commented):
Couple of comments, and just want to make sure that we've made sure this is the _minimal_ set of specs + tests needed to validate the functionality?

jamietanna:
Draft release notes:

<details>


## Notable features

### `Nullable` types, 

It's possible that you want to be able to determine whether a field isn't sent, is sent as `null` or has a value.

For instance, if you had the following OpenAPI property:

```yaml
    S:
    type: object
    properties:
        Field:
          type: string
          nullable: true
    required: []
```

The current behaviour in oapi-codegen is to generate:

```go
type S struct {
	Field *string `json:"field,omitempty"`
}
```

However, you lose the ability to understand the three cases, as there's no way to distinguish two of the types from each other:

- is this field not sent? (Can be checked with `S.Field == nil`)
- is this field `null`? (Can be checked with `S.Field == nil`)
- does this field have a value? (`S.Field != nil && *S.Field == "123"`)

Therefore, as requested in #1039, this is now possible to represent with the `nullable.Nullable` type from [our new library, oapi-codegen/nullable](https://github.com/oapi-codegen/nullable).

If you configure your generator's Output Options as so:

```yaml
output-options:
  nullable-type: true
```

You will now receive the following output:

```go
type S struct {
    Field nullable.Nullable[string] `json:"field,omitempty"`
}
```

**Note** that this is opt-in only, due to it being a break in existing signatures and behaviour.

You can find out more about how this works in a [blog post with further details](https://www.jvt.me/posts/2024/01/09/go-json-nullable/).

</details>

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,282 @@
...
+	patchReq2 := issue1039nullable.PatchRequest{
+		BreakoutGroup: &issue1039nullable.BreakoutGroup{
+			Name: ToStringPointer("test-bk"),
+		},
+		Organization: &issue1039nullable.Organization{
+			AliasName: ToStringPointer("org-alias-name"),
+			Name:      ToStringPointer("org-name"),
+		},
+		Popularity: &popularity,
+	}
```
jamietanna, line 73:
```suggestion
	// exclude some fields in patch request
	patchReq2 := issue1039nullable.PatchRequest{
		BreakoutGroup: &issue1039nullable.BreakoutGroup{
			Name: ToStringPointer("test-bk"),
		},
		// HomeCountry is omitted
		Organization: &issue1039nullable.Organization{
			AliasName: ToStringPointer("org-alias-name"),
			Name:      ToStringPointer("org-name"),
		},
		Popularity: &popularity,
		// SearchBox is omitted
	}
```

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,282 @@
...
+				assert.Falsef(t, obj.SearchBox.IsNull(), "search box should not be null")
+
+				assert.Falsef(t, obj.Popularity.IsSpecified(), "popularity should not be set")
+				assert.Falsef(t, obj.Popularity.IsNull(), "popularity should not be null")
+
+				assert.Falsef(t, obj.BreakoutGroup.IsSpecified(), "breakout group should not be set")
+				assert.Falsef(t, obj.BreakoutGroup.IsNull(), "breakout group should not be null")
+
+				// check for non-nullable field
+				assert.Truef(t, obj.HomeCountry == nil, "home country should be nil")
```
jamietanna, line 129:
```suggestion
				assert.Nilf(t, obj.HomeCountry, "home country should be nil")
```

internal/test/issues/issue-1039/type-config-nullabletype-disabled.yaml
```diff
@@ -0,0 +1,11 @@
+package: issue1039nullable
+generate:
+  models: true
+  embedded-spec: true
+output-options:
+  skip-prune: true
+output: nullable-disabled/types.gen.go
```
jamietanna, line 7: I think this would be clearer if this and the other package were named differently - `nullabledisabled` is a bit unclear at a first glance.

Maybe `defaultbehaviour`?

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,282 @@
...
+)
+
+//go:embed spec.yaml
+var spec []byte
+
+func ToStringPointer(str string) *string {
+	return &str
+}
+
+func TestNullableDisabled(t *testing.T) {
```
jamietanna, line 21: I'd also recommend moving this into the package with the default behaviour, or making it clearer what this is - it was a little confusing reading this one first, then seeing the new behaviour - maybe we should keep this next to the existing behaviour's generated code?

internal/test/issues/issue-1039/types.gen.go
```diff
@@ -0,0 +1,149 @@
...
+	Name *string `json:"name,omitempty"`
+}
+
+// PatchRequest A request to patch an existing user object.
+type PatchRequest struct {
+	// BreakoutGroup A breakout group which an user will be part of
+	BreakoutGroup nullable.Nullable[BreakoutGroup] `json:"breakout_group"`
+
+	// HomeCountry Home country of the user.
+	HomeCountry *HomeCountry `json:"home_country,omitempty"`
```
jamietanna, line 39: At first glance of the tests, I wasn't sure why this was `nil`

It may be worth us tweaking the spec to make the property names clearer what they're testing - i.e. this could be `OptionalButNotNullableProperty` then it's super clear what's being tested (and why it's being checked for `nil` instead of expecting a `nullable.Nullable`)

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,282 @@
...
+			},
+		},
+
+		{
+			name: "when search box is null ",
+			json: []byte(`{"search_box":null}`),
+			assert: func(t *testing.T, obj PatchRequest) {
+				t.Helper()
+				// check for nullable field
+				// check for nullable fields
```
jamietanna, line 223: duplicate?

internal/test/issues/issue-1039/types.gen.go
sonasingh46: This a very good point. Let me do that

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,282 @@
...
+				assert.Falsef(t, obj.Popularity.IsSpecified(), "popularity should not be set")
+				assert.Falsef(t, obj.Popularity.IsNull(), "popularity should not be null")
+
+				assert.Falsef(t, obj.BreakoutGroup.IsSpecified(), "breakout group should not be set")
+				assert.Falsef(t, obj.BreakoutGroup.IsNull(), "breakout group should not be null")
+			},
+		},
+
+		{
+			name: "when search box is null and organization has non zero value",
```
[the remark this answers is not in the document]
sonasingh46: I have checking that to ensure that because the other fields are not provided -- it should be basically not set and not null. But I think now when I am reading your comment, probably, we do not need to duplicate this in every test case.

internal/test/issues/issue-1039/defaultbehaviour/defaultbehaviour_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"encoding/json"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func ptr[T any](v T) *T {
+	return &v
+}
+
+func TestNullableDisabled(t *testing.T) {
```
jamietanna, line 13: did we want to have the same test shown for the Nullable types, so it's clear the difference in functionality?

internal/test/issues/issue-1039/type-config-defaultbehaviour.yaml
```diff
@@ -0,0 +1,11 @@
...
+generate:
+  models: true
+  embedded-spec: true
+output-options:
+  skip-prune: true
+output: defaultbehaviour/types.gen.go
+compatibility:
+  old-aliasing: true
+  old-merge-schemas: true
+  disable-flatten-additional-properties: true
```
jamietanna, line 11: are these definitely required to make this a minimal reproduction case?

internal/test/issues/issue-1039/type-config-defaultbehaviour.yaml
```diff
@@ -0,0 +1,11 @@
+package: defaultbehaviour
+generate:
+  models: true
+  embedded-spec: true
```
jamietanna, line 4: not needed any more?

internal/test/issues/issue-1039/types-config.yaml
```diff
@@ -0,0 +1,12 @@
+package: issue1039
+generate:
+  models: true
+  embedded-spec: true
```
jamietanna, line 4: as above

internal/test/issues/issue-1039/types-config.yaml
```diff
@@ -0,0 +1,10 @@
+package: issue1039
+generate:
+  models: true
+output-options:
+  skip-prune: true
+  nullable-type: true
+output: types.gen.go
+compatibility:
+  old-aliasing: true
+  old-merge-schemas: true
```
jamietanna, line 10: needed?

internal/test/issues/issue-1039/type-config-defaultbehaviour.yaml
```diff
@@ -0,0 +1,9 @@
+package: defaultbehaviour
+generate:
+  models: true
+output-options:
+  skip-prune: true
+output: defaultbehaviour/types.gen.go
+compatibility:
+  old-aliasing: true
+  old-merge-schemas: true
```
jamietanna, line 9: needed?

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func TestNullableTypesMarshal(t *testing.T) {
+	// include all fields in patch request
+	patchReq := PatchRequest{
+		ComplexOptionalNullable: nullable.Nullable[ComplexOptionalNullable]{
+			true: ComplexOptionalNullable{
+				AliasName: nullable.Nullable[string]{
+					true: "foo-alias",
+				},
```
jamietanna, line 23: I'd prefer to avoid the implementation details leaking - please use `NewNullableWithValue`

internal/test/issues/issue-1039/issue_test.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+func ptr[T any](v T) *T {
+	return &v
+}
+
+func TestNullableTypesMarshal(t *testing.T) {
+	// include all fields in patch request
+	patchReq := PatchRequest{
+		ComplexOptionalNullable: nullable.Nullable[ComplexOptionalNullable]{
```
jamietanna, line 19: can the order of fields in this function mirror what's in the other file? Will make it easier to diff

Issue #1301, opened 2023-10-04
Non-required header in the GIN strict server.


ShouheiNishi:
In GIN strict server, there seems to be no way to stop sending non-required headers in the response definition.

wc9384:
Is there a reason #1412 was never reviewed? Confirmed that this is still a problem.

rezass:
I would also love to see this merged. Omitting null response headers is needed for my usecase.

mromaszewicz:
The issue asked for a way to omit non-required response headers in the Gin strict server. Commit a4c36dc2 (PR #2301) added IsOptional()/IsNullable() to ResponseHeaderDefinition and updated strict-interface.tmpl to only emit w.Header().Set() for optional headers when the pointer is non-nil. A repro confirms that optional headers now generate as *string with nil-guard, while required headers remain direct values. PR #1412 (the community fix) is still open but the problem is resolved through a different PR. A HeadersImplicitlyRequired compat flag was also added for backward compatibility.

Closing as fixed. If you can still reproduce on a recent release, please reopen with the exact spec and version used.

PR #1412, opened 2023-12-28
Optional response header

ShouheiNishi:
To fix issue #1301, when header is optional, field for this is generate as pointer type. When this field is nil, corresponding header is not set in functions Visit*.
However, this change introduce some incompatibility.
So we add compatibility option.

And this PR included PR #1411.


mromaszewicz:
Thank you for submitting this PR, @ShouheiNishi — but we need to close it because it has been superseded by #2301 (commit `a4c36dc2`), which landed all three elements you proposed: pointer types for optional headers (`*T` via `GoTypeDef()` in `pkg/codegen/operations.go`), nil-guards on `w.Header().Set()` in the strict-server templates, and a `HeadersImplicitlyRequired` compatibility flag for users relying on the old behavior. The originally-linked issue #1301 is now fixed.

Apologies for the long wait. We really do appreciate the contribution — it pointed at the right problem and the right shape of fix.

PR #1419, opened 2024-01-06
Adopt fiber middleware template for updated GetReqHeaders() method signature

### LINKED ISSUE #1341: Error when generating a swagger with a header
When generating my swagger collection, where I use parameters in the headers, I encountered a problem. The generated code has an error. I'm using the fiber framework. Using a brute force method, I discovered that the error was also present for chi. But since the main direction is echo, I checked it too, there is no error in echo.

I have prepared a minimal version of the swagger yml with which you can reproduce the error

I placed the following code in the example-error.yml file
```yaml
openapi: 3.0.3
info:
  title: error headers
  version: 1.0.0

paths:
  /test:
    get:
      parameters:
        - in: header
          name: TESTV
          schema:
            type: string
            format: uuid
          required: true
      responses:
        '204':
          description: OK
```

then to reproduce the error you need to run the following command to generate code

```shell
oapi-codegen -generate types -o "types.gen.go" -package "main" "example-error.yml"
oapi-codegen -generate fiber -o "fiber-server.gen.go" -package "main" "example-error.yml"
```
After doing this, you will probably be able to reproduce the problem. There will be an error on line 43 in the generated file fiber-server.gen.go

```go
func (siw *ServerInterfaceWrapper) GetTest(c *fiber.Ctx) error {

	var err error

	var params GetTestParams

	headers := c.GetReqHeaders()

	// ------------- Required header parameter "TESTV" -------------
	if value, found := headers[http.CanonicalHeaderKey("TESTV")]; found {
		var TESTV openapi_types.UUID

		// An error occurs here, we are passing value ([]string type), but it is expected (string type)
		err = runtime.BindStyledParameterWithLocation("simple", false, "TESTV", runtime.ParamLocationHeader, value, &TESTV)
		if err != nil {
			return fiber.NewError(fiber.StatusBadRequest, fmt.Errorf("Invalid format for parameter TESTV: %w", err).Error())
		}

		params.TESTV = TESTV

	} else {
		err = fmt.Errorf("Header parameter TESTV is required, but not found: %w", err)
		return fiber.NewError(fiber.StatusBadRequest, err.Error())
	}

	return siw.Handler.GetTest(c, params)
}
```

I suggest running the following code to generate echo server code

```shell
oapi-codegen -generate types -o "types.gen.go" -package "main" "example-error.yml"
oapi-codegen -generate server -o "echo-server.gen.go" -package "main" "example-error.yml"
```

we get code with a similar situation, but the problem is solved, instead of a list, an element with a [0] index is passed to the function
```go
// GetTest converts echo context to params.
func (w *ServerInterfaceWrapper) GetTest(ctx echo.Context) error {
    var err error

    var params GetTestParams

    headers := ctx.Request().Header
// ------------- Required header parameter "TESTV" -------------
    if valueList, found := headers[http.CanonicalHeaderKey("TESTV")]; found {
        var TESTV openapi_types.UUID
        n := len(valueList)
        if n != 1 {
            return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Expected one value for TESTV, got %d", n))
        }

        // a similar function call, with correctly passed arguments and header values
        err = runtime.BindStyledParameterWithLocation("simple", false, "TESTV", runtime.ParamLocationHeader, valueList[0], &TESTV)
        if err != nil {
            return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter TESTV: %s", err))
        }

        params.TESTV = TESTV
    } else {
[example files.zip](https://github.com/deepmap/oapi-codegen/files/13194690/example.files.zip)

        return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Header parameter TESTV is required, but not found"))
    }

    // Invoke the callback with all the unmarshaled arguments
    err = w.Handler.GetTest(ctx, params)
    return err
}
```

I also want to thank you for developing oapi-codegen, I hope that the problem will be solved soon. Since I would like to use the fiber framework

getBolted:
- Bump fiber version to 2.52.0
- Adopted middleware template for fiber to handle new GetReqHeader() method, that has changed signature in fiber 2.50.0 (https://github.com/gofiber/fiber/releases/tag/v2.50.0)

Fixes #1341


ITheCorgi (approved):

examples/Makefile
jamietanna, line 1: @mromaszewicz can you please revert this change? This means that we'll have our tests and examples not running < Go 1.24, which means we may miss other changes

We should make sure that only the modules that need newer Go versions are guarded appropriately

PR #1423, opened 2024-01-10
add unit tests for go type gen function

sonasingh46:
This PR adds unit test for `GoTypeDef()` method.

### DIFF
--- a/pkg/codegen/schema_test.go
+++ b/pkg/codegen/schema_test.go
@@ -158,7 +158,7 @@ func TestProperty_GoTypeDef(t *testing.T) {
 			want: "*int",
 		},
 		{
-			name: "When field is readOnly and optional and and read only pointer disabled",
+			name: "When field is readOnly and optional and read only pointer disabled",
 			fields: fields{
 				GlobalStateDisableRequiredReadOnlyAsPointer: true,
 				Schema: Schema{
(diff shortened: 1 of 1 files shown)


pkg/codegen/schema_test.go
```diff
@@ -0,0 +1,202 @@
...
+			},
+			want: "*int",
+		},
+
+		// When field is write only, it will always be pointer unless pointer is
+		// skipped by setting SkipOptionalPointer flag
+		{
+			name: "When field is write only",
+			fields: fields{
+				GlobalStateDisableRequiredReadOnlyAsPointer: true,
```
jamietanna, line 179: are we missing a case for the case where this is set to `false`?

pkg/codegen/schema_test.go
```diff
@@ -0,0 +1,202 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			globalState.options.Compatibility.DisableRequiredReadOnlyAsPointer = tt.fields.GlobalStateDisableRequiredReadOnlyAsPointer
+			p := Property{
+				Schema:    tt.fields.Schema,
+				Required:  tt.fields.Required,
+				Nullable:  tt.fields.Nullable,
+				ReadOnly:  tt.fields.ReadOnly,
+				WriteOnly: tt.fields.WriteOnly,
+			}
+			assert.Equalf(t, tt.want, p.GoTypeDef(), "GoTypeDef()")
```
jamietanna, line 199:
```suggestion
			assert.Equal(t, tt.want, p.GoTypeDef())
```

jamietanna:
Couple of things to address, then LGTM

pkg/codegen/schema_test.go
sonasingh46: added...

pkg/codegen/schema_test.go
```diff
@@ -174,7 +174,7 @@ func TestProperty_GoTypeDef(t *testing.T) {
 		// When field is write only, it will always be pointer unless pointer is
 		// skipped by setting SkipOptionalPointer flag
 		{
-			name: "When field is write only",
+			name: "When field is write only and read only pointer disabled",
```
jamietanna, line 177: is the name of this test right, when the flag is `GlobalStateDisableRequiredReadOnlyAsPointer: true`? Or should we rename the tests?

sonasingh46: I did not understand what you mean. 
Are you saying rename to the following: 
```
"When field is write only and the flag is GlobalStateDisableRequiredReadOnlyAsPointer: true
```

What is incorrect here?

jamietanna: Ignore me, I'd misread

jamietanna (approved):

PR #1424, opened 2024-01-12
Simplify extensions merge in `mergeOpenapiSchemas`

Juneezee:
From the Go specification [^1]:

>  "3. If the map is nil, the number of iterations is 0."

Therefore, we don't need the `nil` checks for merging the `Extensions`. Just like how we did for `Properties`:

https://github.com/deepmap/oapi-codegen/blob/66f9bb8d73111908bb20d30bae90e65eb49a6770/pkg/codegen/merge_schemas.go#L202-L210

[^1]: https://go.dev/ref/spec#For_range

### DIFF
--- a/pkg/codegen/merge_schemas.go
+++ b/pkg/codegen/merge_schemas.go
@@ -86,19 +86,14 @@ func mergeAllOf(allOf []*openapi3.SchemaRef) (openapi3.Schema, error) {
 // all of whose fields are composed.
 func mergeOpenapiSchemas(s1, s2 openapi3.Schema, allOf bool) (openapi3.Schema, error) {
 	var result openapi3.Schema
-	if s1.Extensions != nil || s2.Extensions != nil {
-		result.Extensions = make(map[string]interface{})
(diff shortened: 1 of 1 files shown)


jamietanna:
Thanks @Juneezee! Would you mind merging the latest changes into your branch, also so CI can pick it up? I'm not sure I'll be able to as you're on your `master` branch.

[More info](https://github.com/deepmap/oapi-codegen/blob/master/CONTRIBUTING.md#before-you-raise-a-pr)

Juneezee:
@jamietanna I just rebased the latest changes. Please take a look again. Thanks :smiley: !

jamietanna (approved):
OK, so it looks like we don't have as good unit test coverage of this as I'd hoped - if I delete this whole block of code, nothing seems to change in generated code, nor do tests fail :sweat_smile: 

I'm happy saying that yes, from review of the changes, this looks good - but in the future (https://github.com/deepmap/oapi-codegen/issues/1617) we should improve this

PR #1438, opened 2024-01-24
Fixes type collision for enum values that start with _ (underscore)

ula:
Resolves an issue with type collision when enum values start with `_` (underscore) value.


jamietanna (approved):
Thank you!

jamietanna:
@ula mind addressing the build failures? Looks like some test changes + running `make generate`

jamietanna:
Hey @ula @Ulad did you still want to get this updated and merged?

(Also, are both of these accounts yours? I see the PR is raised by the former, but the commit by the latter)

PR #1457, opened 2024-02-08
Ensure docstring comment names are set correctly

ConorPKeegan:
This PR ensures that the docstring comment names are set using `x-go-name` so that they actually match the Go name.

Currently, it would generate this as shown in the test:
```
// ReferenceToRenameMe When a Schema is renamed, $ref should refer to the new name
 type ReferenceToRenameMe struct {
        // ToNewName This schema should be renamed via x-go-name when generating
        NewName NewName `json:"ToNewName"`
 }
```

Whereas, it should instead show this:
```
// ReferenceToRenameMe When a Schema is renamed, $ref should refer to the new name
 type ReferenceToRenameMe struct {
        // NewName This schema should be renamed via x-go-name when generating
        NewName NewName `json:"ToNewName"`
 }
```

### DIFF
--- a/internal/test/components/components.gen.go
+++ b/internal/test/components/components.gen.go
@@ -329,7 +329,7 @@ type OneOfVariant6 struct {
 
 // ReferenceToRenameMe When a Schema is renamed, $ref should refer to the new name
 type ReferenceToRenameMe struct {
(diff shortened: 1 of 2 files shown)


jamietanna (approved):
Thanks very much!

PR #1463, opened 2024-02-11
Add support for `net/http`-only routing

### LINKED ISSUE #1068: New Server: Generate `net/http`-only code

jamietanna:
Closes #1068.


jamietanna:
Closing in preference of https://github.com/deepmap/oapi-codegen/pull/1475

PR #1477, opened 2024-02-26
feat(echo): add middleware to handlers

GreyXor:
This allows us to set middleware only for these handlers


GreyXor:
@jamietanna gentle ping

GreyXor:
@jamietanna hello, it's a little patch, can I have your input here ?

GreyXor:
@jamietanna hello ?

Issue #1009, opened 2023-03-22
Add struct tag `query` for binding query parameters in echo framework


t-oki:
When generating models for query parameters, oapi-codegen creates a model like below, generating struct tag `form` and `json`,

```
type FindPetsParams struct {
	// Tags tags to filter by
	Tags *[]string `form:"tags,omitempty" json:"tags,omitempty"`

	// Limit maximum number of results to return
	Limit *int32 `form:"limit,omitempty" json:"limit,omitempty"`
}
```

However, in echo, query binding is done with struct tag `query`.
https://echo.labstack.com/guide/binding/#struct-tag-binding
If you're using oapi-codegen to generate echo-server as well, it won't be a problem since BindQueryParameter method will do the binding for you.
But if you're using oapi-codegen to generate only the models, it will become a problem since echo will not bind query parameters with a field with no `query` struct tag.

So, I propose to add struct tag `query`, and change the generated code like below.

```
type FindPetsParams struct {
	// Tags tags to filter by
	Tags *[]string `form:"tags,omitempty" json:"tags,omitempty" query:"tags,omitempty"`

	// Limit maximum number of results to return
	Limit *int32 `form:"limit,omitempty" json:"limit,omitempty"  query:"tags,omitempty"`
}
```

I think it's a similar issue of https://github.com/deepmap/oapi-codegen/issues/500.

mromaszewicz:
The generated echo server code already handles query binding via `BindQueryParameter`. For models-only generation, you can use `x-oapi-codegen-extra-tags` to add custom struct tags like `query`.

PR #1478, opened 2024-02-27
Add struct tag query for binding query parameters in echo framework

ilick888:
in the Echo framework, query binding is performed using the struct tag query, as detailed in Echo's documentation on struct tag binding (https://echo.labstack.com/guide/binding/#struct-tag-binding). If you are using oapi-codegen to generate an Echo server, there isn't an issue because the BindQueryParameter method handles the binding. But if oapi-codegen is only used to generate models, you'll encounter a problem since Echo won't bind query parameters to a field that lacks the query struct tag.

Therefore, I suggest adding the query struct tag. The modified code would look like this:
```
type FindPetsParams struct {
    // Tags to use as a filter
    Tags *[]string `form:"tags,omitempty" json:"tags,omitempty" query:"tags,omitempty"`

    // The maximum number of results to return
    Limit *int32 `form:"limit,omitempty" json:"limit,omitempty" query:"limit,omitempty"`
}
```


jamietanna:
Hey @ilick888 thanks very much for this, and appreciate you helping close off #1009.

Would you be able to add this as an opt-in flag? I think it's probably best to keep it so folks intentionally opt-in to it, given it's going to change a lot of generated code, mostly unnecessarily.

Or alternatively, we make it so it only applies when just generating types?

Thoughts?

jamietanna (changes requested):
See comment :smile:

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because the maintainer requested making it opt-in and the redesign was never followed up on. The templates have changed since then.

Issue #1429, opened 2024-01-19
enum not generated when used with anyof


Mo3m3n:
The types generated from the following spec sample will not generate the **enum** definition:
````
openapi: 3.0.0
paths:
  /test:
    get:
      operationId: Test
      responses:
        "200":
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/test'
components:
  schemas:
    test:
      type: object
      properties:
      anyOf:
        - properties:
            fieldA:
              type: string
        - properties:
            fieldA:
              type: string
              enum:
                - foo
                - bar
````
resulting in `undefined: Test1FieldA` with the following generated code
````
// Test defines model for test.
type Test struct {
	union json.RawMessage
}

// Test0 defines model for .
type Test0 struct {
	FieldA *string `json:"fieldA,omitempty"`
}

// Test1 defines model for .
type Test1 struct {
	FieldA *Test1FieldA `json:"fieldA,omitempty"`
}

````

I have seen that a similar [issue](https://github.com/deepmap/oapi-codegen/issues/1189) was [fixed](https://github.com/deepmap/oapi-codegen/pull/1207) before but it deals with a different usecase:
```
components:
  schemas:
    test:
      properties:
        fieldA:
          anyOf:
            - type: string
            - type: string
              enum:
                - foo
                - bar
````

jamietanna:
Related to #648

mromaszewicz:
The exact spec from the issue report was run against the current binary and the output correctly includes the Test1FieldA type definition, enum constants (Foo, Bar), and Valid() method. The root fix came via PR #1207 (commit a839952d) which addressed the anyOf+enum type propagation in generateUnion by ensuring AdditionalTypes from inline union elements are propagated upward. There is no dedicated regression test for this exact nesting pattern (anyOf at object level with enum in member properties), but the fix that landed for the related issue #1189 covered the same code path.

Closing as fixed. If you can still reproduce on a recent release, please reopen with the exact spec and version used.

PR #1522, opened 2024-03-28
fix: generate enum definitions for schemas that use anyOf

thedadams:
Fixes: https://github.com/deepmap/oapi-codegen/issues/1429


thedadams:
@jamietanna I know there are a few issues similar to this one, but this particular one was easy to address. I looked at the other related issues, but they were not as straightforward.

grongor:
Hi :) I was trying out your patch to see if it fixes my issues, but not quite. I have `anyOf` inside a response definition, and when debugging the issue it was present in the `schema.AdditionalTypes` field, but never got written out. This fixes it:

```
index a006335..2f5f7ab 100644
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -428,6 +428,10 @@ func GenerateTypeDefinitions(t *template.Template, swagger *openapi3.T, ops []Op
 			return "", fmt.Errorf("error generating Go types for component request bodies: %w", err)
 		}
 		allTypes = append(allTypes, bodyTypes...)
+
+		for _, typ := range allTypes {
+			allTypes = append(allTypes, typ.Schema.GetAdditionalTypeDefs()...)
+		}
 	}

 	// Go through all operations, and add their types to allTypes, so that we can
@@ -529,8 +533,6 @@ func GenerateTypesForSchemas(t *template.Template, schemas map[string]*openapi3.
 			TypeName: goTypeName,
 			Schema:   goSchema,
 		})
-
-		types = append(types, goSchema.GetAdditionalTypeDefs()...)
 	}
 	return types, nil
 }
```

Not sure if this is the right approach...I've never contributed here, I'm using this package for the first time and I'm also OpenAPI novice, so...yeah :laughing: 
Even with these changes it still doesn't produce working code with my schema (the discriminator field has a custom type, and the generated `From*` and `Merge*` methods try to assign a plain string), but I'm again not sure if this is even relevant, or yet another issue...
image


Let me know if you include this or if I should create a separate issue/PR. Cheers!

thedadams:
@grongor Thanks for your input here. It looks like your diff is very similar to the one I have created.

Regarding the discriminator: I haven't looked at this at all, so I am not sure what is going on there. There are a few issues around this type of problem that I found here, but this one fixes #1429 (which is the same as the specific issue I was having). I know @jamietanna was working on a fix to a similar, but different problem, too.

grongor:
Sorry for the delay. You can ignore the enum thing (it was an issue with the schema, where the discriminator was marked as optional).

As for the rest, if I understand it correctly, your changes make sure that all additional types are returned on call to `GetAdditionalTypeDefs` (recursivelly), e.g. fixing a case when a type had additional type that also had additional type (and this final type was previously not generated).
My changes fix other issue: it seems like the method `GetAdditionalTypeDefs` is not used everywhere where it should be (check the usage of the method), so in my case the types were not generated because the method `GenerateTypesForResponses` didn't call this `GetAdditionalTypeDefs`.
That's why I moved the call to `GetAdditionalTypeDefs` after all types are generated (`GenerateTypesForSchemas`, `GenerateTypesForParameters`, `GenerateTypesForResponses` and `GenerateTypesForRequestBodies`).

Here is my fork https://github.com/grongor/oapi-codegen/tree/fix-1429, and the commit specifically https://github.com/deepmap/oapi-codegen/commit/5706088b982e8bff97bd2ff15d26848d1f14fce3.

thedadams:
Are you saying that your approach fixes both issues?

Issue #1525, opened 2024-03-29
`nullable` on a `$ref` is overridden by `required` within `type: object`


brandonbloom:
This:

```yaml
components:
  schemas:
    String:
      type: string
    Thing:
      type: object
      properties:
        x:
          $ref: "#/components/schemas/String"
          nullable: true
      required:
        - x
```

Produces a structure with a non-pointer field.

brandonbloom:
Seems upstream cause is this: https://github.com/getkin/kin-openapi/issues/927

Justinfan827:
FWIW I think this is how you're supposed to do this via OpenAPI 3.0: https://stackoverflow.com/questions/40920441/how-to-specify-a-property-can-be-null-or-a-reference-with-swagger

Justinfan827:
This worked for me:

```
  object:
    allOf:
     - $ref: ./object.yaml
    nullable: true
```

When i needed the object to be nullable

PR #1526, opened 2024-03-29
Provides failing test for #1525

brandonbloom:
Working on a fix, but providing failing test case until then.


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because this was an incomplete work-in-progress test case for nullable required refs with no fix included. The code has been restructured since.

PR #1527, opened 2024-03-30
Use aliases for response type references

Crystalix007:
Uses type aliases for response schema definitions.

While using a schema like:

```yaml
paths:
  /example:
    get:
      responses:
        200:
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/OneOrAnother"

components:
  schemas:
    One:
      type: object

    Another:
      type: object

    OneOrAnother:
      oneOf:
        - $ref: "#/components/schemas/One"
        - $ref: "#/components/schemas/Another"
```

I noticed that this generated "strict" server code that looked like:

``` go
type OneOrAnother struct {
	union json.RawMessage
}

// ...

type GetExampleResponseObject interface {
	VisitGetExampleResponse(w http.ResponseWriter) error
}

type GetExampleResponseObject OneOrAnother

// ...

func (t OneOrAnother) MarshalJSON() ([]byte, error) {
	b, err := t.union.MarshalJSON()
	return b, err
}

func (t *OneOrAnother) UnmarshalJSON(b []byte) error {
	err := t.union.UnmarshalJSON(b)
	return err
}
```

However this doesn't actually work when doing the serialisation, because the type definition hides the definition of `MarshalJSON()`. Since the `union` field is private, this means the response that is actually serialised is `{}` every time.

Instead, we should just inherit that response from the referenced type definition. I.e. with `type GetExampleResponseObject = OneOrAnother`, the underlying `MarshalJSON` implementation will be picked up instead, and we can serialise the underlying union.

### DIFF
--- a/examples/petstore-expanded/strict/api/petstore-server.gen.go
+++ b/examples/petstore-expanded/strict/api/petstore-server.gen.go
@@ -343,7 +343,7 @@ type AddPetResponseObject interface {
 	VisitAddPetResponse(w http.ResponseWriter) error
 }
 
-type AddPet200JSONResponse Pet
+type AddPet200JSONResponse = Pet
 
 func (response AddPet200JSONResponse) VisitAddPetResponse(w http.ResponseWriter) error {
 	w.Header().Set("Content-Type", "application/json")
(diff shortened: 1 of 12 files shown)


mromaszewicz:
Thank you for submitting this PR, @Crystalix007 — but we need to close it because the underlying bug (strict-server response types referencing `oneOf`/`anyOf` unions silently producing `{}`) was fixed via commit `5260c720` taking a different approach: a new `HasCustomMarshalJSON()` method on `Schema` triggers explicit delegating `MarshalJSON`/`UnmarshalJSON` methods on the response type, rather than using type aliases. That handles external-ref unions and doesn't constrain the response type. Issue #970 has a regression test at `internal/test/issues/issue-970/`.

Appreciate the contribution — you correctly diagnosed the missing-method-inheritance issue.

Issue #1530, opened 2024-04-03
Use discriminator with several mappings for one schema


maks56893:
I would like to use two schemas in oneOf with discriminator. I have several enum values for each discriminator fields in schemas. So can I use discriminator like this?

```
ConfigSaveReq:
      oneOf:
      - $ref: "#/components/schemas/ConfigHttp"
      - $ref: "#/components/schemas/ConfigSsh"
      discriminator:
        propertyName: config_type
        mapping:
          ssh_server: "#/components/schemas/ConfigSsh"
          apache_server: "#/components/schemas/ConfigHttp"
          web_server: "#/components/schemas/ConfigHttp"
          another_server: "#/components/schemas/ConfigHttp"
```

Now after generating types I get switch with only two cases. Furthermore, for case with ConfigHttp variable value selecting randomly from ConfigHttp keys

mgabeler-lee-6rs:
Having this same issue, also finding that the generated code in this case is non-deterministic which adds extra pain for CI stesp trying to ensure that the generated code is up to date

benazir-sba:
Having the same issue. Would love to see it solved 🗡

gerhardwagner:
can we see a PR and a merge of this ?

PR #1534, opened 2024-04-08
fix: #1530 use discriminator with several mappings for one schema

jKiler:
I've decided I'll give it a try but as long as fixing `ValueByDiscriminator` seems pretty straightforward, it requires some more changes to support proper generation of `FromX` and `MergeX` methods.

Let's assume our `Dog` has breeds `[ beagle, poodle ]`. It will produce a code:
```go
func (t *Pet) FromDog(v Dog) error {
	v.Breed = "beagle"
	v.Breed = "poodle"
	b, err := json.Marshal(v)
	t.union = b
	return err
}
```
which seems fair if you're not planning to use it, but it compiles only if we specify our `breed` as `required`.

I am not fully aware of the intentions behind setting the mapping field in these methods, so I'm not sure if resigning from setting the field is the solution here.

Another pair of 👀 would be much appreciated. Cheers!


internal/test/issues/issue-1530/issue1530.gen.go
```diff
@@ -0,0 +1,141 @@
...
+func (t Pet) AsCat() (Cat, error) {
+	var body Cat
+	err := json.Unmarshal(t.union, &body)
+	return body, err
+}
+
+// FromCat overwrites any union data inside the Pet as the provided Cat
+func (t *Pet) FromCat(v Cat) error {
+	v.Breed = "maine_coon"
+	v.Breed = "ragdoll"
```
mgabeler-lee-6rs, line 86: This PR goes a long way to fixing the issue for me, but it doesn't address this part of the codegen

mgabeler-lee-6rs: I've submitted an enhanced version of this PR -- https://github.com/oapi-codegen/oapi-codegen/pull/1789 -- which I think should address this issue

Issue #1538, opened 2024-04-12
Add hasPrefix and hasSuffix template funcs


maks56893:
I'm trying use custom generate templates and want to add some methods for only response models. For define what is request or response I have suffixes for models. But for now I can't define it in template

maks56893:
here is [PR](https://github.com/deepmap/oapi-codegen/pull/1539)

Issue #1536, opened 2024-04-11
Add Sprig templates for code generation


deefdragon:
I'm planning on taking on #1521 to make it a bit easier to generate your own templates. 

To make writing templates easier in general, I'd also like to import [Sprig](https://github.com/Masterminds/sprig) and add it to the functions used for the templates. I've been using them in my templates for some time via a fork. Many of the added functions are VERY useful for navigating the OAPI structs and the extensions Ive added.

Wanted to discuss before I PR if there are any questions or potential issues.

jamietanna:
Thanks for the interest! I think right now, I'd prefer to avoid adding the dependency (especially as it'll pull [a number of dependencies](https://github.com/Masterminds/sprig/blob/581758eb7d96ae4d113649668fa96acc74d46e7f/go.mod#L5-L16)) and since the work off the back of #1142 I'd love to keep us as lean as possible.

What sort of template functions are you looking for? 

It is possible to modify [the existing template functions](https://github.com/deepmap/oapi-codegen/blob/80f0b978ef16e837edafb29ad8d6c9e24a044aa9/pkg/codegen/template_helpers.go#L300) to add the helpers we need?

Or to call `oapi-codegen` as a library (but needs better documenting in #1487) ?

deefdragon:
The most common funcs I've pulled in are `dig`, `haskey`, `list`,and `get` functions. Specifically I use those to work with the extensions that I have added to my openapi yaml. (`fail` is useful for validation reasons as well, tho much less necessary as compared to the other 4.)

If I were to generalize, I think the [datastructure funcs](https://github.com/Masterminds/sprig/blob/581758eb7d96ae4d113649668fa96acc74d46e7f/functions.go#L293-L309) are the ones that would provide the most given the purpose of oapi-codegen. Most of them are implementable without imports, tho the merge ones are debatable.

deefdragon:
Ive gone through the list of functions that _may_ be useful from sprig for dict/list manipulation.

- `list`, `append`, `prepend`, `first`, `rest`, `last`, `initial`, `reverse`, `uniq`, `without`, `has`, `slice`, `concat`, `chunk`
- `dict`, `get`, `set`, `unset`, `hasKey`, `pluck`, `keys`, `values`, `pick`, `omit`, `dig`

Most of the `Must` variants etc. are not that useful in this lib I think, so those can be excluded. I also excluded the merge funcs as those pulled in an external library, and I dont think people are going to require merging maps. 

[...]

PR #1539, opened 2024-04-12
fix: #1538 add hasPrefix and hasSuffix template funcs

maks56893:


deefdragon:
As a side note, these differ from the sprig template functions (see #1536 where sprig is discussed) in parameter ordering. I believe this is to allow pipe-lining easier in the templates. In sticking with sprig ordering, vs making the template functions added better match their golang counterparts, I think I lean slightly towards the latter.

Issue #1521, opened 2024-03-28
Generate code from custom templates


sagikazarmark:
I'd like to integrate models and types with a custom (in-house built) framework.

In order to do that, I'd like to generate custom marshaler/unmarshaler functions.

It would be nice if I could add arbitrary templates to the code generation process.

One thing I tried is overriding templates for a server that I don't use. To complicate things, we still use the Chi integration, so I actually need to support two servers at the same time...which is not allowed at the moment.

As a result, I'm now overriding Chi templates by copying the original ones and extending them with my custom code which is not ideal IMO.

sagikazarmark:
What I’d like to do is basically add custom templates that contain additional code (for example: mapping to internal types).

My understanding is that the current template config only allows me to override templates already defined. Is that correct?

deefdragon:
It will pull in _any_ templates provided in the user-templates object. Unfortunately, I did not set it up to have a way to call those templates. 

As such, you need to call those templates somewhere that already exists, which will require over-riding at-least one existing template.

deefdragon:
Made the PR ready for review. Could use a check on if the documentation I put in the readme makes sense @sagikazarmark

PR #1544, opened 2024-04-18
Add method of calling arbitrary user templates

deefdragon:
Adds an option to the `generate` options in the configuration yaml, allowing for defining what additional user-defined templates to call. Data passed to template is an object containing both the array of operations as passed to most other templates, and the raw openapi specification.

Resolves https://github.com/deepmap/oapi-codegen/issues/1521

This is a draft for feedback reasons (should I include anything else to pass to the template, format the config options differently etc.), as well as I still need to test locally to confirm one or two things, and don't want it merged in until Ive tested that.

### DIFF
--- a/examples/additional-templates/only-models.gen.go
+++ b/examples/additional-templates/only-models.gen.go
@@ -29,22 +29,27 @@ type Unreferenced struct {
 //:::::.\::::::::.\::::::::.\::::::::.\::::::::.\::::::::.\::::::::.\::::::::.\::::::::.\::::::::.\::::::::.\::::::::.\
 //'      `--'      `--'      `--'      `--'      `--'      `--'      `--'      `--'      `--'      `--'      `--'      `
 
-//created by inputs-example-template.tmpl
+//User templates are passed a struct that has two fields, spec and ops, described below.
 
-//User templates are passed an object that has two keys, ops and spec, described below.
+///////////
+// spec //
+/////////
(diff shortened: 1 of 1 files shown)


deefdragon:
Finally got around to doing the testing I wanted to do. Will get the merge conflicts resolved and mark as ready for review.

jamietanna (commented):
Thanks for this - mind adding an example to `examples/` to illustrate how it can be used?

deefdragon:
Added an example set that goes into a bit of detail on how to use the user-provided templates.

The one thing that I was reminded of as I was working on this is that the types array is generates in a different location, and not passed in anywhere in the ops array. Types is one of the things that is likely to be most useful in user template generation stuff, but I don't want to make this PR any larger. 

Adding the types to the provided struct would require code changes in other areas to pipe the type array up, and adding more data to the passed in object like that can be done over time.

deefdragon:
@jamietanna Is the awaiting reply tag supposed to auto-remove? And or was there anything else you would like me to change about this?

andrii-hrushetskyi:
I have a use case for this PR right now. I need to have the same tracing and validation code in every method of the `StrictServerInterface` implementation. 
I want to generate a struct, that wraps server implementation and adds that tracing and valudation

deefdragon:
I'll take a look tonight to update the Readme and see if there is anything else I need to do for the docs.

Issue #1362, opened 2023-11-26
Model generation `$ref`-utilising components not consumer friendly


percivalalb:
In larger OpenAPI specs it is useful to split up components into multiple files and use [`$ref`](https://swagger.io/docs/specification/using-ref/) to "stitch" the spec together. This doesn't quite work when generating models, full example:

**Example Spec**
1. `schemas/record.yaml`
```
type: object
required:
  - id
properties:
  id:
    type: number
```
2.1 `schemas/records.yaml`
```
type: array
items:
  $ref: './record.yaml'
```
3. `api.yaml`
```
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Ref Example
paths:
  /records:
    get:
      operationId: getRecords
      responses:
        200:
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Records'
  /record:
    get:
      operationId: getRecord
      responses:
        200:
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Record'      
components:
  schemas:
    Record:
      $ref: './schemas/record.yaml'
    Records:
      $ref: './schemas/records.yaml'
```

The go generated code doesn't recongise the array item definition is the same as the `Record` definition and so doesn't reference `Record` from `Records`. Instead it usings an anonymous type which makes it annoying to use. Ideally it should be `type Records = []Record`
```
// Record defines model for Record.
type Record struct {
	Id float32 `json:"id"`
}

// Records defines model for Records.
type Records = []struct {
	Id float32 `json:"id"`
}
```

**Alternative Approach**

Changing file 2. to .... has a different problem.
2.2 `schemas/records.yaml`
```
type: array
items:
  $ref: '../api.yaml#/components/schemas/Record'
```

```
$ go generate ./...
error generating code: error generating type definitions: error generating Go types for component schemas: error converting Schema Records to Go type: error resolving primitive type: error generating type for array: error turning reference (../api.yaml#/components/schemas/Record) into a Go type: unrecognized external reference '../api.yaml'; please provide the known import for this reference using option --import-mapping
exit status 1
doc.go:6: running "go": exit status 1
```
However the API component in the "external reference" isn't actually external. It is part of this API spec so using import-mapping doesn't make sense or actually work, as the go type is generated in the same file not in another package.

Changing 2. to ... to remove the external reference has the same problem as the original and then also isn't techincally valid OpenAPI (even though the code generation doesn't work).
2.3 `schemas/records.yaml`
```
[...]

PR #1573, opened 2024-04-22
Reconcile references to remote schemas to models from the root spec 

percivalalb:
For https://github.com/deepmap/oapi-codegen/issues/1362

OpenAPI specifications can be spread across different files for maintainability. However this is not great for code-gen which can't recognise when two schema refs refer to the same model. This results in the code-gen using duplicating anonymous structs (which can't easily be used).

I have added a patch to kin upstream (https://github.com/getkin/kin-openapi/pull/945) which identifies when a ref refers to a schema from the root #/components/schemas and makes use of the go model generated from that. I'm not sure if it accounts for every edge case, but certainly works for my case and doesn't break any of the existing example specs within this repo.

See https://github.com/deepmap/oapi-codegen/commit/f662fa7772de5d00b41d9c493303a93eab2460e5 which shows the change codegen change.

Of course the kin PR needs looking at first, and I couldn't decide how much logic should exist there vs here.


pkg/codegen/schema.go
```diff
@@ -237,7 +237,7 @@ func PropertiesEqual(a, b Property) bool {
 	return a.JsonFieldName == b.JsonFieldName && a.Schema.TypeDecl() == b.Schema.TypeDecl() && a.Required == b.Required
 }
 
-func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
+func GenerateGoSchema(sref *openapi3.SchemaRef, path []string, component bool) (Schema, error) {
```
percivalalb, line 240: `component` could do with a better name.

When it's true we are generating a go type for a schema in '#/components/schemas' the root document.

percivalalb:
The changes needed to the kin-openapi library are present in v0.126.0, I'll wait for https://github.com/oapi-codegen/oapi-codegen/pull/1689 to be merged before completing this PR

percivalalb:
This is now ready for review

mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because this was blocked on an upstream kin-openapi dependency merge that was never completed, and the codebase has moved on.

percivalalb:
Hi, the changes to the upstream dependancy were completed and merged in. Oapi-codegen is now using the required version of kin-openapi. I imagine this PR needs rebasing since it's been so long but it should still work fine.

Issue #1554, opened 2024-04-19
Validate number of entries in `x-enum-varnames`


jamietanna:
```diff
diff --git examples/extensions/xenumvarnames/api.yaml examples/extensions/xenumvarnames/api.yaml
index 5601f4a5..8bdd308d 100644
--- examples/extensions/xenumvarnames/api.yaml
+++ examples/extensions/xenumvarnames/api.yaml
@@ -16,4 +16,3 @@ components:
         - EXP
       x-enum-varnames:
         - Active
-        - Expired
diff --git examples/extensions/xenumvarnames/gen.go examples/extensions/xenumvarnames/gen.go
index 6b919353..a6137d39 100644
--- examples/extensions/xenumvarnames/gen.go
+++ examples/extensions/xenumvarnames/gen.go
@@ -5,14 +5,14 @@ package xenumvarnames
 
 // Defines values for ClientType.
 const (
-       ACT ClientType = "ACT"
-       EXP ClientType = "EXP"
+       ClientTypeACT ClientType = "ACT"
+       ClientTypeEXP ClientType = "EXP"
 )
 
 // Defines values for ClientTypeWithExtension.
 const (
-       Active  ClientTypeWithExtension = "ACT"
-       Expired ClientTypeWithExtension = "EXP"
+       ClientTypeWithExtensionActive ClientTypeWithExtension = "ACT"
+       ClientTypeWithExtensionEXP    ClientTypeWithExtension = "EXP"
 )
 
 // ClientType defines model for ClientType.
```

Issue #1487, opened 2024-03-05
Document how to perform codegen-as-a-library


jamietanna:

deefdragon:
When I start on implementing #1536, it may be a good idea to break out the documentation for user templates into a separate markdown file. Thoughts?

jamietanna:
i.e. https://github.com/grafana/grafana/blob/7077a5850e75f7a33b7f2f8cebad2e0633b37e74/pkg/codegen/generators/go_generator.go#L15

PR #1582, opened 2024-04-29
Add Dict and List sprig functions for template generation

deefdragon:
This is a starter on what functions to add from the sprig template funcs. This resolves parts of #1536, as it provides atleast a starting point. 

This is a draft for conformation that these are good. If this looks good, I will go through and add the appropriate tests as well as add documentation for the functions that are listed and the existing functions (Resolving #1487)

I think that I will add `fail` as well if there is no push to not do so. It would assist users id doing their own schema/extension validations IE #1554.


phoenisx:
Maybe we can also add support for string to integer conversion and vice-versa.

This will help a lot with customization based on status code.

Also is this PR ready, or still in draft? @deefdragon 

In case u are not getting time to wrap this up, maybe I can continue on finishing and adding more Template functions, as this seems to be a good change for better customization support for custom config templates.

phoenisx:
Also I feel it would be best to introduce strig as a dep to this project instead of us creating our own duplicate template functions.
We can always merge maps. So the following should work and be simpler compared to us maintaining the same code.

```go
import "github.com/Masterminds/sprig/v3"
// ...
maps.Copy(TemplateFunctions, sprig.FuncMap())
```

**Edit:**
I do see that we can get conflicts between duplicate keys, but feels like if that happens it's high likely that the conflicted functions are doing the same thing.

phoenisx:
For now, since I need this change in my project, I'll fork this project.

But I'd love it if these changes can get reviewed/approved 💯

phoenisx:
@jamietanna Maybe you can help here.

This PR has been open for a year now.
I can work on it in case this has become stale and isn't worked upon anymore.

Or maybe you have better ideas around this?

deefdragon:
Please feel free to take over. Unfortunately life has gotten in the way of me keeping up with the prs I have open, and my fork that just imports sprig as a dep has worked well enough for me that I've not had a push to work on this.

phoenisx:
U can close this one, if u aren't going to contribute anymore.
I'll create a separate PR, after testing 👍🏽

PR #1597, opened 2024-05-05
fix: don't allow `strict-server` to be used on its own

### LINKED ISSUE #1558: fix: return an error when `strict-server` is called but no server is used with it
As it's unable to work without a base server

jamietanna:
As it doesn't actually do anything, as it wraps an existing server i.e.
Chi.

Closes #1558.


mromaszewicz:
Thank you for contributing, and I'm very sorry for taking so long to get to this PR. At this point, the code has changed so much that it's no longer relevant because this was already implemented — configuration validation now prevents `strict-server` without a server type.

Issue #1515, opened 2024-03-25
Combine multiple errors


jamietanna:
i.e. https://github.com/deepmap/oapi-codegen/blob/a14414f04fdd2206ac394e1e45527c1fbcbfc705/internal/test/components/components.gen.go#L2380-L2385 could use `errors.Join`

PR #1598, opened 2024-05-06
fix to return combined multiple errors

eyasy1217:
closes https://github.com/deepmap/oapi-codegen/issues/1515

MarshalJSON returns combined json marshal errors in AdditionalProperties.

e.g.

Before 
`error marshaling 'unsupportedType1': json: unsupported type: chan int`
↓
After 
`error marshaling 'unsupportedType1': json: unsupported type: chan int
'unsupportedType2': json: unsupported type: func()`

(requierd go 1.20)

### DIFF
--- a/internal/test/components/components.gen.go
+++ b/internal/test/components/components.gen.go
@@ -495,15 +495,16 @@ func (a BodyWithAddPropsJSONBody) MarshalJSON() ([]byte, error) {
 		return nil, fmt.Errorf("error marshaling 'name': %w", err)
 	}
 
-	var joinedError error
+	var errs []error
 	for fieldName, field := range a.AdditionalProperties {
 		object[fieldName], err = json.Marshal(field)
 		if err != nil {
-			joinedError = errors.Join(joinedError, fmt.Errorf("'%s': %w", fieldName, err))
+			errs = append(errs, fmt.Errorf("'%s': %w", fieldName, err))
 		}
 	}
-	if joinedError != nil {
-		return nil, fmt.Errorf("error marshaling %w", joinedError)
+	err = errors.Join(errs...)
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling %w", err)
 	}
 
 	return json.Marshal(object)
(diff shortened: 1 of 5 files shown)


jamietanna (changes requested):
Thanks for this!

I was wondering if maybe instead of this, we have i.e.

```go
	var errs []error
	for fieldName, field := range a.AdditionalProperties {
		object[fieldName], err := json.Marshal(field)
		if err != nil {
			errs = append(errs, fmt.Errorf("'%s': %w", fieldName, err))
		}
	}
	err := errors.Join(errs...)
	if err != nil {
		return nil, fmt.Errorf("error marshaling %w", joinedError)
	}
```

Which then uses `errors.Join` in a slightly easier way?

eyasy1217:
> Thanks for this!
> 
> I was wondering if maybe instead of this, we have i.e.
> 
> ```go
> 	var errs []error
> 	for fieldName, field := range a.AdditionalProperties {
> 		object[fieldName], err := json.Marshal(field)
> 		if err != nil {
> 			errs = append(errs, fmt.Errorf("'%s': %w", fieldName, err))
> 		}
> 	}
> 	err := errors.Join(errs...)
> 	if err != nil {
> 		return nil, fmt.Errorf("error marshaling %w", joinedError)
> 	}
> ```
> 
> Which then uses `errors.Join` in a slightly easier way?

It looks to me like this is doing something similar to [the internal code of errors.Join](https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/errors/join.go;l=32-42)
🤔

jamietanna:
Yep, but with an array rather than joining in the loop will probably be a little more efficient, and only does the joining once

eyasy1217:
That's right.
🚀 https://github.com/deepmap/oapi-codegen/pull/1598/commits/52f6cdc8be554cfd98baaa71c7deb16e475fa301

gaiaz-iusipov:
I think the same behavior can be achieved as follows:

```go
var errs error
for fieldName, field := range a.AdditionalProperties {
	object[fieldName], err = json.Marshal(field)
	if err != nil {
		errs = errors.Join(errs, fmt.Errorf("json marshal %q: %w", fieldName, err))
	}
}
if errs != nil {
	return nil, errs
}
```

PR #1600, opened 2024-05-06
Strictgin error handler

actatum:
Strict gin server doesn't currently use an error handler function. This adds the error handler to the strict gin handler


pkg/codegen/templates/strict/strict-gin.tmpl
```diff
@@ -34,19 +47,18 @@ type strictHandler struct {
                 {{if .IsJSON }}
                     var body {{$opid}}{{.NameTag}}RequestBody
                     if err := ctx.ShouldBindJSON(&body); err != nil {
-                        ctx.Status(http.StatusBadRequest)
-                        ctx.Error(err)
+                        sh.options.ErrorHandlerFunc(ctx, err)
```
jamietanna, line 50: should this not still be sending a `http.StatusBadRequest`?

pkg/codegen/templates/strict/strict-gin.tmpl
```diff
@@ -55,15 +67,15 @@ type strictHandler struct {
...
                         return
                     }
                     {{else -}}
                     if _, params, err := mime.ParseMediaType(ctx.Request.Header.Get("Content-Type")); err != nil {
-                        ctx.Error(err)
+                        sh.options.ErrorHandlerFunc(ctx, err)
                         return
                     } else if boundary := params["boundary"]; boundary == "" {
-                        ctx.Error(http.ErrMissingBoundary)
+                        sh.options.ErrorHandlerFunc(ctx, err)
```
jamietanna, line 78: should this not still be sending a `http.ErrMissingBoundary`?

pkg/codegen/templates/strict/strict-gin.tmpl
```diff
@@ -93,14 +105,13 @@ type strictHandler struct {
         response, err := handler(ctx, request)
 
         if err != nil {
-            ctx.Error(err)
-            ctx.Status(http.StatusInternalServerError)
+            sh.options.ErrorHandlerFunc(ctx, err)
```
jamietanna, line 108: should this not still be sending a `http.StatusInternalServerError`?

pkg/codegen/templates/strict/strict-gin.tmpl
```diff
@@ -93,14 +105,13 @@ type strictHandler struct {
...
         } else if validResponse, ok := response.({{$opid | ucFirst}}ResponseObject); ok {
             if err := validResponse.Visit{{$opid}}Response(ctx.Writer); err != nil {
-                ctx.Error(err)
+                sh.options.ErrorHandlerFunc(ctx, err)
             }
         } else if response != nil {
-            ctx.Error(fmt.Errorf("unexpected response type: %T", response))
+            sh.options.ErrorHandlerFunc(ctx, err)
```
jamietanna, line 114:
```suggestion
            sh.options.ErrorHandlerFunc(ctx, fmt.Errorf("unexpected response type: %T", response))
```

pkg/codegen/templates/strict/strict-gin.tmpl
mromaszewicz: It absolutely should.

pkg/codegen/templates/strict/strict-gin.tmpl
mromaszewicz: I'd argue that in this case, it actually shouldn't. Let the user's error handler decide the response code.

pkg/codegen/templates/strict/strict-gin.tmpl
mromaszewicz: Yep, you're right. `err` is guaranteed to be nil here.

pkg/codegen/templates/strict/strict-gin.tmpl
mromaszewicz: I need to think about this a little bit. The user should be able to decide what `Status` to set on `ctx`, but this is difficult to figure out just based on `err`.

mromaszewicz:
Ok, so I know what to do here. I'm going to have three categories of error callbacks.
1) `RequestErrorHandlerFunc` which is invoked when there is an error unmarshaling a request, this will default to HTTP/400
2) `HandlerErrorFunc`, which is probably the one most interesting to users. This is called if the handler code returns an error.
3) `ResponseErrorHandlerFunc`, which gets called if the response serialization vistor encounters an error.

The problem with having a single callback, is that you have no idea where it was called from, so you don't have good context on how to handle that error. With these three error handler types, I think we can retain current behavior (which this PR breaks without @jamietanna 's changes) and allow for much easier error customization.

mromaszewicz:
@greptileai

mromaszewicz:
@greptileai

Issue #1584, opened 2024-04-30
http strict request bodies should be optional by default


emilien-puget:
request bodies should be optional by default
see https://spec.openapis.org/oas/v3.1.0#fixed-fields-10 or https://swagger.io/docs/specification/describing-request-body/

but as we can see here https://github.com/deepmap/oapi-codegen/blob/621b44626ac9a1aa6c9fdda97324a58569227281/pkg/codegen/templates/strict/strict-http.tmpl#L50, we check that a content type is set only when multiple request bodies are defined, so if our one and only request body is not declared as required, this generated code will return an error , i made this custom template to handle that issue
https://gist.github.com/emilien-puget/6d0e0af27c72f4cf331991cccfeb96f4#file-strict-http-tmpl

i also made a custom template
https://gist.github.com/emilien-puget/6d0e0af27c72f4cf331991cccfeb96f4#file-strict-interface-tmpl-L15 to handle required body
https://github.com/deepmap/oapi-codegen/blob/621b44626ac9a1aa6c9fdda97324a58569227281/pkg/codegen/templates/strict/strict-interface.tmpl#L15 as seen here, when the body is set as required we could drop the pointer on the body.

would you be interested having a PR that change the behavior of the generated code to handle such cases ? we could have an operation-output option like "optional-request-bodies" to switch to this new behavior.

jamietanna:
Interesting, thank you! Yes we'd appreciate a PR for that, and agreed that having it behind `optional-request-bodies` would be best :raised_hands:

emilien-puget:
started to do stuff and things regarding this issue [there](https://github.com/deepmap/oapi-codegen/pull/1608)

emilien-puget:
@jamietanna have you had the chance to take a look at my draft?

DavyJ0nes:
Any update on this? This is really useful to have as I got caught out by it today :)

emilien-puget:
> Any update on this? This is really useful to have as I got caught out by it today :) 

Feel free to discuss on the pull request to challenge the implementation

mromaszewicz:
This is a comment from an automated bug scrub, after making a significant number of fixes to the code generation.

We reproduced this end-to-end against the current `main`: an absent optional request body no longer produces an error.

**Spec used** (`spec.yaml`) — the request body is `required: false`:
```yaml
openapi: "3.0.0"
info:
  title: Test API
  version: "1.0.0"
paths:
  /items:
    post:
      operationId: createItem
[...]

PR #1608, opened 2024-05-11
Handles whether a request body is set as required or not in the specification.

emilien-puget:
As briefly discussed [there](https://github.com/deepmap/oapi-codegen/issues/1584)

The current behavior of the code generation is a mix between always required and always optional bodies.
However the open api specification is pretty clear about that requirement of body
see https://spec.openapis.org/oas/v3.1.0#fixed-fields-10 or https://swagger.io/docs/specification/describing-request-body/

> Request bodies are optional by default. 

this PR aims to bring the code generation closer to the intent of the open api specification.

using this spec as an example
```yaml
  /requiredrequests/:
    post:
      requestBody:
        required: true
        content:
....
  /requests/:
    post:
      requestBody:
        required: false
```

### type declaration

The issue here is that a body is always optional, using a pointer to convey this intent.

#### current behavior:

```go
type PostRequiredrequestsRequestObject struct {
	Body *PostRequiredrequestsJSONRequestBody
}
```

a required type will always be a pointer.

#### suggested behavior:

```go
type PostRequiredrequestsRequestObject struct {
	Body PostRequiredrequestsJSONRequestBody
}
```

we drop the ptr since the body was marqued as required in the open api specification.

as seen here, the type generation is considering a request body as being always optional.

### strict function handler

The issue here is that a body is always required.

#### current behavior
```go
var request PostRequestsRequestObject

var body PostRequestsJSONRequestBody
if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
	sh.options.RequestErrorHandlerFunc(w, r, fmt.Errorf("can't decode JSON body: %w", err))
	return
}
request.Body = &body
```

#### suggested behavior:
```go
var request PostRequestsRequestObject

if strings.HasPrefix(r.Header.Get("Content-Type"), "application/json") {

	var body PostRequestsJSONRequestBody
	if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
		sh.options.RequestErrorHandlerFunc(w, r, fmt.Errorf("can't decode JSON body: %w", err))
		return
	}
	request.JSONBody = &body
}

```

since the api specification states that the body is not required,  which is the default value for the required attribute, we need to check that the body is given to us before trying to decode it.

this may not be a great way to check that the body is there, since a user of the api could set the content type header and not setting a body.

I think this proposed change should be at one point the default behavior of the code generation as it is the default behavior of the open api specification

wdyt ?

- [ ] implements the same behavior for all the strict server
- [ ] body check detection using content type header may be a bit si bit simplistic

### DIFF
--- a/pkg/codegen/codegen_test.go
+++ b/pkg/codegen/codegen_test.go
@@ -166,6 +166,108 @@ func TestGoTypeImport(t *testing.T) {
 	}
 }
 
+func TestRequiredBody(t *testing.T) {
+	packageName := "api"
+	opts := Configuration{
+		PackageName: packageName,
+		Generate: GenerateOptions{
+			Models:        true,
+			StdHTTPServer: true,
+			Strict:        true,
+		},
+	}
+	spec := "test_specs/required-body.yaml"
+	swagger, err := util.LoadSwagger(spec)
+	require.NoError(t, err)
+
+	// Run our code generation:
+	code, err := Generate(swagger, opts)
+	assert.NoError(t, err)
+	assert.NotEmpty(t, code)
+
+	// Check that we have valid (formattable) code:
+	_, err = format.Source([]byte(code))
+	assert.NoError(t, err)
+
+	assert.Contains(t, code, `type PostRequestsRequestObject struct {
+	Body *PostRequestsJSONRequestBody
+}`)
+
+	assert.Contains(t, code, `type PostRequiredrequestsRequestObject struct {
+	Body *PostRequiredrequestsJSONRequestBody
+}`)
+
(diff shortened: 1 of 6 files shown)


mromaszewicz:
Thank you for submitting this PR, @emilien-puget — but we need to close it because the primary bug (strict server unconditionally erroring when no body was sent for optional request bodies) was fixed by #2222 (commit `99615d04`). All five strict-server templates (stdhttp, echo, gin, fiber, iris) now correctly check for `io.EOF` and treat an absent body as nil for optional bodies while still requiring one for `required: true` operations. The secondary enhancement you proposed (non-pointer struct fields for required bodies) wasn't adopted, but the core bug is resolved.

Thanks for the contribution.

Issue #1454, opened 2024-02-06
Is there a way to apply "x-go-type-skip-optional-pointer": true to the entire oapi code generation process?


Tanz0rz:
I do not have the ability to edit the swagger files. What is the best way to achieve this behavior?

PR #1694, opened 2024-07-10
feat(output-options): add `prefer-skip-optional-pointer` to default to skipping optional pointers

aksdb:
I had a similar requirement as in issue #1454 and discussion #1460 a while a go and implemented the following solution.

This is likely not of the necessary quality for merging, but I would still like to open this PR as a discussion about the implementation. If I missed anything obvious that could cause problems down the line, I will certainly adjust as necessary.


jamietanna:
Thanks for this - will try and look at it once the backlog of items for v2.4.0 are down!

jamietanna:
FYI this would also link in nicely with https://github.com/oapi-codegen/oapi-codegen/issues/1561 (as two alternative ways)

Napam:
Kind of from the side, but I personally would like something similar to this. More specifically, I would LOVE to have a "pointers-for-values-with-defaults" such that I can set it to "false" (or something like that).  Scattering the `x-go-type-skip-optional-pointer` is very tedious for some of my optonal types that have defaults. I use kin-openapi for validation, and it just sets all the defaults values for me.

mminklet:
Just curious if any progress has been made on getting this merged? Would love to use it

pkg/codegen/schema.go
```diff
@@ -263,11 +264,19 @@ func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
 			return Schema{}, fmt.Errorf("error turning reference (%s) into a Go type: %s",
 				sref.Ref, err)
 		}
+		skipOptionalPointer := globalState.options.OutputOptions.PreferSkipOptionalPointer
+		if schema.Type == nil || slices.Contains(*schema.Type, openapi3.TypeObject) {
+			// For object references pointers are the only way to denote optional, in contrast
+			// to other types where the zero-value can be used for that.
```
jamietanna, line 270: Where does this get triggered? I can't seem to work out where this may occur :thinking:

jamietanna:
Related: https://github.com/oapi-codegen/oapi-codegen/discussions/1860

jamietanna (approved):
Thanks a lot for this :purple_heart:

pkg/codegen/templates/chi/chi-middleware.tmpl
jamietanna, line 1: I should re-validate that everything is covered as part of this change (elsewhere)

pkg/codegen/schema.go
aksdb: Hmm, I actually don't remember anymore. If I spot anything later, I'll add it here (or in a discussion, or issue or whatever would be appropriate then :smile: )

PR #1695, opened 2024-07-10
Support x-go-type-skip-optional-pointer in client query

TelpeNight:
Hello.
We've tried to use x-go-type-skip-optional-pointer with optional slices in our code base. But generated client code became invalid. Original template makes dereference of pure slice.
I've added an extra check for this case, for query parameters. Other param locations may need the same check.
Note, that supporting optional not-nilable types (ex int) with x-go-type-skip-optional-pointer  requires extra work. As `if params.{{.GoName}} != nil` check will not compile. But zero value semantic is not well defined in this case.

### DIFF
--- a/pkg/codegen/templates/client.tmpl
+++ b/pkg/codegen/templates/client.tmpl
@@ -210,7 +210,7 @@ func New{{$opid}}Request{{if .HasBody}}WithBody{{end}}(server string{{genParamAr
 
             {{end}}
             {{if .IsStyled}}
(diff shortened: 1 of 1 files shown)


mromaszewicz:
Thank you for submitting this PR, @TelpeNight — but we need to close it because the root cause was addressed more completely in commits `5e440d55` (mirroring the parameter-level extension in `DescribeParameters`) and `c966b373` (using `RequiresNilCheck` consistently for header/cookie params), with regression fixtures under `internal/test/issues/issue-2329/` and `issue-2238/`. A current-binary repro confirms `x-go-type-skip-optional-pointer` on optional slice query params now generates correct code without the spurious `*params.X` dereference.

Thanks for surfacing this — it was a real bug.

PR #1697, opened 2024-07-12
feat(extension): add `x-oapi-codegen-only-honour-go-name`

### LINKED ISSUE #1638: Unexported fields not working anymore since v2.2.0
Since version v2.2.0 all unexported fields (renamed via `x-go-name`) are now exported even though they have been named lowercase. Reproducible with the example `xgoname` in this repository.
Would need a critical fix for this.

With version 2.1.0 everything works out.

EDIT: Seems to affect fields only, not direct types.
image

jamietanna:
Although originally to fix #1638, which isn't a very recommended
approach as it may lead to a confusing behaviour for users, we can still
add support.

However, we should hide this behind both a global Compatibility Option,
to ensure that it's clearly opted-in, as well as adding a per-field
level to ensure that only the right fields are overridden.

Closes #1638.


lukasbash:
Got an ETA of a merge?

jamietanna:
(Merging as I had a bit of time this evening, not necessarily related to the bump)

PR #1700, opened 2024-07-13
fix(extensions): use the `x-order` from next to `$ref`, if present

paulmach:
This is part 2 of the fix for https://github.com/oapi-codegen/oapi-codegen/issues/1611.

The issues is that something like the below doesn't work.
```
type: object
properties:
 start_port:
   $ref: "#/components/schemas/port"
   x-order: 1
 end_port:
   $ref: "#/components/schemas/port"
   x-order: 2
```

First problem was the kin-openapi does not actually capture these values from the yaml/json. This fixed in https://github.com/getkin/kin-openapi/pull/901, release in v0.126.0 and pulled in here in https://github.com/oapi-codegen/oapi-codegen/pull/1689

This is the second part which actually considers this value in the sort and officially fixes https://github.com/oapi-codegen/oapi-codegen/issues/1611.  It first checks for "next to $ref" then if there is an `x-order` in the schema referenced.

### DIFF
--- a/internal/test/extensions/x-order/spec.yaml
+++ b/internal/test/extensions/x-order/spec.yaml
@@ -21,9 +21,9 @@ components:
         - start
         - end
       properties:
-        start:
-          $ref: '#/components/schemas/Port'
-          x-order: 1
         end:
           $ref: '#/components/schemas/Port'
           x-order: 2
+        start:
+          $ref: '#/components/schemas/Port'
+          x-order: 1
--- a/internal/test/extensions/x-order/spec.yaml
+++ b/internal/test/extensions/x-order/spec.yaml
@@ -13,3 +13,17 @@ components:
           type: string
           format: date
           x-order: 1
+    Port:
+      type: integer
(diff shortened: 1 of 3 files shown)


AndrewCopeland:
I made a comment on the original issue that was closed https://github.com/oapi-codegen/oapi-codegen/issues/1611.

Seems like we are waiting on this PR to be merged to resolve this bug.

internal/test/extensions/x-order/spec.yaml
```diff
@@ -13,3 +13,17 @@ components:
...
           x-order: 1
+    Port:
+      type: integer
+    PortInterval:
+      type: object
+      required:
+        - start
+        - end
+      properties:
+        start:
```
jamietanna, line 24: Let's flip these so it's clear the order is coming from the `x-order`

jamietanna (approved):
Thank you for this! Added in an additional case, and a couple of comments :)

Issue #1682, opened 2024-07-03
Gin has no error types


Guillembonet:
I see that other servers like chi or std return different error types such as `RequiredHeaderError` or `InvalidParamFormatError`, but gin returns errors without a specific type like `fmt.Errorf("Query argument page is required, but not found")`. This makes errors handling code more uglier and more prone to bugs.

Is this for some reason desired behavior?

jamietanna:
Nope, this looks like a gap in the templates - Gin was an earlier implementation than Chi and Gorilla - from memory - so could be we introduced it as a pattern after the fact. Contributions welcome!

Guillembonet:
Okay, I might try to do a PR for this next week! Thanks!

PR #1706, opened 2024-07-20
use error types for gin server

Guillembonet:
Closes: #1682 

I copied the logic from the other servers, though I'm not sure what is the use of having things like:

```
err = fmt.Errorf("Error unescaping cookie parameter '{{.ParamName}}'")
siw.ErrorHandler(c, &UnescapedCookieParamError{ParamName: "{{.ParamName}}", Err: err}, http.StatusBadRequest)
```

Which overwrites the unsmarshalling error and is not printed when using the Error() function. I see that something similar happens with `RequiredHeaderError`.

For now, I copied to match the other servers since maybe I overlooked some reason why it's there, but I can change it if it will not cause problems.

### DIFF
--- a/examples/minimal-server/gin/api/ping.gen.go
+++ b/examples/minimal-server/gin/api/ping.gen.go
@@ -4,6 +4,8 @@
 package api
 
 import (
+	"fmt"
+
 	"github.com/gin-gonic/gin"
 )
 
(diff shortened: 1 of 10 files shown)


mromaszewicz (commented):
This is a good change, it's how the code should have looked like in the first place, however, when people used the old code, they probably did some kind of string comparison to find the errors, and this change will break that.

What do you think about this one @jamietanna ? It's a change which changes behavior. Flagify it?

There's a little more complexity here too. People sometimes generate their servers in parts, by constraining the generated code on tags or something. If you have two Gin sub-parts in the same directory, you'll have type redefinition for your errors. We migth have to make a package to hold these error types.

Guillembonet:
> This is a good change, it's how the code should have looked like in the first place, however, when people used the old code, they probably did some kind of string comparison to find the errors, and this change will break that.
> 
> What do you think about this one @jamietanna ? It's a change which changes behavior. Flagify it?

About this, it's something I also thought as it will happen to me as well. At least in my case, a way to solve it is to have this custom errors return the same message as they did before. What I'm not sure about this approach is that, then, the errors will be slightly different than the ones from the other servers, which is odd.

> 
> There's a little more complexity here too. People sometimes generate their servers in parts, by constraining the generated code on tags or something. If you have two Gin sub-parts in the same directory, you'll have type redefinition for your errors. We migth have to make a package to hold these error types.

This I did not consider, but I think it would be useful to know how is to being done in the other servers. It looked to me that the same would happen there.

Guillembonet:
Hi @jamietanna, this has been stale for a while waiting for your input. It would be nice if you could find time to give it a look! Thanks!

PR #1708, opened 2024-07-23
Fix 867 by rejecting unsupported OpenAPI specs

mromaszewicz:


jamietanna (commented):
Is this not going to be a breaking change? Given we've had to update our test files

Does this impact cases we've got partial specs (ie for external refs)?

I'd seen that when I started with https://github.com/oapi-codegen/oapi-codegen/pull/713

mromaszewicz:
> Is this not going to be a breaking change? Given we've had to update our test files
> 
> Does this impact cases we've got partial specs (ie for external refs)?
> 
> I'd seen that when I started with #713

Yes. It is indeed a breaking change, and one I wanted your opinion on.

A more compatible way would be to shim ourselves into the loading path in Kin somehow, and only complain if there's a known incompatible version present, and accept specs without versions. Let me have a second go at it...

cmd/oapi-codegen/oapi-codegen.go
```diff
@@ -281,20 +282,41 @@ func main() {
 		return
 	}
 
-	swagger, err := util.LoadSwagger(flag.Arg(0))
+	spec, err := util.LoadOpenAPI(flag.Arg(0))
 	if err != nil {
 		errExit("error loading swagger spec in %s\n: %s\n", flag.Arg(0), err)
```
ccoVeille, line 287: Shouldn't be this?

```suggestion
 		errExit("error loading OpenAPI spec in %s\n: %s\n", flag.Arg(0), err)
```

jamietanna: It's unfortunately a wider issue https://github.com/oapi-codegen/oapi-codegen/issues/672

mromaszewicz:
I'm going to do this differently, without affecting the loading of broken specs.

Issue #672, opened 2022-07-12
Replace use of `swagger` with OpenAPI


jamietanna:
There are a few places where we're saying Swagger instead of OpenAPI.

As Swagger has been retrospectively named OpenAPI 2.0, we should try and migrate all references accordingly.

We'll need to work around this to mark the old methods as `Deprecated:` and delegate to the new methods, so we can slowly phase them out of use.

I've made a start on a branch, but thought I'd get an issue raised for visibility.

zhoub:
+1

PR #1709, opened 2024-07-23
Fix #867 by intercepting spec loading

### LINKED ISSUE #867: unexpected reference depth
Hi guys, 

I am getting the following error 

`error generating code: error creating operation definitions: error generating response definitions: error turning reference (#/paths/~1me~1/put/responses/400) into a Go type: unexpected reference depth: 6 for ref: #/paths/~1me~1/put/responses/400 local: true`

I see that there is a validation in the [source](https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/utils.go) code against strictly defined depth, but is there a specific reason for that?

mromaszewicz:
This change adds a shim into the Kin loader which attempts to find an `openapi` or `swagger` version key in the spec, and returns an error if it's an unsupported version. in the case where the version can't be detected, we do nothing and behave as previously, failing in the parsing step.

### DIFF
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -281,7 +281,7 @@ func main() {
 		return
 	}
 
-	swagger, err := util.LoadSwagger(flag.Arg(0))
+	swagger, err := util.LoadOpenAPI(flag.Arg(0))
 	if err != nil {
 		errExit("error loading swagger spec in %s\n: %s\n", flag.Arg(0), err)
 	}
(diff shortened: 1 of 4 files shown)


jamietanna:
If we're doing some renaming here could we do it all (under https://github.com/oapi-codegen/oapi-codegen/issues/672) first as a separate PR?

mromaszewicz:
> If we're doing some renaming here could we do it all (under #672) first as a separate PR?

I didn't want to touch the old function in case someone is using it, so I marked it deprecated and made a new one. This isn't part of some renaming effort.

mromaszewicz:
Ok, I undid the function renaming and after some testing, fixed some issues which would have been regressions. It works as expected now on all kinds of specs.

PR #1717, opened 2024-08-06
Group interfaces

### LINKED ISSUE #1513: Allow generating multiple
Via https://github.com/deepmap/oapi-codegen/issues/249#issuecomment-1517859944 and (TODO: add other issue references) we should make it possible to decompose a large API across multiple types.

This would be opt-in, and maybe configurable as to whether to:

- Provide an `interface` per `operationId`
- Provide an `interface` per set of tags

This would then allow type embedding to produce a single `ServerInterface` that consumes each of these smaller interfaces

jorge-plaza:
Closes #1513 

I've just implemented it for the `std-http` code generation, give me some thoughts about it and I could implement other server frameworks.


jamietanna:
Thanks, will try and have a look once we've got down the https://github.com/oapi-codegen/oapi-codegen/milestone/10 backlog

ziflex:
Hey @jamietanna , any updates on this PR? The feature would be very useful!

anthops:
This would be super useful with echo. Keen to try this out though!

LarsStegman:
@jamietanna  this would be really great to have!

Issue #1718, opened 2024-08-09
stdhttp: Receive a mux as interface instead of *http.ServeMux


sonalys:
I'm currently implementing open-telemetry, and I want to have the route pattern as part of the operation name.
I would be able to get the method and route pattern by injecting a custom `ServeMux` during the handler initialization.

Following the open-closed principle, you should accept interfaces and return structs. Would it be possible to accept a mux interface instead of the `*http.ServeMux` struct?

I volunteer myself for implementing it if approved.

jamietanna:
Would we see this as a breaking change? If so, it's something we can control via a flag.

It sounds reasonable though!

sonalys:
It's not a breaking change. You would simply replace the struct with the expectations of what the struct should implement

jamietanna:
Cool! Yeah we're happy to accept a PR for IT, hopefully can squeeze into the next release (ETA unknown)

sonalys:
@jamietanna I opened a Pull Request at #1720

PR #1720, opened 2024-08-09
feat(std-http): accept `Mux` interface instead of `*http.ServeMux`

sonalys:
# Description
As described on Issue #1718, I'm implementing receiving Mux as interface instead of *http.ServeMux.
Following the open-closed principle, it will allow the library to be more customizable for the developers.

### DIFF
--- a/pkg/codegen/templates/stdhttp/std-http-handler.tmpl
+++ b/pkg/codegen/templates/stdhttp/std-http-handler.tmpl
@@ -3,21 +3,27 @@ func Handler(si ServerInterface) http.Handler {
   return HandlerWithOptions(si, StdHTTPServerOptions{})
 }
 
+// ServeMux is an abstraction of http.ServeMux.
+type ServeMux interface {
+	HandleFunc(pattern string, handler func(http.ResponseWriter, *http.Request))
+	ServeHTTP(w http.ResponseWriter, r *http.Request)
+}
+
 type StdHTTPServerOptions struct {
     BaseURL          string
-    BaseRouter       *http.ServeMux
+    BaseRouter       ServeMux
     Middlewares      []MiddlewareFunc
     ErrorHandlerFunc func(w http.ResponseWriter, r *http.Request, err error)
 }
 
 // HandlerFromMux creates http.Handler with routing matching OpenAPI spec based on the provided mux.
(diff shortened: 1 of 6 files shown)


pkg/codegen/templates/stdhttp/std-http-handler.tmpl
```diff
@@ -3,21 +3,26 @@ func Handler(si ServerInterface) http.Handler {
   return HandlerWithOptions(si, StdHTTPServerOptions{})
 }
 
+type Mux interface {
```
ccoVeille, line 6: You should explain what you wrote in 1718 here. I mean the reason why you define this interface

You should also mention *http.ServeMux implements it

ccoVeille, line 6: Maybe ServeMux would be easier to understand

sonalys: > I'm currently implementing open-telemetry, and I want to have the route pattern as part of the operation name.
I would be able to get the method and route pattern by injecting a custom ServeMux during the handler initialization.

ccoVeille: I know, I meant to write this in the comment of ServeMux interface

But the comment you added is OK, it could have been better, but let's say it's fine for now

ccoVeille (approved):

pkg/codegen/templates/stdhttp/std-http-handler.tmpl
```diff
@@ -3,21 +3,27 @@ func Handler(si ServerInterface) http.Handler {
   return HandlerWithOptions(si, StdHTTPServerOptions{})
 }
 
+// ServeMux is an abstraction of http.ServeMux.
+type ServeMux interface {
```
jamietanna, line 7: Does this need to be exported? Probably no big deal it is, but thought I'd ask

jamietanna (commented):
Looks good, thanks! 1 comment then :shipit:

pkg/codegen/templates/stdhttp/std-http-handler.tmpl
sonalys: It’s good since you use it in the method’s arguments.

jamietanna (approved):
Thank you!

PR #1722, opened 2024-08-12
Generate schema properties in the original order

diamondburned:
This commit temporarily switches to a fork of kin-openapi that provides
the original ordering of the Properties object. This provides a way for
this tool to generate the properties in the original order as they were
written in the YAML file.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/oapi-codegen/oapi-codegen/v2
 
 go 1.20
 
+replace github.com/getkin/kin-openapi => github.com/diamondburned/kin-openapi v0.92.1-0.20240812092412-a5b8ae983de2
+
 require (
 	github.com/getkin/kin-openapi v0.126.0
 	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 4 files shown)


diamondburned:
https://github.com/oapi-codegen/oapi-codegen/pull/1388 will essentially replace this PR with a few simple tweaks -_-

See https://github.com/pb33f/libopenapi/pull/205

mromaszewicz:
Property ordering is cosmetic, and it doesn't affect behavior, so it's better to stay on the official release of kin-openapi than to use someone else's fork.

Why is it important to preserve property order in generated code that you won't be editing?

diamondburned:
> Property ordering is cosmetic, and it doesn't affect behavior, so it's better to stay on the official release of kin-openapi than to use someone else's fork.

Right. There's a PR for this to kin-openapi, which is why this PR is a draft. See https://github.com/getkin/kin-openapi/pull/1003.

> Why is it important to preserve property order in generated code that you won't be editing?

Because the Go types are used for documentation when writing Go code. It is what you see first when you Go to Definition, not an external OpenAPI documentation.

iremsha:
> Property ordering is cosmetic, and it doesn't affect behavior, so it's better to stay on the official release of kin-openapi than to use someone else's fork.
> 
> Why is it important to preserve property order in generated code that you won't be editing?

The order for routes on Fiber is **NOT cosmetic**.

mromaszewicz:
Thank you for submitting this PR, @diamondburned — but we need to close it because the approach (depending on a fork of kin-openapi) is something we'd rather not take on, and the underlying use case is now served by the `x-order` extension that landed in #1190 (Jan 2024). Specs can opt into explicit ordering with `x-order` annotations on individual properties; without it, fields are sorted alphabetically for determinism. There's an integration test at `internal/test/extensions/x-order/spec.yaml` showing the pattern.

Appreciate you flagging this and prototyping a path forward — `x-order` ended up being the shape we went with.

PR #1730, opened 2024-08-18
WIP: Respect default schema value

### LINKED ISSUE #1703: Support defaulting query parameter values when not supplied, and default is set.
Right now, using the `default` parameter in openAPI specs does functionally nothing in the generator.

We could improve the strict generator by setting the query parameters to the default values set, and not have a pointer to them in the code, marking them as optional.

percivalalb:
Closes https://github.com/oapi-codegen/oapi-codegen/issues/1703

Requires https://github.com/oapi-codegen/runtime/pull/49


zak905:
Hi @percivalalb, just wanted to enquire about the status of this feature, is there any help needed here ?

percivalalb:
Sorry, I didn't get very far. You you probably guessed I'm not picking this back up.

PR #1738, opened 2024-08-25
Properly merge discriminator when combining schema's using allOf

alexbakker:
Previously, when combining two schema's using allOf, the discriminator would not carry over to the resulting merged schema.

### DIFF
--- a/pkg/codegen/merge_schemas.go
+++ b/pkg/codegen/merge_schemas.go
@@ -222,9 +222,19 @@ func mergeOpenapiSchemas(s1, s2 openapi3.Schema, allOf bool) (openapi3.Schema, e
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


alexbakker:
@jamietanna What's the status here?

mromaszewicz:
Thank you for submitting this PR, @alexbakker — but we need to close it because the same fix (propagating the discriminator field when merging `allOf` schemas) landed via #667 (commit `05503ded`, merged March 2026). `pkg/codegen/merge_schemas.go:388-397` now propagates `s1.Discriminator` or `s2.Discriminator` to `result.Discriminator` in `allOf` context, with test coverage in `merge_schemas_test.go`.

Thanks for the contribution.

PR #1740, opened 2024-08-27
Allow typeNamePrefix to be capable of resolving `>` and `<`

ansoni:
Hello,
   I have a spec that utilizes property keynames with:

-  `=`
- `>`
- `>=`

`=` correctly resolves to Equals, but `>` and `>=` both resolve to an empty string.

This PR adds support to resolve `>` and `<`


ccoVeille:
Is there any unit tests you could update or add?

ccoVeille (approved):

jamietanna (approved):
Thank you!

PR #1745, opened 2024-09-05
fix: type hint for discriminator types

### LINKED ISSUE #1668: Union tmpl generates wrong code when using non string type
```yaml
openapi: "3.0.0"
info:
  version: 0.0.1
  title: Test
  description: Example
servers:
  - url: /v1

paths:
  /stuff:
    post:
      description: Create stuff
      operationId: createStuff
      security:
        - Bearer: []
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/CreateStuff"
      responses:
        "201":
          description: created stuff
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Stuff"
components:
  schemas:
    CreateStuff:
      oneOf:
        - $ref: "#/components/schemas/CreateStuffBla"
        - $ref: "#/components/schemas/CreateStuffNope"
      discriminator:
        propertyName: which
        mapping:
          bla: "#/components/schemas/CreateStuffBla"
          nope: "#/components/schemas/CreateStuffNope"
    CreateStuffProperties:
      type: object
      required:
        - which
      properties:
        which:
          $ref: "#/components/schemas/WhichStuff"
    CreateStuffBla:
      allOf:
        - $ref: "#/components/schemas/CreateStuffProperties"
        - type: object
          properties:
            settings:
              $ref: "#/components/schemas/StuffBla"
    CreateStuffNope:
      allOf:
        - $ref: "#/components/schemas/CreateStuffProperties"
        - type: object
          properties:
            settings:
              $ref: "#/components/schemas/StuffNope"
    Stuff:
      type: object
      required:
        - which
      properties:
        which:
          $ref: "#/components/schemas/WhichStuff"
    WhichStuff:
      type: string
      nullable: true
      enum:
        [
          bla,
          nope,
        ]
      default: bla
    StuffBla:
      type: object
      properties:
        something:
          type: string
    StuffNope:
      type: object
      properties:
        nothing:
          type: string
```

generated code:
```
// AsCreateStuffBla returns the union data inside the CreateStuff as a CreateStuffBla
func (t CreateStuff) AsCreateStuffBla() (CreateStuffBla, error) {
	var body CreateStuffBla
	err := json.Unmarshal(t.union, &body)
	return body, err
}

// FromCreateStuffBla overwrites any union data inside the CreateStuff as the provided CreateStuffBla
func (t *CreateStuff) FromCreateStuffBla(v CreateStuffBla) error {
	v.Which = "bla"
	b, err := json.Marshal(v)
	t.union = b
	return err
}

// MergeCreateStuffBla performs a merge with any union data inside the CreateStuff, using the provided CreateStuffBla
func (t *CreateStuff) MergeCreateStuffBla(v CreateStuffBla) error {
	v.Which = "bla"
	b, err := json.Marshal(v)
	if err != nil {
		return err
	}

	merged, err := runtime.JSONMerge(t.union, b)
	t.union = merged
	return err
}
```

One of the errors:
```
'"bla"' (type string) cannot be represented by the type *WhichStuff
```

`v.Which = "bla"` (etc.) doesn't work because Which is of type `*WhichStuff` and not a string. In our case it's even a type within another package (i.e. using `x-go-type`). 

Besides fixing this issue, I wonder if we could also make the generation of those `From<Name>` and `As<Name>` and `Merge<Name>` optional. We don't use them at all and don't plan to use them either.

mweibel:
fixes #1668

❓ I'd need help writing the appropriate test code. I haven't found the best way to make sure this works and continues to works in the future yet and would appreciate some help. I tried adding a check (using [parser.ParseFile](https://pkg.go.dev/go/parser@go1.23.0#ParseFile)) in `codegen/codegen_test.go` to make sure the generated code is parseable but wasn't yet able to make it work.

### DIFF
--- a/internal/test/components/components.gen.go
+++ b/internal/test/components/components.gen.go
@@ -1312,7 +1312,6 @@ func (t OneOfObject13) AsOneOfVariant1() (OneOfVariant1, error) {
 // FromOneOfVariant1 overwrites any union data inside the OneOfObject13 as the provided OneOfVariant1
 func (t *OneOfObject13) FromOneOfVariant1(v OneOfVariant1) error {
 	t.Type = "v1"
-
 	b, err := json.Marshal(v)
 	t.union = b
 	return err
@@ -1321,7 +1320,6 @@ func (t *OneOfObject13) FromOneOfVariant1(v OneOfVariant1) error {
 // MergeOneOfVariant1 performs a merge with any union data inside the OneOfObject13, using the provided OneOfVariant1
 func (t *OneOfObject13) MergeOneOfVariant1(v OneOfVariant1) error {
 	t.Type = "v1"
-
 	b, err := json.Marshal(v)
 	if err != nil {
 		return err
@@ -1342,7 +1340,6 @@ func (t OneOfObject13) AsOneOfVariant6() (OneOfVariant6, error) {
 // FromOneOfVariant6 overwrites any union data inside the OneOfObject13 as the provided OneOfVariant6
 func (t *OneOfObject13) FromOneOfVariant6(v OneOfVariant6) error {
 	t.Type = "v6"
-
 	b, err := json.Marshal(v)
 	t.union = b
 	return err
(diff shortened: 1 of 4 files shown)


mweibel:
@jamietanna would you be able to review this? and maybe point me in the direction on writing tests for this particular case

jamietanna:
I'll try and have a look once I've got the v2.4.0 release backlog out ☺

PR #1751, opened 2024-09-09
fix: getting cookie in fiber middleware

### LINKED ISSUE #1752: Cookies will never be used when generating a fiber server
The version of oapi-codegen you're using - v2.1.0

The YAML configuration file you're using:
```yaml
package: openapi
generate:
  fiber-server: true
  models: true
output: server_interface_gen.go
```
The OpenAPI spec you're using:
```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Minimal ping API server
paths:
  /ping:
    get:
      parameters:
        - in: cookie
          name: some_name
          required: true
          schema:
            type: string
      responses:
        '200':
          description: pet response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Pong'
components:
  schemas:
    # base types
    Pong:
      type: object
      required:
        - ping
      properties:
        ping:
          type: string
          example: pong
```
What problem you're seeing:

**Here you can see that the parameter received from the cookie will simply not be used.**
```go
// GetPing operation middleware
func (siw *ServerInterfaceWrapper) GetPing(c *fiber.Ctx) error {

	var err error

	var params GetPingParams

	var cookie string

	if cookie = c.Cookies("some_name"); cookie == "" {
		var value string
		err = runtime.BindStyledParameterWithOptions("simple", "some_name", cookie, &value, runtime.BindStyledParameterOptions{Explode: true, Required: true})
		if err != nil {
			return fiber.NewError(fiber.StatusBadRequest, fmt.Errorf("Invalid format for parameter some_name: %w", err).Error())
		}
		params.SomeName = value

	} else {
		err = fmt.Errorf("Query argument some_name is required, but not found")
		return fiber.NewError(fiber.StatusBadRequest, err.Error())
	}

	return siw.Handler.GetPing(c, params)
}
```
What the expected behaviour is:

I expect the value obtained from the cookie to be passed to the request handler.

What version of Go you're using:

go1.22.6 darwin/arm64

nikitaRykhlov:
fixes: #1752 

❓I haven't found a better way to write tests other than copying parameter tests, but I haven't done that yet. Please let me know what else I can do, I'll be happy to help.

### DIFF
--- a/pkg/codegen/templates/fiber/fiber-middleware.tmpl
+++ b/pkg/codegen/templates/fiber/fiber-middleware.tmpl
@@ -123,7 +123,7 @@ func (siw *ServerInterfaceWrapper) {{$opid}}(c *fiber.Ctx) error {
     {{range .CookieParams}}
       var cookie string
 
-      if cookie = c.Cookies("{{.ParamName}}"); cookie == "" {
+      if cookie = c.Cookies("{{.ParamName}}"); cookie != "" {
 
       {{- if .IsPassThrough}}
         params.{{.GoName}} = {{if not .Required}}&{{end}}cookie


mromaszewicz:
Thank you for submitting this PR, @nikitaRykhlov — but we need to close it because it has been superseded by #2307 (commit `c2095464`), which performed a comprehensive parameter-binding audit across all server templates and explicitly noted "Obsoletes: #1751". The current `fiber-middleware.tmpl` at line 143 has the corrected `if cookie \!= ""` condition, ensuring cookies are properly bound when present.

Thanks for catching the inverted condition.

PR #1763, opened 2024-09-18
Don't swallow the error

perbu:
I made this fix a long time ago when debugging some bad yaml but forgot to contribute it back. The fix is simple, just wrap the error so it can bubble up.


perbu:
heh. an identical patch was already applied. the PR just became empty.

sorry about the noise.

Issue #1277, opened 2023-09-20
Array in content issue


ShouheiNishi:
If the schema is existed in content in operation object and describes array of object with additionalProperties, code generation is failed.

Test data is in https://github.com/ShouheiNishi/oapi-codegen/commit/81494c12c605e380ea880d231767a6fe39c4ba1a.

```
$ git status
HEAD detached at 81494c1
nothing to commit, working tree clean
$ cd internal/test/
$ go generate -v ./issues/content-array/...
issues/content-array/doc.go
error generating code: error formatting Go code // Package array provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/deepmap/oapi-codegen version v0.0.0-00010101000000-000000000000 DO NOT EDIT.
package array

import (
        "bytes"
        "compress/gzip"
        "context"
        "encoding/base64"
        "encoding/json"
        "encoding/xml"
        "errors"
        "fmt"
        "gopkg.in/yaml.v2"
        "io"
        "os"
        "net/http"
        "net/url"
        "path"
        "strings"
        "time"

        "github.com/oapi-codegen/runtime"
        strictecho "github.com/oapi-codegen/runtime/strictmiddleware/echo"
        strictgin "github.com/oapi-codegen/runtime/strictmiddleware/gin"
        strictiris "github.com/oapi-codegen/runtime/strictmiddleware/iris"
        strictnethttp "github.com/oapi-codegen/runtime/strictmiddleware/nethttp"
        openapi_types "github.com/oapi-codegen/runtime/types"
        "github.com/getkin/kin-openapi/openapi3"
        "github.com/go-chi/chi/v5"
        "github.com/labstack/echo/v4"
        "github.com/gin-gonic/gin"
        "github.com/gofiber/fiber/v2"
        "github.com/kataras/iris/v12"
        "github.com/kataras/iris/v12/core/router"
        "github.com/gorilla/mux"
)












// RequestEditorFn  is the function signature for the RequestEditor callback function
type RequestEditorFn func(ctx context.Context, req *http.Request) error

// Doer performs HTTP requests.
//
// The standard http.Client implements this interface.
type HttpRequestDoer interface {
        Do(req *http.Request) (*http.Response, error)
}

// Client which conforms to the OpenAPI3 specification for this service.
type Client struct {
        // The endpoint of the server conforming to this interface, with scheme,
        // https://api.deepmap.com for example. This can contain a path relative
[...]

PR #1772, opened 2024-09-20
chore(tests): add test case for #1277

jamietanna:
Although this now appears to be fixed, we should add a test case for it.


mromaszewicz:
Closing in favor of #1930 (commit `6a413495`), which both fixed the root cause in `strict-interface.tmpl` (ranging over `.Schema.AdditionalTypes`) and added a more complete test at `internal/test/issues/issue-1277/`.

PR #1780, opened 2024-09-24
feat(client): Add `Bytes()` to `ClientWithResponses` responses

grongor:
Fixes https://github.com/oapi-codegen/oapi-codegen/issues/240


jamietanna (approved):
Thanks for helping close this super old issue :face_with_peeking_eye:

PR #1781, opened 2024-09-24
fix(import-mapping): don't produce unused variable when one mapping

### LINKED ISSUE #1779: pathPrefix not used when using self import mapping
The template added in #1774 breaks when there is one import mapping of value `"-"`:

```go
    {{ if .ImportMapping }}
    pathPrefix := path.Dir(pathToFile)
    {{ end }}
    {{ range $key, $value := .ImportMapping }}
    {{ range $key, $value := .ImportMapping }}{{- if ne $value.Path "-"}}
    for rawPath, rawFunc := range {{ $value.Name }}.PathToRawSpec(path.Join(pathPrefix, "{{ $key }}")) {
        if _, ok := res[rawPath]; ok {
            // it is not possible to compare functions in golang, so always overwrite the old value
        }
        res[rawPath] = rawFunc
    }
    {{- end }}
    {{- end }}{{- end }}
    return res
```

Results in this generated code:

```go
// Constructs a synthetic filesystem for resolving external references when loading openapi specifications.
func PathToRawSpec(pathToFile string) map[string]func() ([]byte, error) {
	res := make(map[string]func() ([]byte, error))
	if len(pathToFile) > 0 {
		res[pathToFile] = rawSpec
	}

	pathPrefix := path.Dir(pathToFile)

	return res
}
```
And this error
```
go build
gen.go:366:2: declared and not used: pathPrefix
```

ottaviohartman:
Fixes #1779 .

I chose to inline the `.Dir()` function since it is simplest.

### DIFF
--- a/internal/test/externalref/externalref.gen.go
+++ b/internal/test/externalref/externalref.gen.go
@@ -76,21 +76,19 @@ func PathToRawSpec(pathToFile string) map[string]func() ([]byte, error) {
 		res[pathToFile] = rawSpec
 	}
 
-	pathPrefix := path.Dir(pathToFile)
-
-	for rawPath, rawFunc := range externalRef0.PathToRawSpec(path.Join(pathPrefix, "./packageA/spec.yaml")) {
+	for rawPath, rawFunc := range externalRef0.PathToRawSpec(path.Join(path.Dir(pathToFile), "./packageA/spec.yaml")) {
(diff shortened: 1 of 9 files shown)


jamietanna:
Looks good! We'll possibly also want to make sure this has a test case covering it - I can add it when I'm back (as I need to see where'd make sense)

jamietanna (approved):
Thanks!

PR #1784, opened 2024-09-28
build: remove patch version pinning on `go` directive

jamietanna:
Originally added in f77bd7b1b84a7f215eb43d4cecc1ba10927326ef, we wanted
to pin the version to ensure that we were targeting the latest version
of Go 1.21.

However, there's no real value here, and as noted in [0] it can be
causing issues for some users.

There's nothing explicitly in Go 1.21.x that we need to pin to, just
that we need support for Go 1.21, so we can use Speakeasy's
`openapi-overlay` library.

Therefore we can relax the patch-version requirement, which will
alleviate issues for consumers.

[0]: https://github.com/osbuild/image-builder/pull/1352

### DIFF
--- a/examples/go.mod
+++ b/examples/go.mod
@@ -1,6 +1,6 @@
 module github.com/oapi-codegen/oapi-codegen/v2/examples
 
-go 1.21.13
+go 1.21.0
 
 replace github.com/oapi-codegen/oapi-codegen/v2 => ../
 
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/oapi-codegen/oapi-codegen/v2
 
-go 1.21.13
+go 1.21.0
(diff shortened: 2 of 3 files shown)


lzap:
Thanks a bunch, I really appreciate this.

To give some context for readers of this, most project maintainers do not take enterprise into account where the pace is a tiny bit slower. Of course, major security upgrades are fast but minors are not rolled out that often. In my case, ubi9 golang toolset was upgraded a day after our conversation and it is [now on 1.21.13](https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690?architecture=amd64&image=66f13dd0fee16f12881896dc&container-tabs=overview) so it was about two months late. Our pipelines are strictly ubi-based.

PR #1797, opened 2024-10-10
Added named routes for fiber and echo

bi122:


jamietanna (changes requested):
Thanks!

As per https://github.com/oapi-codegen/oapi-codegen/blob/main/CONTRIBUTING.md#when-may-we-not-change-things we'd prefer that this is an opt-in feature, so we can avoid a lot of generated code changes.

This will allow us to have our users opt-in for the change as they'd like to

Issue #1726, opened 2024-08-14
Generated client is not working when there are more than one API's with the colon in the API path


soodpr:
There seems to be a bug if the API spec contains more than one API's with colon in the API Path, it always return the response of 1st API. 
I have put an example in this repo here: https://github.com/soodpr/oapi-codegen-bug

If we run both server and client code in the above example, the expected output from the client main program is that it first calls the /pets:validate and then /pets:generate API but in reality, its always end up calling /pets:validate API twice.

Let me know if you require any further details on this.

cosban:
soodpr's example is based on echo, but this issue is also present with gin. It seems that the code generator needs to update its `SwaggerUriToGinUri` and `SwaggerUriToEchoUri` methods to include escaping the literal colon characters.

I believe such an update would not take very long to implement and would be willing to submit an PR for it.

https://github.com/labstack/echo/issues/2046
https://github.com/gin-gonic/gin/issues/1432

cosban:
In the meantime, you should be able to work around the issue by manually escaping these paths yourself in the spec.

holyfuchs:
the issue is also present in fiber.
Escaping the paths in the same way fixes it as well.

mromaszewicz:
Revisiting this old bug.

the problem is that we pass the strings, verbatim, to router registration, and they treat them as a parameter delimiter.

We need to detect the `:` and escape it if it doesn't correspond to a parameter.

PR #1815, opened 2024-10-28
feat(#1726): add literal colon support for gin and echo

cosban:
Closes https://github.com/oapi-codegen/oapi-codegen/issues/1726


cosban:
I see the test failures and will get to them in a bit.

cosban:
see https://github.com/oapi-codegen/oapi-codegen/pull/1879

PR #1817, opened 2024-10-29
fix(strict-fiber): return error to fiber server error hanlder directly

MichaelMa-mkk:
This should resolve https://github.com/oapi-codegen/oapi-codegen/issues/1507


hogesako:
@jamietanna 
Is there anything that might prevent the merge from going through?

PR #1388, opened 2023-12-18
Migrate to libopenapi

yozel:
This pull request is a complete migration to libopenapi. Almost the tests and examples generates the same code, except embedded swaggerSpec files. While semantically its the same, due to change of order of object keys in the Json, it generates a different gzipped base64.

There is only one tests that fails, which is `internal/test/issues/issue-removed-external-ref`. And I believe it's related to this: https://github.com/pb33f/libopenapi/issues/219. I'm following/contributing to the thread to solve the issue.

The generated code continues to use kin-api since validation logic depends on middleware repos, which still uses kin-api.

Test results:
```
$ go test ./...
?   	github.com/deepmap/oapi-codegen/v2/pkg/ecdsafile	[no test files]
?   	github.com/deepmap/oapi-codegen/v2/pkg/securityprovider	[no test files]
ok  	github.com/deepmap/oapi-codegen/v2/cmd/oapi-codegen	0.733s
ok  	github.com/deepmap/oapi-codegen/v2/pkg/codegen	6.210s
ok  	github.com/deepmap/oapi-codegen/v2/pkg/util	(cached)
```

The MR also includes regenerated `*.gen.go` files, so you can compare the difference.

```
$ cd /Users/yozel/Projects/oapi-codegen/internal/test
$ go generate -v ./...
all_of/doc.go
all_of/v1/openapi.gen.go
all_of/v2/openapi.gen.go
any_of/param/doc.go
any_of/param/param.gen.go
any_of/param/param_test.go
client/client.gen.go
client/client_test.go
client/doc.go
components/components.gen.go
components/components_test.go
components/doc.go
externalref/doc.go
externalref/externalref.gen.go
externalref/imports_test.go
externalref/packageA/doc.go
externalref/packageA/externalref.gen.go
externalref/packageB/doc.go
externalref/packageB/externalref.gen.go
issues/issue-1087/api.gen.go
issues/issue-1087/doc.go
issues/issue-1087/deps/deps.gen.go
issues/issue-1087/deps/doc.go
issues/issue-1093/doc.go
issues/issue-1093/issue_test.go
issues/issue-1093/api/child/child.gen.go
issues/issue-1093/api/parent/parent.gen.go
issues/issue-1127/api.gen.go
issues/issue-1127/doc.go
issues/issue-1168/api.gen.go
issues/issue-1168/doc.go
issues/issue-1180/doc.go
issues/issue-1180/issue.gen.go
issues/issue-1180/issue_test.go
[...]

PR #1820, opened 2024-11-03
Organise code in packages

kgrama:
Hi, I would like to contribute to the codebase, but found it hard to follow with everything in one package. This is my first attempt at a restructure.  The starting point for any restructure is the global state var in codegen, not moving it to its own package causes a lot of cyclic dependencies


jamietanna (changes requested):
Thanks for the contribution!

Unfortunately, this is something that's unlikely to merge right now, as the package reshuffle looks like it may be a breaking change for anyone consuming `oapi-codegen` as a library (opposed to as a CLI)

jamietanna (changes requested):
And although we appreciate that this may improve understanding + being able to contribute, this particular change (with how it may be breaking and to make it non-breaking we need to add extra pieces to manage it similar to #1388) isn't something we'll be prioritising, but thanks for your work on it!

Hopefully this has given you a chance to understand the codebase more 🤞🏽

PR #1822, opened 2024-11-04
Handle leading underscores in property names

tobio:
I'm working with a spec that has something like:

```
      properties:
        _version:
          type: string
        version:
          type: int
```

Now surely there's some better naming they could have used whilst designing the response, but that's not in my control. As it is, oapi-codegen generates a duplicate `Version` field for both of these properties. 

This PR updates the name prefix to add an `Underscore` prefix in this case, de-duplicating the field names.

### DIFF
--- a/internal/test/issues/issue-illegal_enum_names/issue.gen.go
+++ b/internal/test/issues/issue-illegal_enum_names/issue.gen.go
@@ -22,16 +22,16 @@ import (
 
 // Defines values for Bar.
 const (
-	BarBar     Bar = "Bar"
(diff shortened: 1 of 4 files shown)


jamietanna:
Thanks for the fix, @tobio! Mind running `make generate`, and committing the changes?

jamietanna:
NP! For these ones, we generally don't try and make them backwards-compatible, so I'm happy merging as-is :grinning:

jamietanna (approved):

PR #1825, opened 2024-11-07
feat: Allow resolving references after overlay application

jgraeger:
When working with overlays, oapi-codegen doesn't support external references, even though they are supported for the same file without overlays.

This PR adds a new setting `resolve-refs` to the OverlayOutput configuration. If enabled, the external references are resolved as for a normal file (against the filepath of the spec).

### DIFF
--- a/configuration-schema.json
+++ b/configuration-schema.json
@@ -209,6 +209,11 @@
               "type": "boolean",
               "description": "Strict defines whether the Overlay should be applied in a strict way, highlighting any actions that will not take any effect. This can, however, lead to more work when testing new actions in an Overlay, so can be turned off with this setting.",
               "default": true
+            },
+            "resolve-refs": {
+              "type": "boolean",
+              "description": "ResolveRefs enables reference resolving relative to the base path of the input file after the overlay is applied.",
+              "default": "false"
             }
           },
           "required": [
--- a/cmd/oapi-codegen/oapi-codegen.go
+++ b/cmd/oapi-codegen/oapi-codegen.go
@@ -290,6 +290,9 @@ func main() {
 	if opts.OutputOptions.Overlay.Strict != nil {
 		overlayOpts.Strict = *opts.OutputOptions.Overlay.Strict
 	}
+	if opts.OutputOptions.Overlay.ResolveRefs != nil {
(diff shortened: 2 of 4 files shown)


jamietanna:
Interesting - would you mind adding an example of this to `internal/test`?

I've not hit this before, but would be good to see what we're missing

configuration-schema.json
```diff
@@ -209,6 +209,11 @@
               "type": "boolean",
               "description": "Strict defines whether the Overlay should be applied in a strict way, highlighting any actions that will not take any effect. This can, however, lead to more work when testing new actions in an Overlay, so can be turned off with this setting.",
               "default": true
+            },
+            "resolve-refs": {
+              "type": "boolean",
+              "description": "ResolveRefs enables reference resolving relative to the base path of the input file after the overlay is applied.",
+              "default": "false"
```
jamietanna, line 216: I appreciate you've done the work for wiring this in - I'd say this should be default behaviour

```suggestion
              "default": "true"
```

(I'd argue we could also not make it configurable at all, right now)

jgraeger:
@jamietanna Thank you for the swift review! I agree and removed config flag and enabled it by default. I also added an example to `internal/test` for overlay.

Apparently the issue only happens if the spec is not in the directory oapi-codegen is invoked from.  Therefore i have a subdirectory `spec/`. Running `go generate` after setting [ResolveRefs](https://github.com/oapi-codegen/oapi-codegen/blob/7c3f8c39bbfbb9017ef9c180c5202aa8883dfba0/cmd/oapi-codegen/oapi-codegen.go#L286) to false reproduces the error.

jgraeger:
@jamietanna Is the example sufficient or do you need more information? :)

jamietanna:
Thanks very much for this @jgraeger - this was a good catch, and I'm classing this as a bug, not a feature 😅

I've pushed a couple of tweaks - will wait for CI then merge

Main tweaks:

- no longer making it opt-in, and instead default behaviour
- moving the test to a different directory - you weren't expected to know this, and this is more my personal preference, what you'd done was absolutely great - thank you!

jamietanna (approved):

pixelsoccupied:
Tried this fix on my side and it works well! Is there any ETA for this release @jamietanna? Thanks!

PR #1830, opened 2024-11-10
Resolve 1743 issue, translate * as ... for std/http

hiddenmarten:


hiddenmarten:
I do understand that it's breaking changes, I will need some take to make it work via feature flag.

jamietanna:
This still looks like something that'd be good to introduce :+1:

PR #1833, opened 2024-11-13
Expose raw OpenAPI spec

jedrivisser:
When generating code with `generate.embedded-spec: true`, the openapi spec is embedded in the generated code but there is no nice way to access the raw spec.

The raw spec is useful to have access to if you want to host it for something like swagger-ui to use.

This is especially useful if you are using an [OpenAPI Overlay](https://github.com/oapi-codegen/oapi-codegen?tab=readme-ov-file#modifying-the-input-openapi-specification) to change the original spec and you want to host the modified spec.

This PR:
- Exposes the raw spec
- Regenerates existing example code using `make generate`
- Adds an example of using the exposed raw spec with swagger-ui


cosban:
the raw spec is already made available via the generated `GetSwagger` function when `generate.embedded-spec` is set to true.

Here's an example in this codebase of it already being generated: https://github.com/oapi-codegen/oapi-codegen/blob/main/examples/petstore-expanded/gin/api/petstore-server.gen.go#L252

You just need to marshal it to which ever encoding you would like to use.

For gin, this looks like the following:


```golang
        engine.Handle(
                http.MethodGet, apiPath+"/swagger.json", func(c *gin.Context) {
			if spec, err := generatedAPI.GetSwagger(); err != nil {
				c.JSON(http.StatusInternalServerError, err)
			} else {
				c.JSON(http.StatusOK, spec)
			}
		},
	)
```

jedrivisser:
It feels a bit strange to fist unmarshal the RawSpec and then marshal again
instead of just using it directly, but I guess it does of some other
advantages like allowing you to edit it a bit

jedrivisser:
Thanks for the response!

jamietanna:
When I get back to this:

- [ ] We'll want to split this into two - an example of using Swagger UI, an example of i.e. https://github.com/oapi-codegen/oapi-codegen/issues/197#issuecomment-1231880839
- [ ] We'll want to make this so it doesn't come in as a breaking change - I know of a number of codebases that use the unexported field within the package
- [ ] We'll need to look at the implications of exposing this - could someone modify the underlying slice?

Issue #1824, opened 2024-11-07
InvalidParamFormatError exposes internal Error information to client


leomfn:
If a query parameter has the wrong format, the server responds with a Status 400 (Bad Request) error. The error message contains internal server information, that should not be exposed to the client.

E. g., when a query parameter 'count is defined as

```yaml
paths:
  /test:
    get:
      parameters:
        - name: count
          in: query
          schema:
            type: integer
```

the generated code/the oapi-codegen [runtime tries to convert the supplied query parameter to an integer](https://github.com/oapi-codegen/runtime/blob/912bc9f39f1e8fd175f1f39c48b4adf5ceca39f1/bindstring.go#L64). If that fails, the error message from `strconv.ParseInt` is finally propagated to `ErrorHandlerFunc` (maybe [this](https://github.com/oapi-codegen/oapi-codegen/blob/18c9887189108ed0ebb10df5603cbc73a9a78c81/pkg/codegen/templates/stdhttp/std-http-middleware.tmpl#L82) is the relevant code?) which exposes the error message to the client.

If "a" is supplied as the 'count' query parameter, the error message reads

> Invalid format for parameter count: error binding string parameter: strconv.ParseInt: parsing "a": invalid syntax

I think, such information **should not reach the client**. "Invalid format for parameter count" is helpful, the rest of the message contains information that might be **considered sensitive** and is **hardly helpful to the client**.

As a workaround, I defined a custom ErrorHandlerFunc that I'm passing to the StdHTTPServerOptions, however, I this should not be a long-term solution:

```go
func SanitzedErrorHandler(w http.ResponseWriter, r *http.Request, err error) {
	switch err := err.(type) {
	case *InvalidParamFormatError:
		http.Error(w, fmt.Sprintf("Invalid format for parameter %s", err.ParamName), http.StatusBadRequest)
	default:
		http.Error(w, err.Error(), http.StatusBadRequest)
	}
}
```

Issue #1701, opened 2024-07-13
StrictMiddleware: Custom Error Handler has no access to modified context


sonalys:
I am using a middleware for injecting context values, but when dealing with the error handler, I don't have access to the modified request context with the injected values.

I suggest that the current error handling implementation should be part of the middleware chain.

```go
// Generated code at gen.go

// GetNotifications operation middleware
func (sh *strictHandler) GetNotifications(w http.ResponseWriter, r *http.Request) {
	var request GetNotificationsRequestObject

	handler := func(ctx context.Context, w http.ResponseWriter, r *http.Request, request interface{}) (interface{}, error) {
		return sh.ssi.GetNotifications(ctx, request.(GetNotificationsRequestObject))
	}
	for _, middleware := range sh.middlewares {
		handler = middleware(handler, "GetNotifications")
	}

	response, err := handler(r.Context(), w, r, request)

	if err != nil {
		sh.options.ResponseErrorHandlerFunc(w, r, err) // <-- Here we are not re-utilizing the middleware chain ctx and request.
	} else if validResponse, ok := response.(GetNotificationsResponseObject); ok {
		if err := validResponse.VisitGetNotificationsResponse(w); err != nil {
			sh.options.ResponseErrorHandlerFunc(w, r, err)
		}
	} else if response != nil {
		sh.options.ResponseErrorHandlerFunc(w, r, fmt.Errorf("unexpected response type: %T", response))
	}
}
```

What I'm doing to circumvent this limitation is:

```go
func ErrorHandler(f nethttp.StrictHTTPHandlerFunc, operationID string) nethttp.StrictHTTPHandlerFunc {
	return func(ctx context.Context, w http.ResponseWriter, r *http.Request, request interface{}) (response interface{}, err error) {
		resp, err := f(ctx, w, r, request)
		if err == nil {
			return resp, nil
		}
		// Custom error handling logic here
		return nil, nil
}
```

I would appreciate this change, thanks for your work!

sonalys:
@jamietanna If you could also take a look into this issue.

If you wish, I can also work on this issue.

sonalys:
I can propose a pull request solving this issue. Without breaking changes. This would avoid a lot of workarounds for tracing and telemetry

sonalys:
> https://github.com/oapi-codegen/oapi-codegen/pull/1838#issuecomment-2487595623 reads as this issue stems from substantial architectural issues

[...]

PR #1838, opened 2024-11-19
Proposal: Move request parsing inside the handler

sonalys:
# Summary

Current implementation has several flaws, some of them are critical from my perspective:

* Generated handlers are exposing implementation details, since they do not pass through any Middlewares or authentication
* Middleware context is not propagated to the handling logic, preventing open telemetry and tracing
* Request parameter parsing errors are not going through the specified error handler
* Panics on the request handling flow would not be recovered from recover Middlewares

Issues: #1701 #1824

## Approach

I wrap everything under the handler function, avoiding unnecessary type assertions, respecting the context chain, authentication and recover middlewares, if given.

Here I implemented some breaking changes on the middleware signature, not accepting the request anymore, and not returning the response, only an error.

## Further Details

* I did not implement any new tests, but if the PR gets motion, I can implement tests for the current issues as needed
* I also did not implement this solution for any other generator, only for stdhttp
* It might also be nice to change the error handler signature, to receive the context, respecting the context chain, tracing and so on

### DIFF
--- a/pkg/codegen/templates/strict/strict-http.tmpl
+++ b/pkg/codegen/templates/strict/strict-http.tmpl
@@ -1,5 +1,5 @@
-type StrictHandlerFunc = strictnethttp.StrictHTTPHandlerFunc
-type StrictMiddlewareFunc = strictnethttp.StrictHTTPMiddlewareFunc
+type StrictHandlerFunc = func(ctx context.Context, w http.ResponseWriter, r *http.Request) error
+type StrictMiddlewareFunc = func(f StrictHandlerFunc, operationID string) StrictHandlerFunc
 
 type StrictHTTPServerOptions struct {
     RequestErrorHandlerFunc  func(w http.ResponseWriter, r *http.Request, err error)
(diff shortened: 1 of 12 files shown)


pkg/codegen/templates/strict/strict-http.tmpl
```diff
@@ -1,5 +1,5 @@
-type StrictHandlerFunc = strictnethttp.StrictHTTPHandlerFunc
-type StrictMiddlewareFunc = strictnethttp.StrictHTTPMiddlewareFunc
+type StrictHandlerFunc = func(ctx context.Context, w http.ResponseWriter, r *http.Request) error
```
sonalys, line 1: I wouldn't open a PR for the core library for now. So I implemented the types, which are a breaking change, here instead.
If this gets motion I can open a PR for the core library.

sonalys:
I realized some part is still missing, I will re-open when I have everything done

sonalys:
@jamietanna There are some major concerns with OAPI-Codegen:

* Strict mode has middlewares, but needs to be used as Handler, Handler also has middlewares. They stack differently when using parameter validation, since they never "reach" the strict middlewares, authentication tracing and so on. Being a vulnerability in my opinion.
* You provide context to strict mode, even though request already has a Context, leaving the developer with 2 options for context.
* Possible panics on the OAPI Codegen layer are not wrapped by any Recover middleware

It tried working on it yesterday, and it really seems it would need to go some breaking changes to reach a satisfactory level.
Where would be the correct place to discuss these things?
If you are inclined to, I can still work on a proposal for a better API for the strict stdhttp mode

PR #1842, opened 2024-11-21
fix(codegen): Recursively propagate external references when merging

### LINKED ISSUE #1841: External references are not propagated recursively when merging "allOf" fields
Consider these two specifications:
```yaml
#a.yaml
openapi: 3.0.0                
components:
  schemas:
    Merged:
      allOf:
        - $ref: '#/components/schemas/Example'
        - $ref: 'b.yaml#/components/schemas/Merge'
    Example:
      type: object
      properties:
        a:
          type: string
        c:
          $ref: 'b.yaml#/components/schemas/Element'
```
```yaml
#b.yaml
openapi: 3.0.0
components:
  schemas:
    Merge:
      type: object
      properties:
        b:
          type: array
          items:
            $ref: "#/components/schemas/Element"
    Element:
      type: object
      properties:
        c:
          type: string
```
As well as the following options:
```yaml
package: example
generate:
  models: true
import-mapping:
  b.yaml: example.com/b
output-options:
  skip-prune: true
```
Running `oapi-codegen/oapi-codegen -config options.yaml a.yaml` results in the following type for `Merged`:
```go
type Merged struct {
	A *string               `json:"a,omitempty"`
	B *[]Element            `json:"b,omitempty"`
	C *externalRef0.Element `json:"c,omitempty"`
}
```
Where `B` should be `*[]externalRef0.Element` instead.

hego-fdc:
Fixes #1841


LarsStegman:
@jamietanna any chance this could get merged? I think it would fix many open issues.

#1841
#2008
#2060 
#2088

mromaszewicz:
Thank you for submitting this PR, @hego-fdc — but we need to close it because the recursive external-ref propagation work landed via #2299 (commit `1c89479b`), which extracted a `propagateRemoteRefs` helper that walks `Items`, `AdditionalProperties`, and `AllOf`/`AnyOf`/`OneOf`/`Not`. A repro of the issue #1841 spec now generates `B *[]externalRef0.Element` correctly.

Thanks for the contribution — you correctly identified the gap.

PR #1876, opened 2025-01-06
use `x-go-name` as the exact value of the field

paulmach:
This is in response to https://github.com/oapi-codegen/oapi-codegen/issues/1851 which wants to use `HARD_CODED_ID` and the field name is CamelCased and the `_` are removed.

This PR treats the `x-go-name` as the exact name for the field (or parameter). We trust the user that if they're overriding, they're doing it the exact way they want. This supports `unexportedFields` or fields with `_`s.

As a result there is no need for the extra `x-oapi-codegen-only-honour-go-name` extension and `allow-unexported-struct-field-names` global option, so those have been removed.


jamietanna:
Did you intentionally close this, @paulmach?

paulmach:
@jamietanna yeah.... I don't personally need it and didn't want to deal with the rebase.

Issue #1872, opened 2025-01-01
strict fiber server returns nothing when using schema with oneOf


holyfuchs:
oapi-codegen --version: v2.4.1
go 1.23.4

Generating code with the following config:
```
package: main
generate:
  fiber-server: true
  strict-server: true
  models: true
```

and a openapi.yaml with a oneof:
```
openapi: 3.0.0
paths:
  /test:
    get:
      responses:
        '200':
          description: Ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Test'
components:
  schemas:
    Test:
      oneOf:
        - $ref: '#/components/schemas/Test1'
        - $ref: '#/components/schemas/Test2'
    Test1:
      type: object
      properties:
        test1:
          type: string
      required:
        - test1
    Test2:
      type: object
      properties:
        test2:
          type: string
      required:
        - test2
```
generates code:
```
type Test struct {
	union json.RawMessage
}
type GetTest200JSONResponse Test
func (response GetTest200JSONResponse) VisitGetTestResponse(ctx *fiber.Ctx) error {
	ctx.Response().Header.Set("Content-Type", "application/json")
	ctx.Status(200)

	return ctx.JSON(&response)
}
```
since union starts with a lower case letter the endpoint won't return anything.
I would expect it to return my Test struct.

changing the generated code to `return ctx.JSON(&response.union)` fixed the issue

holyfuchs:
it seems a workaround is changing the openapi.yaml to the following:
```
        '200':
          description: Ok
          content:
            application/json:
              schema:
                oneOf:
                  - $ref: '#/components/schemas/Test1'
                  - $ref: '#/components/schemas/Test2'
```
and to make sure the now unused Test is still generated adding the following to the output options
```
output-options:
  skip-prune: true
```

mromaszewicz:
The bug was that `type GetTest200JSONResponse Test` (a defined type, not an alias) did not inherit `Test.MarshalJSON()`, so `ctx.JSON(&response)` serialized the struct without exported fields and emitted `{}`. Commit 5260c720 (2026-04-21) fixed this by adding `HasCustomMarshalJSON` to schema.go and generating MarshalJSON/UnmarshalJSON delegators in all strict server templates when the response schema is a $ref to a oneOf/anyOf union. The repro spec now generates correct delegators in the current codebase. The fix was landed as 'Fixes #970' but resolves the identical underlying problem reported in #1872.

[...]

PR #1877, opened 2025-01-07
fix(codegen): handle ref with oneOf in json response body

holyfuchs:
Fix codegen: Handle ref with oneOf in JSON response body
#1872 

- Added a new boolean field RefIsUnion to the schema model to indicate if the reference is a union (oneOf or anyOf).
- Updated the tmpl to handle cases where RefIsUnion is true:
    - Generates code in the format: response.union.
- Added tests to the strict server


mromaszewicz:
Thank you for submitting this PR, @holyfuchs — but we need to close it because the underlying bug (strict-server response type as `$ref` to `oneOf`/`anyOf` serializing as `{}`) was fixed via commit `5260c720` taking a different approach: a `HasCustomMarshalJSON()` method on `Schema` triggers delegating `MarshalJSON`/`UnmarshalJSON` on the response type, rather than the `RefIsUnion` boolean field this PR proposed. Issue #970 has a regression test at `internal/test/issues/issue-970/`.

Appreciate the diagnosis — you pointed at the right symptom.

PR #1879, opened 2025-01-08
feat(#1726): add literal colon support for gin, echo, fiber

### LINKED ISSUE #1726: Generated client is not working when there are more than one API's with the colon in the API path

cosban:
Closes https://github.com/oapi-codegen/oapi-codegen/issues/1726

This MR supersedes https://github.com/oapi-codegen/oapi-codegen/pull/1815


mromaszewicz:
@greptileai

mromaszewicz (approved):

mromaszewicz:
I've basically rewritten this PR as an additional commit. Thanks for submitting this ages ago, and sorry for such a long delay.

PR #1880, opened 2025-01-09
support all x-* extensions and with ref

paulmach:
This is meant to address https://github.com/oapi-codegen/oapi-codegen/issues/863 which asks for `x-go-type` to work with `$ref`. It also fixes my use case which is for `x-oapi-codegen-extra-tags` like:
```
obj:
  type: object
  properties:
    start_port:
      $ref: '#/components/schemas/Port'
      x-oapi-codegen-extra-tags:
        bson: start_port
    end_port:
      $ref: '#/components/schemas/Port'
      x-oapi-codegen-extra-tags:
        bson: end_port
```

I have added tests in the `internal/tests/extensions` directory to cover a few of the extensions, I can add a complete list if desired. 

We already treat `x-order` in this way (use the value next to the ref if provided) as of https://github.com/oapi-codegen/oapi-codegen/pull/1700 and this pr makes the functionality complete and consistent for all supported extensions.


mromaszewicz:
@greptileai Please review this commit, with a thorough focus on whether having an extension as a sibling of $ref is valid - specifically, see if you can infer how this will interact with request validation using the openapi3 filter (https://github.com/getkin/kin-openapi/tree/master/openapi3filter).

More test cases might be needed. This code change is in a very sensitive path.

mromaszewicz:
This change makes sense to me. I'm totally fine doing things that aren't recommended by the spec, as long as they work for us.

Normally, when $ref is set in a kin SchemaRef, the Schema value is inlined, so I'm going to dig into this a bit more thoroughly to see how this interacts with the schema lookup in our ref handling. I don't think we're consistent internally.

mromaszewicz:
@greptileai - new code has been pushed, re-evaluate on latest commit. I found some other places that needed improvement around losing sibling ref extensions. The only one with noticeable impact was on parameters.

mromaszewicz (approved):

PR #1895, opened 2025-02-07
fix: Print the warning message to stderr instead of stdout

ignassew:
This is the same PR as #1891 but without not on the main branch.

Warning being printed to stdout effectively disables the use of redirecting output to file. This PR fixes it.


jamietanna:
Thanks! Will want to double check that there's nowhere else we've got a similar thing, agree that this isn't helpful when you're redirecting to a file, apologies!

jamietanna (approved):

PR #1901, opened 2025-02-12
build: build against Go 1.24

jamietanna:
Now it's out, we should make sure folks using the project on the newest
version are still supported.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,7 @@ jobs:
         - "1.21"
         - "1.22"
         - "1.23"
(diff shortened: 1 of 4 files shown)


Priyadarshini-Mitra:
Hello @jamietanna when shall this supported version be released ? I am looking for a stable oapi-codegen version with go 1.24 support. Thanks !

jamietanna:
Hey, `oapi-codegen` as of the last release will work with Go 1.24 as-is, there's nothing needed (that we've seen or had reported) for building old versions with Go 1.24

PR #1910, opened 2025-02-21
docs: simplify go tool example

gaiaz-iusipov:


jamietanna:
Hmm, I didn't think that worked - mind updating the example in https://github.com/oapi-codegen/oapi-codegen/tree/main/examples/minimal-server/stdhttp-go-tool to match, too?

gaiaz-iusipov:
Sure, of course

gaiaz-iusipov:
It looks like the linter is failing due to the Go version in the `go.mod` file. Should I revert the file to its previous version? In that case, the code example might not be entirely correct

gaiaz-iusipov:
Alright, I don't think there's any simple solution other than `go mod edit -go=1.22`.
I reverted the version back

jamietanna:
Thanks! I think that change you reverted is something worthwhile, so I'll get a PR raised for that and merge it prior to this

jamietanna (approved):
Thanks again!

PR #1911, opened 2025-02-23
Add two tools for full API first design experience

oussamarouabah:
Adding 2 tools in the README for full API first design experience:

Imagine this:
DSL to OpenAPI to Golang code should be better than writing the OpenAPI spec by hand which was the drawback of the design first approach but now with these tools, this should be better:

- TypeSpec from Microsoft, most of Azure backend is using this tool https://typespec.io/
- Smithy from Aws https://smithy.io/

I hope you merge this, these tools have been so helpful to me especially TypeSpec because it kind of feels like TypeScript

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,10 @@
 
 `oapi-codegen` is a command-line tool and library to convert OpenAPI specifications to Go code, be it [server-side implementations](#generating-server-side-boilerplate), [API clients](#generating-api-clients), or simply [HTTP models](#generating-api-models).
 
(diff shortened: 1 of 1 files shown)


jamietanna:
Thanks for this, but having not used these, we can't recommend them. Thanks though!

oussamarouabah:
@jamietanna did you review these tools before closing the PR?
 I ask because they can be really helpful for the community. When I started with oapi-codegen, I used `github.com/swaggo/swag/v2/cmd/swag@v2.0.0-rc4` with decorator comments and it was useful, but these tools can help define APIs in a nice/better way.
If you'd prefer, I can rephrase this from a recommendation to simply informing people that these tools exist. Please let me know if you'd like me to adjust the wording.

PR #1913, opened 2025-02-24
feat: opai-codegen#1912 add optional default format for numbers

matgr1:
adds `default-number-format` to output options to allow configuring the default format used for `type: number` (without this, it defaults to 32-bit `float`)


mromaszewicz:
Thank you for submitting this PR, @matgr1 — but we need to close it because the more general `output-options.type-mapping` feature delivered via #2223 covers this use case: setting `output-options.type-mapping.number.default.type: float64` does what the proposed `default-number-format` flag would have. The narrower flag wasn't added, but the underlying need is met.

Thanks for the contribution.

PR #1922, opened 2025-03-12
fix: use `.String()` method for making a string from uuid

### LINKED ISSUE #1914: Cannot convert text/plain uuid in requestBody to string
Hello, oapi-codegen generates an invalid code when I define `requestBody` as text/plain uuid.

Image

I would appreciate it if you could advise me on how to fix this.

## How to reproduce

```shell
$ cat > openapi.yaml <<EOS
openapi: 3.0.3
paths:
  /pet:
    post:
      requestBody:
        content:
          text/plain:
            schema:
              type: string
              format: uuid
EOS

$ go run github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@latest openapi.yaml > gen.go
$ go mod init openapi
$ go mod tidy

$ go run github.com/golangci/golangci-lint/cmd/golangci-lint@latest run
```

output:

```
gen.go:1: : # gen
./gen.go:135:40: cannot convert body (variable of array type PostPetTextRequestBody) to type string (typecheck)
// Package openapi provides primitives to interact with the openapi HTTP API.
exit status 1
```

Lastly, thank you for maintaining such a great tool. Really appreciate it!

MilkeeyCat:
fixes #1914

### DIFF
--- a/pkg/codegen/templates/client.tmpl
+++ b/pkg/codegen/templates/client.tmpl
@@ -147,6 +147,8 @@ func New{{$opid}}Request{{.Suffix}}(server string{{genParamArgs $pathParams}}{{i
             return nil, err
         }
         bodyReader = strings.NewReader(bodyStr.Encode())
+    {{else if and (eq .NameTag "Text") (eq .Schema.OAPISchema.Format "uuid") -}}
+        bodyReader = strings.NewReader(body.String())
     {{else if eq .NameTag "Text" -}}
         bodyReader = strings.NewReader(string(body))
     {{end -}}


0rid:
@MilkeeyCat You can achieve a similar solution using by overriding the property type in `overlay.yaml`

ysuzuki19:
This PR looks good!
I have successfully tested this on my local environment, and it works as expected.

It would be even better if you could add or update a test.

For example, you might want to consider adding a paths in:

file: `internal/test/schemas/schemas.yaml`
line: `150`
content:
```yaml
  /issues/1914:
    post:
      operationId: Issue1914
      description: |
        Ensure that anonymous inner types are generated correctly.
      requestBody:
        content:
          text/plain:
            schema:
              type: string
              format: uuid
```

and then run `make generate` on this repository root.

jamietanna:
Thanks!

Looking at this, I'm wondering if we should put in a wider fix, which is that we try and convert the type to a `string` i.e. with `fmt.Sprintf` or something similar, so we end up returning something rather than hitting the compilation error - thoughts?

MilkeeyCat:
> so we end up returning something rather than hitting the compilation error - thoughts?
Hm, honestly I prefer compilation error over incorrectly stringified value

jamietanna:
Hmm - I get your thinking, but also not sure :thinking: I've raised https://github.com/oapi-codegen/oapi-codegen/pull/1975 in the meantime, let's see what some others think about the behaviour :+1:

jamietanna:
Maybe let's add a `compatibility-option` to disable the fallback?

MilkeeyCat:
I don't understand what fallback you're talking about, but #1975 looks good to me

jamietanna:
Thanks - so I'm thinking we could add an entry into [the `CompatibilityOptions`](https://pkg.go.dev/github.com/oapi-codegen/oapi-codegen/v2/pkg/codegen#CompatibilityOptions) to allow us to push #1975 through as-is, but allow folks to opt-in to "I want to have a compilation error when there's not a `String()` method on the type"

MilkeeyCat:
With the way #1975 is done I personally don't think this option is needed.

> ... allow folks to opt-in to "I want to have a compilation error when there's not a String() method on the type"

How would that work with integer, string types? They don't have `.String()` method

jamietanna:
> With the way #1975 is done I personally don't think this option is needed.
> 
> > ... allow folks to opt-in to "I want to have a compilation error when there's not a String() method on the type"
> 
> How would that work with integer, string types? They don't have `.String()` method

Ah yeah, so we'd probably have the fallback of "if someone's using the compatibility option for the fallback, use `string(body)`, and if they're not, use `fmt.Sprintf(...)`"

mromaszewicz:
I think that automatically looking for a Stringer implementation via a safe type cast is about as good as we can do.

Marshaling primitive types as text/plain makes sense.

In the case of having text/plain media types for complex types, I think we should error out, the question is where to do it. We could look for a stringer implementation at runtime, and if one is there, use that. The way the code works currently is that on the server side, it's up to the user how to encode something as text/plain, we stay out of it.

If a user wants to do some fancy text encoding on complex types, let them implement Stringer manually on those types.

mromaszewicz:
Ok, I think the way this should work is that we don't even look at UUID explicitly, the flow would be:

```
if json:
  ...
elif form
  ...
elif text:
   if bodyStringer, ok := body.(fmt.Stringer); ok : 
      bodyReader = strings.NewReader(bodyStringer.String())
   else if is_primitive_type(body):
      bodyReader = strings.NewReader(string(body))
    else:
      return error
```

The is_primitive_type(body) can be deduced by the codegen and provided as a template context flag.

Not having a codegen error allows a person to implement stringer. If they don't, they're going to get a runtime error.

mromaszewicz:
It looks like PR #1975 is close to what I want, so I think it supersedes this PR.

jamietanna:
Thanks for this - we've gone with #1975, as it handled a couple of extra things

PR #1928, opened 2025-03-15
fix: multiline indented lines in parameter descriptions break overlay functionality

### LINKED ISSUE #1927: Description with indentation indicator may break overlay functionality
We have hit one case where overlay functionality gets completely broken due to unserialize-serialize-deserialize of YAML files and may be a bug in underlying yaml library.

This API spec works fine if not using overlay functionality:

```yaml
openapi: 3.0.1
info: {}
paths:
  /info:
    get:
      parameters:
        - description: |2-
              a
            b
          name: filter
          schema:
            type: string
          in: query
      responses: {}
```

However, once overlay is added, tool fails with an error:
```
error loading swagger spec in spec.yaml
: failed to parse spec from "spec.yaml": yaml: line 8: did not find expected key
```

It seems like it's because `LoadSwaggerWithOverlay` re-marshals the spec into this:

```yaml
info:
    title: ""
    version: ""
openapi: 3.0.1
paths:
    /info:
        get:
            parameters:
                - description: |4-
                      a
                    b
                  in: query
                  name: filter
                  schema:
                    type: string
            responses: {}
```

.. which is no longer a valid YAML.

This seems to be specific to parameter list though; if adding description to "get" directly, this bug does not appear.

viktorasm:
Fixes #1927


mromaszewicz:
Thank you for submitting this PR, @viktorasm — but we need to close it because the same workaround (setting the YAML encoder indent to 2, working around go-yaml#76) was independently merged via #2172 (commit `77bda899`). `pkg/util/loader.go:53` now contains `enc.SetIndent(2)` with the comment referencing the upstream issue.

Thanks for the diagnosis.

PR #1929, opened 2025-03-15
fix: properly handle quotes in content-type when generating server code

### LINKED ISSUE #1799: Quotes in Content-Type cause "error generating code: error formatting Go code"
Sample `api.yml`:
```
openapi: "3.0.0"

info:
  version: 0.0.1
  title: example

paths:
  /object:
    get:
      responses:
        '200':
          content:
            "application/ld+json; profile=\"https://www.w3.org/ns/activitystreams\"":
              schema:
                type: string
```

Sample `config.yml`:
```
package: main
generate:
  gorilla-server: true
  strict-server: true
output: test.go
```

The result I get is

```
error generating code: error formatting Go code 
[ ... dump of the code it is trying to generate ... ]
: main.go:221:79: missing ',' in argument list (and 10 more errors)
```

The dumped code has some invalid syntax:

```
func (response GetObject200ApplicationLdPlusJSONProfilehttpswwwW3OrgnsactivitystreamsResponse) VisitGetObjectResponse(w http.ResponseWriter) error {
    w.Header().Set("Content-Type", "application/ld+json; profile="https://www.w3.org/ns/activitystreams"")
    w.WriteHeader(200)
````

This can be worked around by double-escaping the quotes in the input (ie using `\\\"` instead of `\"`).

viktorasm:
fixes #1799

OpenAPI media types may contain characters that require escaping in Go string literals, for example, `application/ld+json; profile="https://..."` which the templates were emitting verbatim, producing invalid Go.

PR #2217 had a partial fix for this, and this PR fixes the other cases.

Adds a regression test under `internal/test/issues/issue1799/` covering chi, client, echo, fiber, gin, gorilla, iris, and std-http generation modes with a spec containing a quoted media-type parameter.


FenTiger:
I gave this a quick try. It's now able to generate code for my original API definition, so this looks good to me.

mromaszewicz:
I'm going to rebase and fix up this commit then merge it. It does address a few bugs, thanks.

mromaszewicz (approved):

PR #1931, opened 2025-03-17
fix(codegen): update JSON tags to include omitempty for nullable fields

Dumbris:
# Fix omitempty tag for nullable fields

## Description
This PR fixes a bug in the JSON tag generation for struct fields that represent nullable but non-required properties in OpenAPI schemas. 

The issue was in the conditional logic that determines when to add the `omitempty` tag:

```go
// Before
shouldOmitEmpty := (!p.Required || p.ReadOnly || p.WriteOnly) &&
    (!p.Required || !p.ReadOnly || !globalState.options.Compatibility.DisableRequiredReadOnlyAsPointer)

omitEmpty := !p.Nullable && shouldOmitEmpty
```

The bug occurred because `!p.Nullable` would be false for nullable fields, causing `omitEmpty` to be false even when the field wasn't required. This led to nullable but optional fields incorrectly missing the `omitempty` tag.

## Changes
- Fixed the conditional logic to correctly apply `omitempty` to all non-required fields regardless of nullability
- Updated tests to verify the corrected behavior

## Why this makes sense
1. According to the OpenAPI spec, `nullable: true` means a value can be null, while the `required` array determines if a property must be present
2. In Go, the `omitempty` tag controls whether a field is omitted when it's the zero value (not directly related to nullability)
3. Non-required fields should have `omitempty` to correctly handle JSON serialization when the field is absent

This change ensures the generated code better aligns with both OpenAPI semantics and Go's JSON marshaling conventions.

### DIFF
--- a/pkg/codegen/codegen_test.go
+++ b/pkg/codegen/codegen_test.go
@@ -110,8 +110,8 @@ func TestExtPropGoTypeSkipOptionalPointer(t *testing.T) {
 	assert.NoError(t, err)
 
 	// Check that optional pointer fields are skipped if requested
-	assert.Contains(t, code, "NullableFieldSkipFalse *string `json:\"nullableFieldSkipFalse\"`")
-	assert.Contains(t, code, "NullableFieldSkipTrue  string  `json:\"nullableFieldSkipTrue\"`")
+	assert.Contains(t, code, "NullableFieldSkipFalse *string `json:\"nullableFieldSkipFalse,omitempty\"`")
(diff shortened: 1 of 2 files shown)


mromaszewicz:
Thank you for submitting this PR, @Dumbris — but we need to close it because the same fix (removing the `\!p.Nullable` guard from `omitEmpty` so optional nullable fields receive the `omitempty` JSON tag) was independently merged via #2221 (commit `d567e49b`). `pkg/codegen/schema.go:796` now reads `omitEmpty := shouldOmitEmpty` with no nullable gating, and `codegen_test.go:120-121` asserts the corrected behavior.

Thanks for the contribution.

PR #1944, opened 2025-04-01
feat(extensions): add support for `x-omitzero`

lzap:
SSIA

I do not want to add this by default, not sure what consequences would be, but I want to be able to start playing around with it.

I can do add documentation as well to the README if you like it.


examples/extensions/xomitzero/api.yaml
```diff
@@ -0,0 +1,26 @@
...
+          type: number
+    ClientWithExtension:
+      type: object
+      required:
+        - name
+      properties:
+        name:
+          type: string
+          x-omitempty: true
+          x-omitzero: true
```
jamietanna, line 24: Do both of these need to be set, here? Or should we only have `x-omitzero`?

lzap: Sure, probably a leftover. Rebased.

lzap:
Any other comments? This aligns with prior art and does not break anything existing.

pkg/codegen/schema.go
```diff
@@ -724,31 +731,32 @@ func GenFieldsFromProperties(props []Property) []string {
...
-			fieldTags["json"] = p.JsonFieldName + ",omitempty"
-			if globalState.options.OutputOptions.EnableYamlTags {
-				fieldTags["yaml"] = p.JsonFieldName + ",omitempty"
-			}
-			if p.NeedsFormTag {
-				fieldTags["form"] = p.JsonFieldName + ",omitempty"
-			}
+		fieldTags["json"] = p.JsonFieldName +
+			stringOrEmpty(omitEmpty, ",omitempty") +
+			stringOrEmpty(omitZero, ",omitzero")
```
jamietanna, line 753: I'm not sure if we should have _both_ `omitempty` and `omitzero` :thinking:

lzap: Well, it is a valid Go struct syntax and while I cannot come up with a valid usecase now, it does not mean it does not exist. Let me put it this way: why would be a problem to support both at the same time?

jamietanna:
(Ignore that check, it's fine!)


I'm still working on this one - trying to get a good example of exactly how `omitzero` works and what differences in behaviour we'll see - largely out of my own learning!

pkg/codegen/schema.go
jamietanna: Sorry, thought I'd replied to this! It turns out there is _likely_ no issue ([via](https://pkg.go.dev/encoding/json)):

> If both "omitempty" and "omitzero" are specified, the field will be omitted if the value is either empty or zero (or both).

jamietanna:
I'll follow-up with a PR to allow tuning _only_ one or the other of `omitempty` and `omitzero` in case folks want more control

pkg/codegen/schema.go
lzap: Yeah I was not sure, I saw some articled explaining that having both would make sense but I could not google them back.

PR #1957, opened 2025-04-22
fix(codegen): allow using x-go-type and x-go-type-skip-optional-pointer together

Nivl:
When using `x-go-type`, `GenerateGoSchema` returns early, meaning that any extensions that are checked after this one are no-op.

This PR moves the code for `x-go-type-skip-optional-pointer` higher up, so it can be used alongside `x-go-type`.

This PR only fixes a bug, not the root issue. There may be more similar issues, and the code makes it easy to introduce new similar bugs. Ultimately we'd need to refactor `GenerateGoSchema` to:
1. Add tests
2. Not return early (to avoid having the same issue on). We return early at multiple places in this method.

This isn't the goal of that PR though, so maybe in another one.


jamietanna:
Thanks! Will take a look hopefully today, and see if I can get some tests added to cover this

jamietanna:
Do you have an example (minimal spec) that this change improves the codegen for? In my limited testing, I can't seem to get much meaningfully different when applying this PR

I _can_ see that looking a this with an `allOf` does actually provide some differences in the generated code for a case I'm interested in, but can't seem to see this PR as-is changing much, but maybe I'm missing something?

jamietanna (approved):

PR #1961, opened 2025-04-30
Backporting fix for optional parameters from 1.10.1

tanmaydalvivmw:
from the commit:
7ee55a9 Fix parse forms explode object with optional args


jamietanna:
Was this unintentionally raised here, and intended for a fork?

tanmaydalvivmw:
Yes, unintentionally raised here. Sorry for being a bother.

PR #1971, opened 2025-05-08
fix: follow-ups to #1957

jamietanna:


jamietanna (commented):
I'd hoped that #1610 had fixed this, but unfortunately not - need to keep digging

jamietanna:
https://github.com/oapi-codegen/oapi-codegen/pull/1981 is required for this

jamietanna:
OK, the issue I'm trying to fix is that:

- IsGoTypeReference needs to get called as part of the `AllOf` merging, but isn't - so we're calling `valueWithPropagatedRef` which then returns the value, not a `$ref`

mromaszewicz:
Closing — the follow-ups landed via commit `cc72767e` and PR #2042 (commit `702cc917`); a repro confirms `x-go-type` + `x-go-type-skip-optional-pointer` together work correctly in `allOf`. Test infrastructure from this PR is partially merged via commits `b4a1b00c` and `cc72767e`.

Issue #1262, opened 2023-09-14
Form tag is not generated on structs when using multipart/form-data


sonu27:
I get `request body has an error: failed to decode request body: part Fields.0.Name: undefined`

But this goes away if I manually change the generated struct to have `Name     string `form:"name",json:"name"`

Bug?

DMRobertson:
I'm not sure what the original reporter's setup was, but I noticed this when generating a (non-strict) echo server.

For example, the OpenApi schema

```yaml
openapi: "3"
info:
  title: "Issue 1262 - form tags for multipart/form-data"
  version: "1"
paths:
  /attachment:
    post:
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                description:
                  type: string
                  description: Summary description of the upload.
                data:
                  type: string
                  format: binary
                  description: User-provided file.
              required:
                - data
      responses:
        204:
          description: "No content"
```

generates the type

```go
// PostAttachmentMultipartBody defines parameters for PostAttachment.
type PostAttachmentMultipartBody struct {
	// Data User-provided file.
	Data openapi_types.File `json:"data"`

	// Description Summary description of the upload.
	Description *string `json:"description,omitempty"`
}
```

I expected to be able to `ctx.Bind()` a request body into a `PostAttachmentMultipartBody` value, where `ctx` is an echo.Context. Digging in to the codegen package and echo's internals, I found a few obstacles:

1. It looks like echo's default binder doesn't consider `openapi_types.File` a valid destination for a multipart file value; it uses [an explicit allowlist](https://github.com/labstack/echo/blob/de44c53a5b16f7dca451f337f7221a1448c92007/bind.go#L421-L441) of stdlib types.
2. Echo only considers [fields tagged with `form`](https://github.com/labstack/echo/blob/de44c53a5b16f7dca451f337f7221a1448c92007/bind.go#L110-L112) for binding, but `oapi-codegen` doesn't emit these.
3. Binding to form tags marked as `omitempty` is unsupported (or considered invalid?) by echo, per https://github.com/labstack/echo/issues/2612 . Not sure what the rest of the ecosystem does here though.
   - Digging in: `inputFieldName` [here](https://github.com/labstack/echo/blob/de44c53a5b16f7dca451f337f7221a1448c92007/bind.go#L235) is e.g. `description,omitempty` rather than just `description`.

triet-truong-siglarcarbon-vn:
[...]

PR #1974, opened 2025-05-10
fix: allow parsing non-file data in multipart form data bodies via echo binding

DMRobertson:
For your consideration: this is intended to reproduce #1262, see https://github.com/oapi-codegen/oapi-codegen/issues/1262#issuecomment-2868180395 in particular. I've also included a proposed fix which lets you `Bind()` to read non-file parameters in an `multipart/form-data`. It looked like getting echo to `Bind() file-type parts would be a more invasive change though.

I'm not sure if this is the _right_ change, or even if you agree that the report in 1262 is valid! But I spent enough time investigating that I thought it was worth sharing.


internal/test/issues/issue-1262/issue_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	/* Disabled: echo doesn't recognise openapi_types.File as something that
+		   a multipart.Reader---see echo.DefaultBinder.bindData.
+	t.Run("parses form files", func(t *testing.T) {
+		assert.Equal(t, "script.txt", server.ParsedUpload.Data.Filename())
+		dataBytes, err := server.ParsedUpload.Data.Bytes()
+		require.NoError(t, err)
+		assert.Equal(t, "Once upon a time in a galaxy far, far away\n", string(dataBytes))
+	})
+	*/
```
DMRobertson, line 66: There's some more info in https://github.com/oapi-codegen/oapi-codegen/issues/1262#issuecomment-2868180395, but TL;DR: echo is expecting to bind to a stdlib multipart type, so this doesn't work.

pkg/codegen/schema.go
```diff
@@ -733,7 +733,7 @@ func GenFieldsFromProperties(props []Property) []string {
 				fieldTags["yaml"] = p.JsonFieldName + ",omitempty"
 			}
 			if p.NeedsFormTag {
-				fieldTags["form"] = p.JsonFieldName + ",omitempty"
+				fieldTags["form"] = p.JsonFieldName
```
DMRobertson, line 736: This got my test passing, but I wasn't sure if it was a good idea. If it is a good idea, does it merit a backwards compatibility flag?

Summarising https://github.com/oapi-codegen/oapi-codegen/issues/1262#issuecomment-2868180395, echo doesn't consider `omitempty` to make sense in forms. Not sure about other HTTP servers. I quick search found https://github.com/go-playground/form/issues/18 but I don't think that's used in an HTTP server like echo or gin etc.

mromaszewicz: his isn't safe, because it will affect everyone's generated code.

mromaszewicz:
Thanks for this PR. I'm fixing it up to merge it shortly. I'll push a rebase and some updates to your PR branch. I can't take all your changes as-is, because they will break some people's code.

pkg/codegen/schema.go
mromaszewicz: Your fix to remove `,omitempty` globally might have been the right one, hmm. I have to see what impact this has on other routers. Maybe the fix needs to be constrained to just models built for Echo.

mromaszewicz:
This seems to work on our current versions of gin, fiber and iris. It's not applicable in the other routers because they don't help you Bind stuff.

Removing `omitempty` has no bad effect on Binding in all those routers, and it fixes the Echo case.

So, I looked into `runtime.MarshalForm` to see how it would be affected, and I discovered that it's buggy. It's using the `json` tag to control `form` output, which is incorrect. If I globally disable `omitempty` it will affect the form marshaling as expected - zerovalue fields would be emitted with their zerovalue.


We've hit the first case where we want to have a model look differently to different routers.

mromaszewicz:
I know what I am going to do. In my experimental version at `oapi-codegen/oapi-codegen-exp`, I've prototyped full user customization of the struct tags. I think I will back-port that change to this repo, and them adapt your PR and merge it afterwards.

mromaszewicz:
It looks like we were overzealous in deciding when to apply the form tag to fields. It doesn't need to apply to form parameters, as those aren't of the form media type.

mromaszewicz:
@greptileai Please review this. I am intentionally making a change to remove `omitempty` from things that aren't specified as being form encoded. This includes form type parameters. You can ignore the code review directive here about making breaking changes, since the new behavior is correct.

mromaszewicz:
I think that I will do this first (https://github.com/oapi-codegen/oapi-codegen/issues/2368) then revisit this PR.

Echo users will need to remove the `,omitempty` from their models that need `application/x-www-form-urlencoded`

Issue #1052, opened 2023-05-19
Have an option to create directories if they don't already exist


renxinhe:
I'm getting the following error, because my repo currently only has "pkg/entity", and "oapimodel/" doesn't already exist.

```go
//go:generate oapi-codegen -package oapimodel -generate types,skip-prune -o pkg/entity/oapimodel/oapimodel.gen.go openapi_spec.yaml
```

> error writing generated code to file: open pkg/entity/oapimodel/oapimodel.gen.go: no such file or directory

I'm wonder if we can quickly support some input flag, or modify the default behavior, to create the necessary directories to support any destination path?

jamietanna:
I would recommend probably something like the following in the short term:

```
//go:generate mkdir -p pkg/entity/oapimodel
//go:generate oapi-codegen -package oapimodel -generate types,skip-prune -o pkg/entity/oapimodel/oapimodel.gen.go openapi_spec.yaml
```

This appears to be the first time I've seen this happen, usually the codegen is done to the same location as the `go:generate` directives.

This is something I think that we'd be happy to get a PR raised for, but in the meantime, feel free to use a hacky solution like above 😁

renxinhe:
Thanks @jamietanna! That's exactly what I did. Not the prettiest, but hey it works! Hope to see a quality-of-life improvement in a future release!

PR #1994, opened 2025-05-24
fix: create directories if they do not exist before writing output file

kf-pineapple:
#1052 

If the parent directory of the output file doesn't exist, it will be automatically created before writing the generated code. This prevents errors if the specified output path contains a non-existent directory.


jamietanna (approved):
Thanks! I've manually tested this - I don't think we have a straightforward way of adding a test for this, at this point, but I'm happy it being a manual case

PR #2000, opened 2025-06-11
docs(extensions): update example code for `x-enumNames`

jamietanna:
As the copy in the README was out-of-sync with the generated code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2774,17 +2774,26 @@ const (
 	EXP ClientType = "EXP"
 )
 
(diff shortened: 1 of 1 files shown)


jamietanna:
Similar to https://github.com/oapi-codegen/oapi-codegen/pull/1999, it looks like these docs weren't as well reviewed (by me) before they were merged

jamietanna:
Thanks @pugnascotia for the report!

PR #2002, opened 2025-06-14
feat(generate): allow generating Server URL boilerplate

jamietanna:
When working with a client generated by `oapi-codegen`, it can be a
little awkward to manage changing URLs for the `client.WithBaseURL`.

Although it's possible to manage these ourselves - with hardcoded values
- it can be handy to have this pre-generated, especially if the input
spec already defines `$.servers`.

This introduces the capability to opt-in to the generation (as we try to
make all changes opt-in where possible) of these.

This is a little more complicated than ""just"" generating constants, as
a Server object could introduce a templated URL.

This requires we:

- generate the `const`s for non-templated URLs
- generate the more in-depth boilerplate for templated URLs

We try and do most of the generating in the template, rather than in Go
code, although it's not straightforward to generate the method
parameters for the generated function (i.e.
`NewServerUrlTheProductionAPIServer`) so we create a
`genServerURLWithVariablesFunctionParams` function.

We'll leave a few cases of additional validation or handling of edge
cases to a follow-up.


pkg/codegen/server_urls.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"fmt"
+	"strconv"
+	"text/template"
+
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+const serverURLPrefix = "ServerURL"
```
jamietanna: :stop_sign: this needs to follow `nameNormalizer`

PR #2021, opened 2025-07-15
fix(output-options): obey `prefer-skip-optional-pointer` in reference types + add tests for `preferskipoptionalpointer`

jamietanna:


pkg/codegen/schema.go
```diff
@@ -270,10 +270,11 @@ func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
...
 		return Schema{
-			GoType:         refType,
-			Description:    schema.Description,
-			DefineViaAlias: true,
-			OAPISchema:     schema,
+			GoType:              refType,
+			Description:         schema.Description,
+			DefineViaAlias:      true,
+			OAPISchema:          schema,
+			SkipOptionalPointer: globalState.options.OutputOptions.PreferSkipOptionalPointer,
```
jamietanna, line 277: There seem to be a couple of other places that `Schema`s are constructed without this being set, but - as far as I can see from some initial testing - it's not straightforward to trigger that codepath from `examples/output-options/preferskipoptionalpointer` to validate them

PR #2042, opened 2025-07-23
Respect setting of `x-go-type-skip-optional-pointer`on referenced types

magnetikonline:
I've noted a regression as part of #2021.

Previously, since `Schema{}.SkipOptionalPointer` was never set, it _always_ defaulted to `false`, which suited our use case.

With the change as part of #2021 and in the new https://github.com/oapi-codegen/oapi-codegen/releases/tag/v2.5.0 release - this now uses the _global_ output option setting of `prefer-skip-optional-pointer:` in all cases - and can't be over-ridden locally within a type.

This change fixes that  behaviour, with `x-go-type-skip-optional-pointer` checked and evaluated up front for the type/referenced type - meaning a schema blob such as below will now respect the `x-go-type-skip-optional-pointer: false` value where referenced against `apples|oranges` and of course fallback to the global output option if unset:

```yaml
MyAction:
  type: object
  additionalProperties: false
  required:
    - description
  properties:
    description:
      type: string
      description: Your reasons for doing this.
      example: We need more fruit!
    apples:
      description: Hello apples.
      $ref: '#/components/schemas/Referenced'
    oranges:
      description: Hello oranges.
      $ref: '#/components/schemas/Referenced'

Referenced:
  type: object
  description: Please reference me.
  additionalProperties: false
  x-go-type-skip-optional-pointer: false
```


jamietanna:
Hey @magnetikonline thanks for raising this! 

Taking a look at the example you've shared in the body of the PR, I can't seem to make this fail on `main`, before these changes are applied.

### On `main`

If I make the changes:

```diff
diff --git examples/output-options/preferskipoptionalpointer/api.yaml examples/output-options/preferskipoptionalpointer/api.yaml
index 0395e89..684432d 100644
--- examples/output-options/preferskipoptionalpointer/api.yaml
+++ examples/output-options/preferskipoptionalpointer/api.yaml
@@ -37,3 +37,26 @@ components:
         client:
           description: This field is optional, but the `prefer-skip-optional-pointer` Output Option ensures that this should not have an optional pointer.
           $ref: '#/components/schemas/Client'
+
+    MyAction:
+      type: object
+      additionalProperties: false
+      required:
+        - description
+      properties:
+        description:
+          type: string
+          description: Your reasons for doing this.
+          example: We need more fruit!
+        apples:
+          description: Hello apples.
+          $ref: '#/components/schemas/Referenced'
+        oranges:
+          description: Hello oranges.
+          $ref: '#/components/schemas/Referenced'
+
+    Referenced:
+      type: object
+      description: Please reference me.
+      additionalProperties: false
+      x-go-type-skip-optional-pointer: false
```

This then generates the code:

```diff
diff --git examples/output-options/preferskipoptionalpointer/gen.go examples/output-options/preferskipoptionalpointer/gen.go
index 0d6a9fd..9c3c270 100644
--- examples/output-options/preferskipoptionalpointer/gen.go
+++ examples/output-options/preferskipoptionalpointer/gen.go
@@ -24,7 +24,22 @@ type ClientWithExtension struct {
        PointerId *float32 `json:"pointer_id,omitempty"`
 }
 
+// MyAction defines model for MyAction.
+type MyAction struct {
+       // Apples Please reference me.
+       Apples *Referenced `json:"apples,omitempty"`
+
+       // Description Your reasons for doing this.
+       Description string `json:"description"`
+
+       // Oranges Please reference me.
+       Oranges *Referenced `json:"oranges,omitempty"`
+}
+
 // NestedType defines model for NestedType.
 type NestedType struct {
        Client Client `json:"client,omitempty"`
 }
+
+// Referenced Please reference me.
+type Referenced = map[string]interface{}
```

### With this PR

If I then use the changes in this PR, this doesn't end up changing anything :thinking: Mind taking a look?

magnetikonline:
Thanks @jamietanna will do.

Things are certainly working as I'd expect with our internal schemas (which of course I can't post here) - but I'll get the example into a state which trips the change I'm trying to make. 👍 

Cheers.

jamietanna:
NP, thanks! I've pushed a (failing) change for the example above, which flags that the example you'd shared doesn't seem to work right now

(I'm very happy to collaborate on this, or if you'd prefer I move the fix for the issue I've just pushed to a separate PR)

magnetikonline:
Hey @jamietanna - was having a brain fart yesterday trying to work this out 😄 - with fresh eyes today and reviewing our internal schemas - I've been able to produce an example that shows drift between the prior `v2.5.0` release, `v2.5.0` and what this PR will remediate.

It's when the referenced type is a _child_ of an `allOf` / `anyOf` / etc.

As a diff to your update:

```diff
diff --git examples/output-options/preferskipoptionalpointer/api.yaml examples/output-options/preferskipoptionalpointer/api.yaml
index 369d281..8bb64c8 100644
--- examples/output-options/preferskipoptionalpointer/api.yaml
+++ examples/output-options/preferskipoptionalpointer/api.yaml
@@ -51,7 +51,8 @@ components:
           # TODO
           $ref: '#/components/schemas/ReferencedWithoutExtensionMap'
         withExtension:
-          $ref: '#/components/schemas/ReferencedWithExtension'
+          allOf:
+            - $ref: '#/components/schemas/ReferencedWithExtension'

     ReferencedWithExtension:
       type: object
```

This _will_ give different behaviour between this PR and the `v2.5.0` release - which I've tested.

To expand as to why?

Pre `v2.5.0` https://github.com/oapi-codegen/oapi-codegen/blob/e93501d11774bb6e327c17a77666e968830907b8/pkg/codegen/schema.go#L272-L278

Since `Schema{}` was never returned with property `SkipOptionalPointer` set, it was _always `false`_ - which wasn't quite right, but suited our usage patterns.

Moving to https://github.com/oapi-codegen/oapi-codegen/commit/fad490f90982ec216cd2f9a00541cc3799af1da9 - now `SkipOptionalPointer` is set to the global skip pointer setting - totally changing the behavior for these type definitions and the Golang client(s) generated.

This PR will repair both these above cases. I still have to make changes to our schemas with an additional sprinkingling of `x-go-type-skip-optional-pointer: false` properties - but it is now more "correct" and I've got total control over how the Golang clients are generated. 

I've not committed to my branch - I didn't want to clobber your commits - but I hope this should give you a good working example of what this PR fixes. I've also updated the PR description to suit to match the test example. 👍

> (I'm very happy to collaborate on this, or if you'd prefer I move the fix for the issue I've just pushed to a separate PR)

nah more than happy to pair on this PR here 👍

jamietanna:
Hey @magnetikonline happy for you to push up to this branch (you could do it as separate commits on top of what's pushed here, as we generally squash-merge external contributions)

Thanks for digging into that, that makes sense

magnetikonline:
> Hey @magnetikonline happy for you to push up to this branch (you could do it as separate commits on top of what's pushed here, as we generally squash-merge external contributions)

Thanks @jamietanna - I've tried to keep the tests same/same to the others, I think that's enough to show it does what it proposes on the tin. 👍

> Thanks for digging into that, that makes sense

not a problem - happy I was able to explain myself!

jamietanna (approved):

magnetikonline:
Thanks @jamietanna for your time on this and getting it merged. Most appreciated. 👍

Issue #2007, opened 2025-06-14
Client Server URLs: validate the `default`'s value is found in the `enum`


jamietanna:
Likely as part of https://github.com/oapi-codegen/oapi-codegen/issues/2006

PR #2045, opened 2025-07-25
Fix server variables conflict

### LINKED ISSUE #2003: Client Server URLs: support handling an enum with `default` as a value clashing with the default value constant
```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Server URLs can be optionally generated
servers:
- url: https://{username}.gigantic-server.com:{port}/{basePath}
  description: The production API server
  variables:
    # TODO this conflict will cause broken generated code
    conflicting:
      enum:
        - 'default'
        - '443'
      default: 'default'
```

acm19:
Adds a property to opt-in avoiding the default value colliding with the value `default` for enum server variables. When set to true `Enum` is added to the variable name so that it doesn't clash with the default variable name. This behaviour is opt-in to avoid backward compatibility, but the default is actually a bug, so it make sense to change that default in the next major release.

Resolves: https://github.com/oapi-codegen/oapi-codegen/issues/2003

### DIFF
--- a/configuration-schema.json
+++ b/configuration-schema.json
@@ -110,7 +110,11 @@
         },
         "preserve-original-operation-id-casing-in-embedded-spec": {
           "type": "boolean",
-          "description": "When `oapi-codegen` parses the original OpenAPI specification, it will apply the configured `output-options.name-normalizer` to each operation's `operationId` before that is used to generate code from.\nHowever, this is also applied to the copy of the `operationId`s in the `embedded-spec` generation, which means that the embedded OpenAPI specification is then out-of-sync with the input specificiation.\nTo ensure that the `operationId` in the embedded spec is preserved as-is from the input specification, set this. NOTE that this will not impact generated code.\nNOTE that if you're using `include-operation-ids` or `exclude-operation-ids` you may want to ensure that the `operationId`s used are correct."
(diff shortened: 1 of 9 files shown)


jamietanna:
Nice, thank you!

configuration-schema.json
jamietanna, line 1: Can we please revert the unrelated changes? Although those full stops at the end may be reasonable, I'd prefer to remove them from this PR to keep it atomic. Thanks!

acm19: Sure, that's done. I did leave a typo fix.

acm19:
Hey @jamietanna, bumping this in your inbox so it doesn't get forgotten. I've replied to all your comments so far.

mromaszewicz:
Thank you for submitting this PR, @acm19 — but we need to close it because the underlying bug (#2003) was just fixed via a different approach in #2358, merged as commit `08b30183`.

Your diagnosis was correct: an enum value of `default` collides with the named default-pointer constant. Your PR fixed it by adding an opt-in `Enum` infix to enum-value identifiers; the merged fix instead renames the default-pointer constant from `<Prefix>Default` to `<Prefix>DefaultValue` *only* when the spec actually contains a colliding enum value, so adopters whose specs compile cleanly under the current codegen see no identifier change at all. Same root cause, different namespace separation — and yours pointed us at the right shape of solution.

The merged PR also picks up four other bugs in the server-URLs feature (#2004 declared-but-unused variables generating dead parameters, #2005 undeclared placeholders making the function permanently error, #2006 missing `Valid()` method, #2007 default not in enum producing broken Go) that share the same root cause: the server-URLs template re-implemented its own type/enum emission instead of going through the generic codegen pipeline.

Apologies for the long wait. Thanks for the contribution.

PR #2047, opened 2025-08-02
feat(client): add abbility to set custom marshal/unmarshal func for client

sterligov:
In this pull request I added the ability to set custom marshal/unmarshal functions for the client.
This functionality is enabled via the client-custom-marshaler-funcs option.

The default marshal/unmarshal function of json is not the most optimal, I want to be able to marshal/unmarshal for example using the [sonic](https://github.com/bytedance/sonic) package


jamietanna:
Thanks - will review in more depth at some point, but the immediate thought I have is that this should be an opt-in config via `output-options` - see https://github.com/oapi-codegen/oapi-codegen/blob/main/CONTRIBUTING.md#making-changes-that-tweak-generated-code for a bit more info

Thanks!

sterligov:
> Thanks - will review in more depth at some point, but the immediate thought I have is that this should be an opt-in config via `output-options` - see https://github.com/oapi-codegen/oapi-codegen/blob/main/CONTRIBUTING.md#making-changes-that-tweak-generated-code for a bit more info
> 
> Thanks!

I added a flag `client-custom-marshaler-funcs` that, when enabled, generates additional code that allows you to set custom marshalling functions

PR #2056, opened 2025-08-13
build: test against Go 1.25

jamietanna:


jamietanna:
Hmm, given these failures:

```
# golang.org/x/tools/internal/tokeninternal
Error: ../../../../go/pkg/mod/golang.org/x/tools@v0.21.1-0.20240508182429-e35e4ccd0d2d/internal/tokeninternal/tokeninternal.go:64:9: invalid array length -delta * delta (constant -256 of type int64)
```

This may be blocked by https://github.com/oapi-codegen/oapi-codegen/pull/1641 and / the `/x/ packages needing an update, but as it pulls https://github.com/oapi-codegen/oapi-codegen/pull/1972 it then needs to bump the `go` directive, which we prefer to avoid where possible

jamietanna:
https://github.com/golang/go/issues/74462

PR #2062, opened 2025-08-19
fix: getting cookie in fiber middleware

YATAHAKI:
Closes: #1752 

Related to: https://github.com/oapi-codegen/oapi-codegen/pull/1751


pkg/codegen/templates/fiber/fiber-middleware.tmpl
```diff
@@ -127,7 +127,7 @@ func (siw *ServerInterfaceWrapper) {{$opid}}(c *fiber.Ctx) error {
     {{range .CookieParams}}
       var cookie string
 
-      if cookie = c.Cookies("{{.ParamName}}"); cookie == "" {
+      if cookie = c.Cookies("{{.ParamName}}"); cookie != "" {
```
mromaszewicz, line 130: Oh wow, nice find.

PR #2071, opened 2025-09-02
Support unions with multiple mappings pointing to a single underlying type

tobio:
Fixes https://github.com/oapi-codegen/oapi-codegen/issues/1530
Fixes https://github.com/oapi-codegen/oapi-codegen/issues/1664

This PR:
* Correctly generates the `ValueByDiscriminator` function when multiple discriminator values map to a single type
* Prevents `From*/Merge*` functions from setting an arbitrary discriminator in the case where there isn't a 1:1 mapping between type and discriminator value


tobio:
@jamietanna are you able to take a look at this one?

tobio:
@mromaszewicz are you able to take a look at this?

mromaszewicz:
This looks good, I had to think about it for a while, since this code path is so fragile. Thanks for doing this!

PR #2075, opened 2025-09-03
fix(client): correctly prepare requests when `omitZero` is set on fields 

### LINKED ISSUE #2074: Generated clients fail to invoke `New...Request` with `value is a nil pointer` when using `prefer-skip-optional-pointer-with-omitzero`
> Also, if you generate clients with this:
> ```yaml
>   prefer-skip-optional-pointer: true
>   prefer-skip-optional-pointer-with-omitzero: true
> ```
> and add this to a parameter like this:
> ```yaml
> - name: parentTag
>   in: header
>   description: Version string of parent model, if any model
>   x-go-type-skip-optional-pointer: false
>   required: false
>   schema:
>     type: string
>     example: parentTag
> ```
> 
> Client call will fail with `value is a nil pointer` error. If you remove `x-go-type-skip-optional-pointer` from the field, the client adds the parameters to the query even if they are zero-values, resulting in a 400 on the server side. 

 _Originally posted by @juan-carvajal in [#2044](https://github.com/oapi-codegen/oapi-codegen/issues/2044#issuecomment-3228751999)_

jamietanna:
Closes #2074.


mromaszewicz:
Closing — the underlying bug (parameter-level `x-go-type-skip-optional-pointer: false` overriding the global `prefer-skip-optional-pointer: true`) was fixed via #2329 (commit `44f8bb62`). A repro confirms client output now correctly handles both schema-level and parameter-level cases.

PR #2077, opened 2025-09-04
fix: handle text refs in strict echo handler

### LINKED ISSUE #2076: Strict server generates error for text/plain response types
When generating strict server  using version 2.5.0 and respons type of text/plain the generated code looks something like this (for echo):

```
type InternalServerErrorTextResponse string

type DeleteRequest500TextResponse struct {
	InternalServerErrorTextResponse
}

func (response DeleteRequest500TextResponse) VisitDeleteRequestResponse(w http.ResponseWriter) error {
	w.Header().Set("Content-Type", "text/plain")
	w.WriteHeader(500)

	_, err := w.Write([]byte(response))
	return err
}
```
This does not compile with an error  cannot convert response (variable of struct type DeleteRequest500TextResponse) to type []byte

mtandersson:
Fixes a problem where the server generated code did not compile for strict echo.

Also adds a test case to trigger the problem and then verify the update

fixes #2076

### DIFF
--- a/internal/test/strict-server/chi/server.gen.go
+++ b/internal/test/strict-server/chi/server.gen.go
@@ -531,6 +531,8 @@ type ReusableresponseJSONResponse struct {
 	Headers ReusableresponseResponseHeaders
 }
 
+type UnauthorizedTextResponse string
+
 type JSONExampleRequestObject struct {
 	Body *JSONExampleJSONRequestBody
 }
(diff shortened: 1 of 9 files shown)


mromaszewicz:
Thank you for submitting this PR, @mtandersson — but we need to close it because the same fix (extending the Multipart special-case in `strict-interface.tmpl` to also cover Text responses, generating a named type alias instead of a struct embedding) was merged via #2225 (commit `091742e8`). A repro confirms `$ref` text/plain responses now compile. Issue #2190 was the canonical duplicate that drove the merged fix.

Thanks for the contribution.

PR #2083, opened 2025-09-09
docs(sponsors): remove Elastic as a sponsor

jamietanna:
As I no longer work there, I'm no longer benefiting from the 4hr/month
benefits that I had in the past.

Conversations are still ongoing internally as to whether they'll become
a 1hr/month sponsor.


jarangutan:
Fingers crossed you get that 1/hr a month <3

You're awesome for all the goodness you bring

PR #2086, opened 2025-09-12
Allow Struct Embedding with `allOf`

clarkjohnd:
A rebase of 2 year old PR (https://github.com/oapi-codegen/oapi-codegen/pull/1295) (main code courtesy of Lahabana!) to the latest main branch with a very small tweak to error handling. This is a feature I find very useful. Also addresses https://github.com/oapi-codegen/oapi-codegen/issues/1622#issue-2310736019, where the original draft PR was linked.

Using one of my own app examples, adding `x-go-allof-embed-refs: true` to `allOf` components:

```yaml
    WorkItemView:
      x-go-allof-embed-refs: true
      allOf:
      - $ref: '#/components/schemas/WorkItem'
      - type: object
        properties:
          project:
            $ref: '#/components/schemas/ProjectSummary'
          organisation:
            $ref: '#/components/schemas/OrganisationSummary'
```

```go
type WorkItemView struct {
    // all fields duplicated from root component WorkItem
    // ...
    // ...
    Organisation *OrganisationSummary `json:"organisation,omitempty"`
    Project      *ProjectSummary      `json:"project,omitempty"`
}
```

to:

```go
type WorkItemView struct {
    // Embedded struct due to allOf(#/components/schemas/WorkItem)
    WorkItem `yaml:",inline"`
    // Embedded fields due to inline allOf schema
    Organisation *OrganisationSummary `json:"organisation,omitempty"`
    Project      *ProjectSummary      `json:"project,omitempty"`
}
```

This helps build reusable functions for "extendable" types - e.g. I have a function to populate `WorkItem` - now if I extend work item to add other fields for other API paths (like above example), I can reuse the same function to manage the `WorkItem` nested struct, rather than rebuilding an entire new function for the extended type.


marcoabreu:
Would love to see this merged, good job!

mromaszewicz:
@greptileai

mromaszewicz:
This is a very tricky change, despite seeming simple, and I think we need lots more testing to ensure it doesn't break behavior.

When you embed structs, you also embed their methods.

In your example, say, `WorkItem` is a complex schema with a complex structure that requires implementing the `json.Marshaler` and `json.Unmarshaler` interfaces, which we do automatically.


Due to this embedding, when the json library is doing its unarshaling, it will find that `WorkItemView` has JSON marshaling specializations, and it will call those, so it will attempt to unmarshal `WorkItemView` using `WorkItem.UnmarshalJSON`.

I've hit these problems before.

Dynamic typing in a stricly typed language is quite hard, and allOf/anyOf/oneOf has been the most difficult thing to support in oapi-codegen.

PR #2087, opened 2025-09-12
Bug Report: demonstration of overlays and allOf not playing together nicely

### LINKED ISSUE #2335: x-go-type on a schema with allOf is silently ignored; first member's x-go-type leaks through instead
## Summary

When a schema has both `allOf` and `x-go-type`, the `x-go-type` directive on the schema itself is silently ignored. Worse, if any `allOf` member has its own `x-go-type`, that override leaks through and becomes the alias target for the outer schema — producing a nonsensical alias that drops fields the user expects.

This was originally surfaced as an overlay issue in https://github.com/oapi-codegen/oapi-codegen/pull/2087, but the bug is **unrelated to overlays**. Overlays merely make it easy to add `x-go-type` to a third-party spec; the same misbehavior reproduces with a hand-edited spec containing no overlay at all.

## Reproduction

Spec:

```yaml
components:
  schemas:
    Client:
      x-go-type: OverlayClient
      type: object
      required: [name]
      properties:
        name: { type: string }
    ClientWithId:
      x-go-type: OverlayClientWithId
      allOf:
        - $ref: '#/components/schemas/Client'
        - properties:
            id: { type: integer }
          required: [id]
```

User-provided Go types `OverlayClient{Name}` and `OverlayClientWithId{Id, Name}`.

Expected output:

```go
type Client       = OverlayClient
type ClientWithId = OverlayClientWithId
```

Actual output:

```go
type Client       = OverlayClient
type ClientWithId = OverlayClient   // wrong
```

PR #2087 contains a runnable demonstration in `examples/anyof-allof-oneof/`.

## Root cause

There are two compounding bugs.

**(1) `pkg/codegen/schema.go:335-342`** — The `allOf` branch in `GenerateGoSchema` returns before the `x-go-type` check at line 344:

```go
if schema.AllOf \!= nil {
    mergedSchema, err := MergeSchemas(schema.AllOf, path)
    ...
    return mergedSchema, nil   // early return; x-go-type never consulted
}

// Check x-go-type, which will completely override the definition ...
if extension, ok := schema.Extensions[extPropGoType]; ok { ... }
```

The code comment explicitly states `x-go-type` "will completely override the definition of this schema" — this contract is violated when the outer schema uses `allOf`.

**(2) `pkg/codegen/merge_schemas.go:113-116`** — `mergeOpenapiSchemas` unions extensions from both sides without filtering out Go-specific directives:

```go
result.Extensions = make(map[string]any, len(s1.Extensions)+len(s2.Extensions))
maps.Copy(result.Extensions, s1.Extensions)
maps.Copy(result.Extensions, s2.Extensions)
```

So `Client`'s `x-go-type: OverlayClient` is copied into the merged schema's extensions. `mergeSchemas` then recursively calls `GenerateGoSchema` on the merged schema (line 62), which this time has no `AllOf`, hits the `x-go-type` branch, and emits `type ClientWithId = OverlayClient`.

## Version

Reproduced on `main` as of PR #2087 (commit f64f92b).

nelzblooket:
Fixes: #2335

@nelzblooket reports:

Following [the instructions](https://github.com/oapi-codegen/oapi-codegen/blob/main/CONTRIBUTING.md#minimal-reproductions):

Version of `oapi-codegen` - from `main` pulled today:
```shell
$ go run github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen -version
github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen
(devel)
```

Go version:
```shell
$ go version
go version go1.24.0 darwin/arm64
```

Problem statement:
When doing an overlay of a type that has an `allOf` with a reference to another overlay, the subset overlay is being created as the alias in the generated file.

-----

I've added commits to the original problem report to move the test to a more appropriate place, and fix the resulting issues. This problem is a bug in how `allOf` handles the merging of extended properties, and the overlays aren't relevant directly.


examples/anyof-allof-oneof/anyofallofoneof.gen.go
```diff
@@ -25,10 +23,7 @@ type ClientOrIdentity struct {
 }
 
 // ClientWithId defines model for ClientWithId.
-type ClientWithId struct {
-	Id   int    `json:"id"`
-	Name string `json:"name"`
-}
+type ClientWithId = OverlayClient
```
nelzblooket: I would expect this to be `OverlayClientWithId`
```suggestion
type ClientWithId = OverlayClientWithId
```

mromaszewicz:
@greptileai

mromaszewicz (approved):

mromaszewicz:
@greptileai

mromaszewicz:
> ### Greptile Summary
> 
> This PR fixes a bug where `x-go-type` extensions on a base schema (e.g. applied via overlay) incorrectly bled into composed `allOf` schemas, causing the derived type to be aliased to the base's override type and silently dropping extra fields. The fix has two parts: (1) in `mergeSchemas`, detect the "decorator idiom" (inline extension-only `allOf` member) and, when absent, delete `x-go-type`/`x-go-type-name`/`x-go-type-import` from the merged schema's extensions; (2) in `GenerateGoSchema`, move the `x-go-type` check to run _before_ AllOf processing so that an explicit override on the outer schema always takes precedence. A regression test covering both the overlay alias path and the leaked-extension path is included.
> ### Confidence Score: 4/5
> 
> The core fix is correct and well-tested; two prior-thread edge-case concerns about decorator-idiom detection remain open and could reintroduce the bug in unusual schema shapes.
> 
> The primary regression (x-go-type leaking through allOf into derived types) is correctly fixed and covered by two targeted test cases. The selective delete of only type-identity extensions (x-go-type, x-go-type-name, x-go-type-import) is sound. The x-go-type early-return move in schema.go is logically correct. Score is 4 rather than 5 because the prior review thread raised two P1 concerns — false positives in isExtensionOnlySchema (structural fields like nullable not checked) and the decorator-idiom check inspecting m.Value instead of the SchemaRef wrapper — that have not been addressed and could cause the original aliasing bug to resurface on certain schema patterns.
> 
> pkg/codegen/merge_schemas.go — the isExtensionOnlySchema implementation and the decorator-idiom detection loop warrant attention before merge.
> ### Important Files Changed
> Filename 	Overview
> pkg/codegen/merge_schemas.go 	Core fix: adds decorator-idiom detection and selective deletion of x-go-type/x-go-type-name/x-go-type-import from merged schemas; prior-thread edge-case concerns about isExtensionOnlySchema false positives and m.Value inspection remain open.
> pkg/codegen/schema.go 	x-go-type check moved before AllOf handling so an explicit override on the outer schema always wins; logically correct and handles the ClientWithId overlay case properly.
> internal/test/extensions/allof-merge-extensions/overlays_test.go 	Adds two regression tests: one for the overlay alias path (ClientWithId → OverlayClientWithId) and one for the leaked-extension path (DerivedNoOverride must not be aliased to OverlayBaseOnly and must preserve the Extra field).
> internal/test/extensions/allof-merge-extensions/spec.yaml 	Well-designed spec covering both regression scenarios: direct overlay aliasing (Client/ClientWithId) and base-only extension leakage (BaseOnly/DerivedNoOverride).
> internal/test/extensions/allof-merge-extensions/overlay.yaml 	Overlay applies x-go-type to Client, ClientWithId, and BaseOnly; intentionally omits DerivedNoOverride to exercise the extension-leakage regression path.
> internal/test/extensions/allof-merge-extensions/overlays.go 	Hand-written overlay types; OverlayBaseOnly deliberately lacks an Extra field to make the aliasing bug a compilation error.
> internal/test/extensions/allof-merge-extensions/config.yaml 	Standard codegen config wiring the overlay and output file; no issues.
> internal/test/extensions/allof-merge-extensions/generate.go 	Standard go:generate directive; no issues.
> 
> Reviews (2): Last reviewed commit: ["Implement Greptile suggestions"](https://github.com/oapi-codegen/oapi-codegen/commit/abb60472d79ab32baa280ee2b7abc4376f513c16) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29517407)

@greptileai, i think you are incorrect, in that all your suggestions have been noted and fixed in my amended PR.

PR #2110, opened 2025-10-22
feat: #1536 add sprig template functions

phoenisx:
Resolves: #1536

This is a new PR for previously opened PR: https://github.com/oapi-codegen/oapi-codegen/pull/1582, integrating all go template functions from `sprig` library.


jamietanna:
Before I get a chance to properly review, can you please make sure the original author is given a `Co-authored-by` in the commit please? Given this is their work, it'd be good to make sure that's retained

phoenisx:
I would love to add `Co-authored-by`, but there are no changes from the original PR.

I raised this (and found an existing PR) today, because I was facing issues creating my own custom template and found there was an open discussion on `sprig` library.

~~In case you still want me to add them as co-author, please do let me know, will update my commit message~~

I updated the commit message 💯

PR #2150, opened 2025-11-03
chore(renovate): run `make tidy` after dependency updates to `go.mod`

jamietanna:
As we've had it allowlisted by Mend, we can now run `make tidy` to get
PRs automagically updated.

This will update the whole branch (in the case of multiple dependency
bumps on the branch) and only target Go-specific dependency changes,
avoiding unnecessary updates when doing i.e. `golangci-lint` updates.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -39,5 +39,15 @@
 			"matchCurrentVersion": "v0.0.0-00010101000000-000000000000",
 			"enabled": false
 		}
-	]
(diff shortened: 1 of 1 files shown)


jamietanna:
Hmm, this is leading to https://github.com/oapi-codegen/oapi-codegen/pull/2153#issuecomment-3504929037

PR #2175, opened 2025-12-17
fix: make it possible to have the same path parameter twice in URL

juniorrantila:
This makes it possible to parse the keycloak openapi spec, which for one API has a URL of '/admin/realms/{realm}/clients/{client-uuid}/roles/{role-name}/composites/clients/{client-uuid}'.

### DIFF
--- a/pkg/codegen/template_helpers.go
+++ b/pkg/codegen/template_helpers.go
@@ -53,10 +53,15 @@ func genParamArgs(params []ParameterDefinition) string {
 	if len(params) == 0 {
 		return ""
 	}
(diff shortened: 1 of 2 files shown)


jamietanna:
Related: https://github.com/oapi-codegen/oapi-codegen/issues/1574

mromaszewicz:
Thanks for doing this, @juniorrantila , however, this isn't the right fix for the problem. The problem needs to be fixed at the time the schemas for the parameters are gathered for each path. I'm closing this commit, and added you as a Co-Author on a new one, so you get some credit for finding this.

PR #2178, opened 2025-12-29
#2103 fix fiber template with header params

### LINKED ISSUE #2103: Cannot use 'value' (type []string) as the type string
Error:
Cannot use 'value' (type []string) as the type string

The Problem is that the generated code line:

`if value, found := headers[http.CanonicalHeaderKey("Authorization")]; found {`

The value is the **wrong type**.

**yml:**

`  /upload:
    post:
      summary: Direct storage upload
      description: >-
        Uploads an image file directly to storage using a time-limited upload token.
        Use the `upload_url` returned by `POST /upload/sessions` as the target URL.
      operationId: postDirectUpload
      tags:
        - Upload
      servers:
        - url: http://localhost:8080/api/internal
          description: Internal storage upload server (development)
        - url: https://pixelfox.cc/api/internal
          description: Internal storage upload server (production)
      security:
        - UploadTokenAuth: []
      parameters:
        - name: Authorization
          in: header
          description: "Bearer <UPLOAD_TOKEN>. Alternative: submit as multipart field 'token'"
          required: false
          schema:
            type: string
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                file:
                  type: string
                  format: binary
                token:
                  type: string
                  description: "Alternative to Authorization header: pass upload token as multipart field"
              required: [file]
      responses:
        '200':
          description: Upload accepted
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/StorageUploadResponse'
        '400': { $ref: '#/components/responses/BadRequest' }
        '401': { $ref: '#/components/responses/Unauthorized' }
        '413': { $ref: '#/components/responses/PayloadTooLarge' }
        '415': { $ref: '#/components/responses/UnsupportedMediaType' }
        '429': { $ref: '#/components/responses/TooManyRequests' }
        '500': { $ref: '#/components/responses/InternalError' }`

**Generated Code:**

`// PostDirectUpload operation middleware
func (siw *ServerInterfaceWrapper) PostDirectUpload(c *fiber.Ctx) error {

	var err error

	c.Context().SetUserValue(UploadTokenAuthScopes, []string{})

	var params PostDirectUploadParams

	headers := c.GetReqHeaders()

	// ------------- Optional header parameter "Authorization" -------------
	if **value**, found := headers[http.CanonicalHeaderKey("Authorization")]; found {
		var Authorization string

		err = runtime.BindStyledParameterWithOptions("simple", "Authorization", value, &Authorization, runtime.BindStyledParameterOptions{ParamLocation: runtime.ParamLocationHeader, Explode: false, Required: false})
		if err != nil {
			return fiber.NewError(fiber.StatusBadRequest, fmt.Errorf("Invalid format for parameter Authorization: %w", err).Error())
		}

		params.Authorization = &Authorization

	}

	return siw.Handler.PostDirectUpload(c, params)
}`

LimitR:
Fix #2103
This fix will be very helpful for the latest versions of library dependencies. I've tested everything; this should have been fixed a long time ago, as it's the second time I've encountered this issue.


mromaszewicz:
Thank you for sending this PR, however I merged an earlier one (#1419) which also fixes this issue, and it supersedes this PR.

PR #2179, opened 2026-01-02
sample service scaffolding added

sajjanjyothi:
oapi-codegen --service=servicename will create a bare minimal backend service scaffolding which devs can use to create a common folder structure service.


pkg/codegen/sample_service/sample_service_app/internal/pkg1/pkg1.go
LimitR, line 1: normal filename? (and package name)

sajjanjyothi: just a sample package, let me change to something meaningful

sajjanjyothi: done , changed

PR #2181, opened 2026-01-09
Update enum generation to generate a map to validate enums.

AndreasAbdi:
Customers often need to validate whether a value is a legit enum value. 
We implement that validation by creating helper functions.

With the helper function, customers no longer have to manually write their own version of the helper.

i.e. 
```
const (
	ClientTypeWithNamesExtensionActive  ClientTypeWithNamesExtension = "ACT"
	ClientTypeWithNamesExtensionExpired ClientTypeWithNamesExtension = "EXP"
)

var ClientTypeWithNamesExtensionValues = map[ClientTypeWithNamesExtension]struct{}{
	ClientTypeWithNamesExtensionActive:  {},
	ClientTypeWithNamesExtensionExpired: {},
}

func (s ClientTypeWithNamesExtension) IsValid() bool {
	_, ok := ClientTypeWithNamesExtensionValues[s]
	return ok
}
```
and then the customer uses it like: 
```
func legit() {
	k := ClientTypeWithNamesExtension("")
	isValid := k.IsValid()
	fmt.Print(isValid)
}
```

### DIFF
--- a/examples/extensions/xenumnames/gen.go
+++ b/examples/extensions/xenumnames/gen.go
@@ -9,17 +9,47 @@ const (
 	EXP ClientType = "EXP"
 )
 
+var ClientTypeValues = map[ClientType]struct{}{
+	ACT: {},
+	EXP: {},
+}
+
+func (s ClientType) IsValid() bool {
+	_, ok := ClientTypeValues[s]
+	return ok
+}
+
 // Defines values for ClientTypeWithNamesExtension.
 const (
 	ClientTypeWithNamesExtensionActive  ClientTypeWithNamesExtension = "ACT"
 	ClientTypeWithNamesExtensionExpired ClientTypeWithNamesExtension = "EXP"
 )
 
+var ClientTypeWithNamesExtensionValues = map[ClientTypeWithNamesExtension]struct{}{
+	ClientTypeWithNamesExtensionActive:  {},
+	ClientTypeWithNamesExtensionExpired: {},
+}
+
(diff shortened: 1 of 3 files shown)


jamietanna (commented):
I'd say we probably shouldn't export the map as then it's mutable outside of the package? 

It's probably also worth making this opt-in, via `output-options`, so it's something that reduces generated code changes

mromaszewicz:
Can your problem be solved via request validation middleware? For example, 
- https://github.com/getkin/kin-openapi/tree/master/openapi3filter
- https://github.com/pb33f/libopenapi-validator

In all the servers that I've written, I've left the problem of validating inputs against schemas to other, excellent projects that already exist.

I think oapi-codegen should be as narrowly focused as a code generator as possible.

jamietanna:
While it could be, Marcin, I've found myself wanting something like this in the past - i.e. to add a unit test to clarify that all of my business layer states can correctly map to the state in the HTTP layer's enums

Related: https://github.com/oapi-codegen/oapi-codegen/issues/1537 / https://github.com/oapi-codegen/oapi-codegen/issues/1721

mromaszewicz:
What do you guys think of doing it like this? https://github.com/oapi-codegen/oapi-codegen/pull/2227

We don't need a map, we've got a code generator that can generate a nice Valid() function.

mromaszewicz:
Thanks for this PR, but I am closing it, because we have an alternate fix already merged. The Enum types have a Validate function now, which just uses a switch statement to check for valid values.

PR #2184, opened 2026-01-14
Fix exploded parameter marshaling and binding inconsistencies

TelpeNight:
I've added tests for https://github.com/oapi-codegen/oapi-codegen/issues/2183
But maybe adding more test cases would be helpful. It is pretty generic.

SpecQuery can be commented our to prevent comparison. In this case first test will pass, the second will fail.


mromaszewicz:
Please have a look at https://github.com/TelpeNight/oapi-codegen/pull/1, I sent you a PR with the fix. WE can merge all in one once you approve.

There's a limitation dependent on a runtime bug. I'll fix that separately.

mromaszewicz:
Oh, and you will need to rebase for CI to pass.

mromaszewicz:
@greptileai

mromaszewicz (dismissed):

mromaszewicz:
Could you allow maintainers to push changes to your branch in the PR settings so that I can resolve the conflicts and merge, or could you do a rebase and fix them yourself?

TelpeNight:
Hello. I've opened access to maintainers.

TelpeNight:
I've tried to resolve conflicts myself. But after regenerating tests, some of them start to fail.
For example `internal/test/issues/issue-2031/prefer/issue2031_test.go`
Old version did encode `user_ids[]` as "user_ids%5B%5D". This seems to be correct.
https://www.codestudy.net/blog/is-array-syntax-using-square-brackets-in-url-query-strings-valid/
https://swagger.io/specification/#percent-encoding-and-form-urlencoded

According to the specification `[]` should be encoded even if allowReserved is true.

New implementation seems to violate this. So I don't think it is ready to be merged.

mromaszewicz:
I resolved the conflicts, but we've caused a behavior regression with these changes (the CI will fail). I'm digging into it more. There is a behavioral change in how parameters are encoded now `user_ids[]=1` isn't RFC compliant, but widely accepted in various tools, while `user_ids%5B%5D=1` is spec compliant.

mromaszewicz:
I pushed a fix which still encodes parameter names, but honestly, I am unsure if this is proper behavior.

There's also an OpenAPI parameter property, `allowReserved` which we are ignoring.

This code is also very old with an `.IsStyled` check in the template, which was added after basic parameters were handled. Honestly, everything should go through through the `runtime.StyleParameter`. I think I need to do some refactoring.

I've removed the approval for this PR for the moment while I think about this.

TelpeNight:
I believe that the better approach is to be spec compliant. Even `allowReserved` doesn't allow `gen-delims`. They should be escaped. Just like `&` character from the same set, as it would screw a query up.

TelpeNight:
After studding the runtime source a little bit, I think that it lacks a good test suit for serialisation/deserialisation process. The case I've started this issue from comes directly from oapi spec. Nothing extra-ordinal to be honest.

mromaszewicz (approved):

PR #490, opened 2021-11-26
158 Generate Nullable items in arrays

wjase:
Fix for Issue 158

PR #2185, opened 2026-01-20
Support nullable slice elements and map values

### LINKED ISSUE #158: Checking for the *[] vs []*
when we try to nest a struct 
e.g. :
response{
   total,
   orderby,
    []data
}

``` 
total:
    type: integer
    format: int32  
data:
    type: array
    items:
      $ref: searchcomponent.yml
```

code genererated is
```
type SearchResponse struct {
	Data *[]struct {
		// Embedded struct due to allOf(#/components/schemas/CommonParent)
		someparentstruct
		// Embedded fields due to inline allOf schema
		Uri *string `json:"uri,omitempty"`
	} `json:"data,omitempty"`
	Total *int32 `json:"total,omitempty"`
}
```

above *[]struct should be []*struct

iamtakingiteasy:
Adds support for nullable slice elements, e.g. `[]*string`, if `items` are declared as `nullable: true` (analogous to `map[string]*string` if `additionalProperties` is having `nullable: true` and to fields describing/referencing nullable types), allows handling of values like `["a", null, null, "d", ""]` and occasionally improves the UX with DTO processing.

Closes #158 
Based on and re-implements #158-specific portion of #490, also with tests for the map case and none of the unrelated changes.


iamtakingiteasy:
Sorted testcases by nullability.

mromaszewicz:
For nullable types, we have a `Nullable[T]`, so I wonder if it makes sense to use Nullables or pointers. I see that someone added nullable support to maps via the pointer type, so this would be similar to that, while we use the Nullable wrapper in other cases. I don't like that we have two ways of doing this. It's too late now, though.

mromaszewicz (approved):

mromaszewicz:
Once you rebase and CI runs, this should be submittable.

iamtakingiteasy:
Regarding, `Nullable[T]`, it may be useful for isolated values, allowing to distinguish the tri-state, but there is still the need to represent regular slices and maps where bi-state suffices and requires no overhead, allowing to make use of generated models in regular go code, where mutability of slice/map elements may be desired, which may be the case even during request/response construction.

jamietanna:
This appears to cause a regression - https://github.com/rootlyhq/rootly-go/pull/41 is now receiving i.e.

```diff
 type Alert struct {
        // AlertFieldValuesAttributes Custom alert field values to create with the alert
-       AlertFieldValuesAttributes []struct {
+       AlertFieldValuesAttributes []*struct {
```

When using `prefers-skip-optional-pointer`

jamietanna:
As discussed in https://github.com/oapi-codegen/oapi-codegen/issues/2240, this wasn't actually an issue for Rootly's API

However, I don't know if this behaviour is correct when using the `nullable.Nullable` type? 

Edit: I've raised #2242

PR #2204, opened 2026-02-07
Use "any" instead of "interface{}"

mromaszewicz:


jamietanna (changes requested):
Please raise against https://github.com/oapi-codegen/oapi-codegen-exp

PR #2205, opened 2026-02-07
build: use a re-usable, single, workflow for running CI

jamietanna:
As a way to keep updates to the CI pipelines more straightforward, we
can extract this out to a shared, versioned (and updated by Renovate)
Action.

We can make sure to keep our binary builds going, but only for the
current versions of Go - other build failures will be caught by our
shared Action.


jamietanna:
cc @oapi-codegen/maintainers - all new PRs will need to make sure they're passing this new CI check, i.e. by updating them from `main`

PR #2211, opened 2026-02-10
feat: add support for custom package alias for external ref imports

InventivetalentDev:
Adds support for specifying a package alias for external imports, which allows using custom names instead of auto-generated aliases like `externalRef0`, `externalRef1`, ...
This allows for more consistent import names when dealing with large projects and multiple linked specs, where the same `externalRef0` import can end up referencing different source packages.

## Motivation 
I'm using `oapi-codegen` in a complex project in combination with [`tygo`](https://github.com/gzuidhof/tygo) to generate TS types for my frontend code. `tygo` also supports external imports, but they need to be explicitly mapped from the Go package to the TS file, so my config for that currently looks something like this:
```yaml
type_mappings:
  time.Time: "string /* RFC3339 */"
  null.String: "null | string"
  null.Bool: "null | boolean"
  uuid.UUID: "string /* uuid */"

  externalRef0.User: User
  externalRef1.Project: Project
  externalRef1.Repo: Repo
  externalRef2.Identity: Identity
  // same source package but different ref index for a spec that doesn't import User or Project, leading to duplicate type mapping
  externalRef0.Identity: Identity 
```
needing to map the same package/type multiple times since different files have varying starting indexes for externalRef imports. 
Using the custom alias allows me to simply do this instead:
```yaml
# config.yaml
import-mapping:
  ./common/users.yaml: users example_package/users
  ./common/projects.yaml: projects example_package/projects
  ./common/identities.yaml: identities example_package/identities
```

```yaml
# tygo.yaml
  users.User: User
  projects.Project: Project
  projects.Repo: Repo
  identities.Identity: Identity
```
  

_previous implementation_: https://github.com/oapi-codegen/oapi-codegen/pull/2193/

### DIFF
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -90,7 +90,7 @@ func (im importMap) GoImports() []string {
 
 func constructImportMapping(importMapping map[string]string) importMap {
 	var (
-		pathToImport = map[string]goImport{}
+		pathToImport = importMap{}
 		result       = importMap{}
 	)
 


mromaszewicz:
Your CI failure has nothing to do with your code. 
https://github.com/oapi-codegen/oapi-codegen/issues/2214

mromaszewicz (approved):

mromaszewicz:
We will merge this once I get my CI fix approved, and you rebase to run precommit using that fix.

mromaszewicz:
Thank you, @InventivetalentDev

Issue #1474, opened 2024-02-24
Redeclared Types when model name matches client endpoint id response name


hackerrdave:
When generating `types,client` for the [openai openapi spec](https://github.com/openai/openai-openapi/tree/6b64280c3db0082cbafa34495b9f3a3a58eb960d), I see many errors similar to this example:

Duplicate types
```golang
/ CreateChatCompletionResponse Represents a chat completion response returned by model, based on the provided input.
type CreateChatCompletionResponse struct {
	// Choices A list of chat completion choices. Can be more than one if `n` is greater than 1.
	Choices []struct {
		// FinishReason The reason the model stopped generating tokens. This will be `stop` if the model hit a natural stop point or a provided stop sequence,
		// `length` if the maximum number of tokens specified in the request was reached,
		// `content_filter` if content was omitted due to a flag from our content filters,
		// `tool_calls` if the model called a tool, or `function_call` (deprecated) if the model called a function.
		FinishReason CreateChatCompletionResponseChoicesFinishReason `json:"finish_reason"`

		// Index The index of the choice in the list of choices.
		Index int `json:"index"`

                //etc...
}

//...

type CreateChatCompletionResponse struct {
	Body         []byte
	HTTPResponse *http.Response
	JSON200      *CreateChatCompletionResponse
}
```

Error:
```
CreateChatCompletionResponse redeclared in this block 
```

I believe this occurs when the schema for an operation's response has the same name as the generated name for <OperationId>Response, this is the yaml in question:

```yaml
/chat/completions:
    post:
      operationId: createChatCompletion
      tags:
        - Chat
      summary: Creates a model response for the given chat conversation.
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/CreateChatCompletionRequest"
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/CreateChatCompletionResponse"
```

Is there an expected flag to use to workaround this? Or is this a bug?

pvanbuijtene:
I had the same and would see this as a bug, end-users should not have to worry about this.
You can work around it by setting a response type suffix in you config, like this:
```
output-options:
  response-type-suffix: Resp
```

jamietanna:
Interesting! Yeah this looks like a bug. Glad there's a workaround in the meantime - I'll get it officially documented in #1254 - but an ideal solution would be to detect when there may be a clash, and rename the generated response type

[...]

Issue #2097, opened 2025-09-24
Generated path responses can conflict with schema definitions


joshprzybyszewski-wf:
_Example: https://github.com/oapi-codegen/oapi-codegen/pull/2096_

I noticed when using this generator that when attempting to use the "client with response", the returned type is busted. In particular, it's because there are two `type`s with the same name: one seems to be defined in the schemas and one seems to be generated from the `paths` definition.

mromaszewicz:
This was addressed in #2213, please enable name conflict resolution via the output options and file a new issue if it doesn't work for you

Issue #1881, opened 2025-01-12
failed to generate code when defined components/requestBodies have multiple content types


ibbema:
Generating code with components/requestBodies  xtype which have multiple content types are failing with 
error generating code for type definitions: duplicate typename xtype

providing example here server-types and openapi spec in attachment 
running command 
 **oapi-codegen  -config server-types.yaml patch-me-api.yaml** 
give me error 
**error generating code: error generating type definitions: error generating code for type definitions: duplicate typename 'AddressRequest' detected, can't auto-rename, please use x-go-name to specify your own name for one of them**

[exampefiles.zip](https://github.com/user-attachments/files/18387760/exampefiles.zip)

mromaszewicz:
This was addressed in #2213, please enable name conflict resolution via the output options and file a new issue if it doesn't work for you

Issue #1713, opened 2024-07-31
CreateBlueprintResponse redeclared in this block


lzap:
Hello,

I am hitting this error

```
internal/clients/http/image_builder/client.gen.go:3631:6: CreateBlueprintResponse redeclared in this block
```

when I try to compile a client generated from this YAML:

https://gist.github.com/lzap/55d2144c19d85b1072b6d32cfbe9a6c9

```
$ bin/oapi-codegen --version
github.com/deepmap/oapi-codegen/v2/cmd/oapi-codegen
v2.0.0
```

Can someone explain to me why am I hitting this? Apparently, renaming the component does help.

jamietanna:
It's likely a conflict between https://gist.github.com/lzap/55d2144c19d85b1072b6d32cfbe9a6c9#file-image-builder-yaml-L978 and the autogenerated type from https://gist.github.com/lzap/55d2144c19d85b1072b6d32cfbe9a6c9#file-image-builder-yaml-L145

I can't remember offhand what the best course of action is - renaming the `CreateBlueprintResponse` model _should_ do it though :thinking:

mromaszewicz:
This was addressed in #2213, please enable name conflict resolution via the output options and file a new issue if it doesn't work for you

PR #2213, opened 2026-02-11
feat: multi-pass type name resolution

mromaszewicz:
I've been meaning to use this approach for a long time, because the attempts at avoiding type collisions via structure suffixes or prefixes work sporadically, at best.

Conflict resolution is fundamentally a global problem, not a local problem when doing recursive traversal, so this PR splits the code generation into two parts. First, the OAPI document structure is traversed, and all the schemas that we generate are gathered up into a list of candidates, then we do global conflict resolution across the space of all schemas. This allows us to preserve the functionality of things affected by schema name - `$ref`, `required` properties, and so forth.

This fixes issue #1474 (client response wrapper type colliding with a component schema of the same name and improves issue #200 handling (same name across schemas, parameters, responses, requestBodies, headers).

The new system is gated behind the existing `resolve-type-name-collisions` output option. When disabled, behavior is unchanged, oapi-codegen exits with an error. This flag is default false, so there is no behavior change to oapi-codegen unless it's specified. All current test files regenerate without any differences.

Added a comprehensive test which reproduces the scenarios in all the PR's and Issues below, and adds a few more, to make sure that references to renamed targets are also correct..

Key changes:
- gather.go: walks entire spec collecting schemas with location metadata
- resolve_names.go: assigns unique names via context suffix, per-schema disambiguation, and numeric fallback strategies
- Component schemas are privileged and keep bare names on collision
- Client response wrapper types now participate in collision detection
- Removed ComponentType/DefinedComp from Schema struct
- Removed FixDuplicateTypeNames and related functions from utils.go

Obsoletes issues:
- #1474 Schema name vs client wrapper (CreateChatCompletionResponse)
- #1713 Schema name vs client wrapper (CreateBlueprintResponse)
- #1450 Schema name vs client wrapper (DeleteBusinessResponse)
- #2097 Path response type vs schema definition (Status)
- #255  Endpoint path vs response type (QueryResponse)
- #899  Duplicate types from response wrapper vs schema (AccessListResponse)
- #1357 Schema vs operationId response (ListAssistantsResponse, OpenAI spec)
- #254  Cross-section: requestBodies vs schemas (Pet)
- #407  Cross-section: requestBodies vs schemas (myThing)
- #1881 Cross-section: requestBodies with multiple content types

Obsoletes PRs:
- #292  Parameter structures params postfix (superseded by context suffix)
- #1005 Fix generate equals structs (superseded by multi-pass resolution)

EOF
)

### DIFF
--- a/internal/test/name_conflict_resolution/name_conflict_resolution.gen.go
+++ b/internal/test/name_conflict_resolution/name_conflict_resolution.gen.go
@@ -56,6 +56,16 @@ type Order struct {
 	Product *string `json:"product,omitempty"`
 }
 
+// Outcome defines model for Outcome.
+type Outcome struct {
+	Value *string `json:"value,omitempty"`
+}
+
+// Payload defines model for Payload.
+type Payload struct {
+	Content *string `json:"content,omitempty"`
+}
+
 // Pet defines model for Pet.
 type Pet struct {
 	Id   *int    `json:"id,omitempty"`
@@ -72,6 +82,11 @@ type Qux = CustomQux
 
 // CustomQux defines model for .
 type CustomQux struct {
+	Label *string `json:"label,omitempty"`
+}
+
+// SpecialName defines model for Renamer.
+type SpecialName struct {
 	Label *string `json:"label,omitempty"`
 }
 
@@ -131,8 +146,18 @@ type BarResponse struct {
 	Value2 *Bar2 `json:"value2,omitempty"`
 }
 
+// OutcomeResult defines model for Outcome.
+type OutcomeResult struct {
+	Result *string `json:"result,omitempty"`
+}
+
 // QuxResponse defines model for Qux.
 type QuxResponse struct {
+	Data *string `json:"data,omitempty"`
+}
+
+// Renamer defines model for Renamer.
+type Renamer struct {
 	Data *string `json:"data,omitempty"`
 }
 
@@ -162,6 +187,11 @@ type OrderRequestBodyJSON2 = []struct {
 // OrderRequestBodyJSON3 defines model for Order.
 type OrderRequestBodyJSON3 struct {
 	Product *string `json:"product,omitempty"`
+}
+
+// PayloadBody defines model for Payload.
+type PayloadBody struct {
+	Data *string `json:"data,omitempty"`
 }
 
 // PetRequestBody defines model for Pet.
@@ -212,6 +242,11 @@ type CreateOrderApplicationMergePatchPlusJSONBody struct {
 	Product *string `json:"product,omitempty"`
 }
 
+// SendPayloadJSONBody defines parameters for SendPayload.
+type SendPayloadJSONBody struct {
+	Data *string `json:"data,omitempty"`
+}
+
 // CreatePetJSONBody defines parameters for CreatePet.
 type CreatePetJSONBody struct {
 	Name    *string `json:"name,omitempty"`
@@ -238,6 +273,12 @@ type CreateOrderApplicationJSONPatchPlusJSONRequestBody = CreateOrderApplication
 // CreateOrderApplicationMergePatchPlusJSONRequestBody defines body for CreateOrder for application/merge-patch+json ContentType.
 type CreateOrderApplicationMergePatchPlusJSONRequestBody CreateOrderApplicationMergePatchPlusJSONBody
 
+// PostOutcomeJSONRequestBody defines body for PostOutcome for application/json ContentType.
+type PostOutcomeJSONRequestBody = Outcome
+
+// SendPayloadJSONRequestBody defines body for SendPayload for application/json ContentType.
+type SendPayloadJSONRequestBody SendPayloadJSONBody
+
 // CreatePetJSONRequestBody defines body for CreatePet for application/json ContentType.
 type CreatePetJSONRequestBody CreatePetJSONBody
 
@@ -246,6 +287,9 @@ type QueryJSONRequestBody QueryJSONBody
 
 // PostQuxJSONRequestBody defines body for PostQux for application/json ContentType.
 type PostQuxJSONRequestBody = Qux
+
+// PostRenamedSchemaJSONRequestBody defines body for PostRenamedSchema for application/json ContentType.
+type PostRenamedSchemaJSONRequestBody = SpecialName
 
 // PatchResourceJSONRequestBody defines body for PatchResource for application/json ContentType.
 type PatchResourceJSONRequestBody = ResourceMVO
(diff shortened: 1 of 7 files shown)


mitsaucepls:
Hello!
I tested this branch against the TMF622 ProductOrdering v5.0.0 spec with `resolve-type-name-collisions: true` and hit the following error:

```text
error generating code: error generating type definitions: error generating code for type definitions:
duplicate typename 'ProductOrderMVORequestBodyJSONRequestBodyJSONRequestBodyJSONRequestBodyJSONRequestBodyJSON'
detected, can't auto-rename, please use x-go-name to specify your own name for one of them
```

The TMF spec defines `ProductOrder_MVO` in both components/schemas and components/requestBodies.

The requestBody variant has multiple content types:
- application/json
- application/merge-patch+json
- application/json-patch+json
- application/json-patch-query+json

These are all valid, standardized media types.
The +json structured syntax suffix is also valid.

The problem is in contentTypeSuffix() it uses strings.Contains(ct, "json")
which matches all four types and maps them all to the same "JSON" suffix:

```go
case strings.Contains(ct, "json"):
    return "JSON"
```

This causes a loop in resolveCollisions.

I think using exact matches for the common media types instead of substring matching,
letting the +json subtypes fall through to mediaTypeToCamelCase, is a better way to go:

```go
func contentTypeSuffix(ct string) string {
    if ct == "" {
        return ""
    }
    ct = strings.ToLower(ct)
    switch ct {
    case "application/json":
        return "JSON"
    case "application/xml", "text/xml":
        return "XML"
    case "application/x-www-form-urlencoded", "multipart/form-data":
        return "Form"
    case "text/plain":
        return "Text"
    case "application/octet-stream":
        return "Binary"
    case "application/yaml", "application/x-yaml":
        return "YAML"
    default:
        return mediaTypeToCamelCase(ct)
    }
}
```

Additionally, I think strategyContextSuffix should track whether a context suffix has already been applied to prevent the loop.

What do you think about that? Thank you for your work!

mromaszewicz:
Thank you for finding this. i will add it to my test cases.

I'm trying my best to make these changes without breaking backward compatibility, and if I CamelCase-convert the full media type, it's going to change the code generation output for existing users who might not expect their JSON suffix to change to MergePatchJSON or something.

I'm back-porting these fixes from an experiment that I am doing, and I think the only thing I can really do is delegate the choice of converting the full media type to a short name to the user via configuration options.

Look here at the documentation for `content-type-short-names` configuration option:
https://github.com/oapi-codegen/oapi-codegen-exp/blob/main/experimental/Configuration.md

Whenever I make what I think are simple test cases that are representative, people in the real world have more complex usage :) I never really thought about returning multiple JSON models from a single API.

mitsaucepls:
Ah okay, the `content-type-short-names` declaration could also work, looks promising. 

Trying to fix that issue from above highlighted more issues I briefly described in my PR on your fork here: https://github.com/mromaszewicz/oapi-codegen/pull/1

mromaszewicz:
Thanks for doing that. This code, as is, should resolve your conflict by appending numerical suffixes to the end after all the collisions.

I think I'm doing the multi-pass wrong. Each strategy needs to be applied globally over all schemas before trying the next one.

mromaszewicz:
@mitsaucepls , please try this latest patch on your schema. I've added your specific example as a test case, and it now works. When, after all rounds of conflict resolution fail, the numeric fallback catches this, and you end up with `1` and `2` suffixes on the conflicting types.

The problem was that you can't try conflict resolution strategies on each conflict group, you must run each strategy globally until it makes no more changes, before moving onto the next one. When you rename things in one conflict group, you may actually rename them _into_ conflict with another schema.

This problem is very, very complicated, which is why I avoided it until now. Thanks for those PR's you are sending, but they're not the right fix for the issue because they can put the new names into conflict with other stuff.

This PR doesn't contain the short names map, that will be a separate pull request which will let you avoid the numerical suffixes and replace them with more meaningful content type abbreviations.

mitsaucepls:
Hey, it definitely generates now! But there are still errors because of undefined types.

I found that helper types are not generated and that causes the undefined type behavior.
So in my case only Responses are affected. I changed the following and all my errors were resolved.

```go
index 279ed9d..437af32 100644
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -695,6 +695,7 @@ func GenerateTypesForResponses(t *template.Template, responses openapi3.Response
                        }

                        types = append(types, typeDef)
+                       types = append(types, goType.AdditionalTypes...)
                }
        }
        return types, nil
```

Without knowing I can imagine other GenerateTypesFor... functions are also affected, just not in my specific case.

mromaszewicz:
Thanks @mitsaucepls . You found another old issue. Neither Responses or RequestBodies handle the additional types, so they miss generating some models.

mromaszewicz:
Please try it again. If that spec continues to have problems, could you attach it here? I couldn't find the OpenAPI file to download.

mitsaucepls:
I just realized I was on main for my last testing... sorry :D

This is the OpenAPI Yaml file Ive been using. [TMF622-ProductOrdering-v5.0.0.oas.yaml](https://github.com/user-attachments/files/25295632/TMF622-ProductOrdering-v5.0.0.oas.yaml) alternatively downloadable [here](https://www.tmforum.org/oda/open-apis/directory/product-ordering-management-api-TMF622/v5.0) but it requires an account.

This is the config.yaml I used so far:
```yaml
package: api
generate:
  echo-server: true
  strict-server: true
  models: true
output-options:
  resolve-type-name-collisions: true
output: gen.api.go
```

When using the fix/name-collisions branch there are still some errors. But fewer than before!
And on main there are, with the AdditionalTypes fix, none anymore.

mromaszewicz:
I'm going to split that AdditionalTypes fix out of this PR and into a new one, with a new issue, because they are independent problems.

mromaszewicz:
Man, that spec you attached is a doozy. It exposed a few more issues. Now, it generates cleanly with the latest update.

mromaszewicz:
@greptileai

mromaszewicz:
Yeah, you're right. This code path ignores the name overrides and the name normalizer function. It's an oversight on my part. 

I'm going to change it so that user provided names are "pinned", even if that causes conflict, we don't want to second-guess them, but other names will change around them.

Fixing the initialisms is easy. Another update, with test coverage for all these scenarios, is coming soon.

jamietanna:
I've regenerated what this looks like as [on-by-default](https://github.com/oapi-codegen/oapi-codegen/commit/6632b8adb75a9426d033766d9437f1da49011ff0) and the changes look good, as well as with rootly-go, thanks!

PR #2217, opened 2026-02-13
fix: escape quoted media type directives

brahmlower:
Fixes https://github.com/oapi-codegen/oapi-codegen/issues/1529

Added tests
- validating the new function for quoting the go string
- new "issues" test in internal/test/issues.

I appreciate yall keeping this project going- happy to make any requested changes!

### DIFF
--- a/internal/test/issues/issue-1529/strict-fiber/Makefile
+++ b/internal/test/issues/issue-1529/strict-fiber/Makefile
@@ -17,19 +17,20 @@ fi \
 endef
 
 lint:
-	$(GOBIN)/golangci-lint run ./...
+	$(call execute-if-go-124,$(GOBIN)/golangci-lint run ./...)
 
 lint-ci:
-	$(GOBIN)/golangci-lint run ./... --output.text.path=stdout --timeout=5m
+
+	$(call execute-if-go-124,$(GOBIN)/golangci-lint run ./... --output.text.path=stdout --timeout=5m)
 
 generate:
-	go generate ./...
+	$(call execute-if-go-124,go generate ./...)
 
 test:
-	go test -cover ./...
+	$(call execute-if-go-124,go test -cover ./...)
 
 tidy:
-	go mod tidy
+	$(call execute-if-go-124,go mod tidy)
 
 tidy-ci:
-	tidied -verbose
+	$(call execute-if-go-124,tidied -verbose)
--- a/internal/test/issues/issue-1529/strict-fiber/go.mod
+++ b/internal/test/issues/issue-1529/strict-fiber/go.mod
@@ -1,6 +1,6 @@
 module github.com/oapi-codegen/oapi-codegen/v2/internal/test/issues/issue-1529/strict-fiber
 
-go 1.25.0
+go 1.24.0
 
 replace github.com/oapi-codegen/oapi-codegen/v2 => ../../../../../
 


brahmlower:
And of course, as soon as I open the PR I realize I didn't actually enable the strict-server generation 🤦‍♂️

pkg/codegen/templates/strict/strict-interface.tmpl
```diff
@@ -74,7 +74,7 @@
                 {{if eq .NameTag "Multipart" -}}
                     writer := multipart.NewWriter(w)
                 {{end -}}
-                w.Header().Set("Content-Type", {{if eq .NameTag "Multipart"}}{{if eq .ContentType "multipart/form-data"}}writer.FormDataContentType(){{else}}mime.FormatMediaType("{{.ContentType}}", map[string]string{"boundary": writer.Boundary()}){{end}}{{else if .HasFixedContentType }}"{{.ContentType}}"{{else}}response.ContentType{{end}})
+                w.Header().Set("Content-Type", {{if eq .NameTag "Multipart"}}{{if eq .ContentType "multipart/form-data"}}writer.FormDataContentType(){{else}}mime.FormatMediaType("{{.ContentType}}", map[string]string{"boundary": writer.Boundary()}){{end}}{{else if .HasFixedContentType }}{{.ContentType | toGoString}}{{else}}response.ContentType{{end}})
```
brahmlower, line 77: This line is not easy to read. This change was as minimal as I could make it.

At first, I ignored the use of `{{.ContentType}}` in the call to `mime.FormatMediaType`, but now that I'm reviewing this more, the same sort of bug can arise when
- `eq .NameTag "Multipart"` is true, and
- `eq .ContentType "multipart/form-data"` is false

It's not immediately clear to me what part of the oapi spec the `NameTag` field refers to. I'll figure it out and include a fix and test later this evening.

brahmlower: I've looked into the use of `mime.FormatMediaType("{{.ContentType}}", ...`.

Testing locally, I found that changing it to use the `toGoString` function will result in function name collisions when multiple content types are defined. I managed to work around that by changing [these lines](https://github.com/oapi-codegen/oapi-codegen/blob/15c1dafe763ea2cfc0b9be0c136a79fb33bca2d6/pkg/codegen/operations.go#L821-L822) to:
```go
case strings.HasPrefix(contentType, "multipart/"):
	tag = mediaTypeToCamelCase(contentType)
```

However that caused name changes for multiple generated functions, which I think counts as a breaking change.

It seems unlikely that someone would have several possible multipart responses using quoted directives, so I don't think it's worth changing this `mime.FormatMediaType` call in this PR. Better to make that change in a separate PR, so greater focus can be given to possible breaking changes, if that's even something yall want to entertain.

mromaszewicz:
Make sure that `make test` and `make lint` work at the top level.

brahmlower:
I'm learning a lot about the projects CI setup this morning 😁

brahmlower:
Most of the failures seemed to be small things I missed with the tests. Like a Makefile being expected in the fiber tests directory because it had a go mod file, then correctly including the oapi-codegen tool definition in go mod, and then subsequently failing because matrix tests covered versions of go that don't support the tool directive. 

Definitely several silly mistakes on my part, but mostly test related.

I had hoped that rebasing via the github ui would keep the commit history somewhat cleaner, but looks like it represented rebase the same as if I had force pushed from my local. So, sorry about that

brahmlower:
I've really stumbled my way through the go version matrix tests- sorry about the mess of commits

PR #2219, opened 2026-02-13
Fix schema gathering oversight

mromaszewicz:
Response and RequestBodies need to include any additional types into model generation. This is an ancient bug, from the very first days of oapi-codegen.

Extend the issue-200 spec with inline objects containing additionalProperties inside components/responses and components/requestBodies. The test confirms the codegen produces named types (Bar_Pagination, Bar_Metadata) for these inline schemas.


jamietanna (approved):
We really need a custom linting rule to catch all the places we've missed the `AdditionalTypes...` needing to be added 🫣

PR #2221, opened 2026-02-14
fix: add omitempty to optional nullable fields

### LINKED ISSUE #2091: optional nullable field does not get omitempty json tag as mention in Readme
Readme states that by default nullable fields get `omitempty` tag. See https://github.com/oapi-codegen/oapi-codegen?tab=readme-ov-file#generating-nullable-types. But that doesn't seem to be the case.

```
#[0]$ cat config.yaml2
package: preferskipoptionalpointerwithomitzero
output: gen.go
generate:
  models: true
output-options:
  skip-prune: true
```

```
#[0]$ cat spec.yaml2
openapi: "3.0.0"
info:
  version: 1.0.0
  title: prefer-skip-optional-pointer-with-omitzero
components:
  schemas:
    ClientWithExtension:
      type: object
      properties:
        newfield:
          type: string
          nullable: true
```

```
#[0]$ oapi-codegen --config config.yaml2 spec.yaml2

#[0]$ cat gen.go
// Package preferskipoptionalpointerwithomitzero provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/oapi-codegen/oapi-codegen/v2 version v2.5.0 DO NOT EDIT.
package preferskipoptionalpointerwithomitzero

// ClientWithExtension defines model for ClientWithExtension.
type ClientWithExtension struct {
        Newfield *string `json:"newfield"`
}
```

```
#[0]$ oapi-codegen --version
github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen
v2.5.0

#[0]$ go version
go version go1.24.4 linux/amd64
```

mromaszewicz:
Fixes #2091

The `omitempty` JSON tag was not being added to optional nullable fields. The condition `!p.Nullable && shouldOmitEmpty` explicitly prevented any nullable field from receiving `omitempty`, even when the field was optional (not required). This contradicted the documented behavior.

The fix removes the `!p.Nullable` guard so nullable fields follow the same `omitempty` rules as non-nullable fields. The special-case exception for the `nullable-type` output option is no longer needed since the logic is now uniform.

Note: `x-go-type-skip-optional-pointer: true` on a nullable field suppresses the pointer (generating `string` instead of `*string`) but still correctly receives `omitempty` when the field is optional. Whether skip-optional-pointer should be allowed to suppress the pointer on nullable fields is a separate concern.

### DIFF
--- a/internal/test/issues/issue-1039/defaultbehaviour/types.gen.go
+++ b/internal/test/issues/issue-1039/defaultbehaviour/types.gen.go
@@ -6,7 +6,7 @@ package defaultbehaviour
 // PatchRequest A request to patch an existing user object.
 type PatchRequest struct {
 	// ComplexOptionalNullable Complex, optional and nullable
-	ComplexOptionalNullable *ComplexOptionalNullable `json:"complex_optional_nullable"`
+	ComplexOptionalNullable *ComplexOptionalNullable `json:"complex_optional_nullable,omitempty"`
 
 	// ComplexRequiredNullable Complex required and nullable
 	ComplexRequiredNullable *ComplexRequiredNullable `json:"complex_required_nullable"`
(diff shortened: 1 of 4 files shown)


mromaszewicz:
@greptile

PR #2223, opened 2026-02-14
feat: add configurable type mapping for OpenAPI primitive types

mromaszewicz:
Supersedes #428, since the original author's branch is gone and his PR only exists in the ref log.

Added a TypeMapping configuration that lets users override the default OpenAPI type/format to Go type mappings. User-specified mappings are merged on top of built-in defaults, so only overrides need to be specified. For example, mapping string/date-time to a civil.DateTime or changing the default integer type to int64.

Changes:
- New TypeMapping, FormatMapping, SimpleTypeSpec types with Merge() and Resolve() methods (pkg/codegen/typemapping.go)
- TypeMapping field added to Configuration struct
- Merged type mapping stored on globalState, initialized in Generate()
- Replaced hardcoded switch statements in oapiSchemaToGoType() with map-based lookups via globalState.typeMapping
- Updated configuration-schema.json with type-mapping property and reusable $defs for format-mapping and simple-type-spec
- Unit tests for merge, resolve, and default mapping completeness


configuration-schema.json
```diff
@@ -260,6 +260,25 @@
         }
       }
     },
+    "type-mapping": {
```
jamietanna: Should this maybe be under `output-options`?

mromaszewicz: Sure, I moved it. I put it under `import-mapping` since it was similar in concept

mromaszewicz:
Simple example added.

Once I'm done with behavioral additions, I intend to make a detailed document about all the options, linking to examples. Our configuration is under-documented.

PR #2226, opened 2026-02-16
fix(server-urls): use URL in GoDoc if `description` is empty

jamietanna:
Otherwise the sentence ends abruptly.


examples/generate/serverurls/gen.go
jamietanna, line 1: ⚠️ this is a regression

PR #2227, opened 2026-02-17
feat: add Valid() method to generated enum types

mromaszewicz:
Generate a Valid() bool method on each enum type that returns true when the receiver matches one of the defined enum constants and false otherwise. This lets callers validate enum values at runtime with a simple method call instead of hand-writing switch statements.

This is default-on because it only adds a new method to an already generated type -- existing code that does not call Valid() is completely unaffected, so this should be very unlikely to break anything.

### DIFF
--- a/internal/test/components/components.gen.go
+++ b/internal/test/components/components.gen.go
@@ -18,12 +18,40 @@ const (
 	Enum1Two   Enum1 = "Two"
 )
 
+// Valid indicates whether the value is a known member of the Enum1 enum.
+func (e Enum1) Valid() bool {
+	switch e {
+	case Enum1One:
+		return true
+	case Enum1Three:
+		return true
+	case Enum1Two:
+		return true
+	default:
+		return false
+	}
+}
+
 // Defines values for Enum2.
 const (
 	Enum2Four  Enum2 = "Four"
 	Enum2Three Enum2 = "Three"
 	Enum2Two   Enum2 = "Two"
 )
+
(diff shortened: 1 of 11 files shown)


jamietanna (approved):
Nice!

jamietanna (commented):
From https://github.com/oapi-codegen/oapi-codegen/pull/2181 I do wonder if it'd be useful to be able to get the list of valid enum names, though, i.e. as a slice

But maybe we can see if this solves the underlying need folks have

mromaszewicz:
> From #2181 I do wonder if it'd be useful to be able to get the list of valid enum names, though, i.e. as a slice
> 
> But maybe we can see if this solves the underlying need folks have

I think it does, because you can always validate in your own code with something like `MyEnum(x).Valid()`. We can certainly return all the enum values in the future if needed, or even make a `Stringer` out of it.

rkosegi:
> This is default-on because it only adds a new method to an already generated type -- existing code that does not call Valid() is completely unaffected, so this should be very unlikely to break anything.

If this is default-on, how can we turn it off explicitly? 
I have a case where I need to add method to generated type which has pointer receiver and generated `Valid()` method has value receiver. It compiles but it's discouraged and I actually don't need Valid() in this case.

mromaszewicz:
Ha! You are right, @rkosegi , I forgot to add the flag. I will do so in a subsequent commit. Whenever I add something I think is useful, it inevitably clashes with someone's code.

PR #2231, opened 2026-02-19
fix(deps): use updated import path for `github.com/speakeasy-api/openapi/overlay`

netixx:
Fixes: https://github.com/oapi-codegen/oapi-codegen/issues/2230


go.mod
```diff
@@ -1,20 +1,20 @@
 module github.com/oapi-codegen/oapi-codegen/v2
 
-go 1.22.5
+go 1.24.3
```
jamietanna, line 4: 🛑 we don't want to bump this high

jamietanna:
OK, so this looks like this may also involve bumping https://github.com/oapi-codegen/oapi-codegen/pull/2139 (and our `go` directive), which we try hard to avoid - we'll have to think about this

go.mod
alexandear: But why? Even Go 1.24 is not supported anymore since Go 1.26 was released.

jamietanna: https://github.com/oapi-codegen/oapi-codegen/blob/main/SUPPORT.md#minimum-required-go-toolchain-version

netixx:
in the meantime, to you have a solution to fix the broken dependency chain ?

mromaszewicz:
I think we need to talk about the required Go version. I originally tried to keep the oldest version possible, but the most important part to me is that the generated code isn't too version specific.

I think that jumping to 1.24 would allow us to move to the new usage of `go tool` have only one way of using and installing oapi-codegen. Being 1.22 based forces us to use very old versions of some dependencies.

jamietanna:
We'll bump to 1.24 as part of the 2.7.0 release

go.mod
mromaszewicz: I've bumped everything to 1.24 for now, so this conversation is moot.

mromaszewicz:
Thank you, @netixx

PR #2233, opened 2026-02-23
feat: allow preserving specific schemas during prune (preserve-schemas + x-oapi-codegen-keep-unused)

aureleoules:
Add two ways to keep unreferenced component schemas from being pruned:

- output-options.preserve-schemas: list of schema names to always keep, even when not referenced anywhere in the spec.
- x-oapi-codegen-keep-unused: set to true on a schema in the OpenAPI spec so that schema is never pruned when unreferenced.

pruneUnusedComponents() now takes an optional preserveSchemas []string (from config); schemas in that list or with the extension are not removed. Existing call sites use nil. Tests and configuration-schema/README updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2017,7 +2017,10 @@ output: only-models.gen.go
 generate:
   models: true
 output-options:
-  # NOTE that this is only required for the `Unreferenced` type
+  # To keep only specific unreferenced schemas, use preserve-schemas instead of skip-prune:
+  # preserve-schemas: [Unreferenced]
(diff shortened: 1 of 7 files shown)


mromaszewicz:
@aureleoules , we already have this behavior:
https://github.com/oapi-codegen/oapi-codegen/blob/main/pkg/codegen/configuration.go#L248

Is this not sufficient for what you are trying to accomplish?

Go does dead code pruning when it compiles, so having some unused schemas in the generated go code when setting `skip-prune: true` is ok. i don't think we need to pick, per-schema, which one to include.

PR #2241, opened 2026-02-24
fix: qualify external ref schema types in default response codes

### LINKED ISSUE #2113: $ref  in responses is not adding the imported reference properly
Using this model definition as an example:

```yaml
openapi: 3.0.4

components:
  schemas:
    ProblemDetails:
      type: object
      description: RFC 7807 Problem Details
      properties:
        type:
          type: string
          format: uri
          description: A URI reference that identifies the problem type
        title:
          type: string
          description: A short, human-readable summary of the problem type
        status:
          type: integer
          description: The HTTP status code
      required: [title, status]
  responses:
    StandardError:
      description: Error response using RFC 7807 Problem Details
      content:
        application/problem+json:
          schema:
            $ref: "#/components/schemas/ProblemDetails"
```

Next definition is working fine:
```yaml
        default:
          description: Error (RFC 7807)
          content:
            application/problem+json:
              schema:
                $ref: "./common/problemdetails.yaml#/components/schemas/ProblemDetails"
```

generating this code:

```go
import externalRef3 "github.com/XXXXXX/YYYYYY/generated/go/common/problemdetails"

type UsersListdefaultApplicationProblemPlusJSONResponse struct {
	Body       externalRef3.ProblemDetails
	StatusCode int
}
```

But this definition is not generating the right code :
```yaml
        default:
          $ref: "./common/problemdetails.yaml#/components/responses/StandardError"
```

generating this code, where is not using the reference to access to the model:

```go
import externalRef3 "github.com/XXXXXX/YYYYYY/generated/go/common/problemdetails"

type UsersListdefaultApplicationProblemPlusJSONResponse struct {
	Body       ProblemDetails // FAILING HERE. Should be externalRef3.ProblemDetails
	StatusCode int
}
```

Next is the config file:
```yaml
package: users
output: ../../../../generated/go/users/handlers.chi.gen.go
generate:
  models: true
  embedded-spec: true
  strict-server: true
  chi-server: true
output-options:
  skip-prune: true
import-mapping:
  ./common/problemdetails.yaml: "github.com/XXXXXX/YYYYYY/generated/go/common/problemdetails"
```

mromaszewicz:
Closes #2113

When a response uses a $ref to an external components/responses definition, the generated strict-server code for non-fixed status codes (e.g. default) used the inner schema type without the external package qualifier, causing a compile error. Apply
ensureExternalRefsInSchema to response content schemas after resolving external response refs.


mromaszewicz:
@greptileai

PR #2256, opened 2026-02-26
feat: Pass schema type/formats to runtime v1.2.0 to allow better parameter serialization

mromaszewicz:


mromaszewicz:
@greptileai

PR #2260, opened 2026-02-27
chore: refactor all tests by category

mromaszewicz:
Replace the organically grown layout — 56 issues/issue-NNNN directories
plus a few kitchen-sink specs — with 13 feature categories (schemas,
aggregates, parameters, bodies, servers, clients, events, references,
extensions, options, naming, openapi31, spec_validation).

Each scenario is now a focused spec/config/test unit under its category,
named for what it tests rather than an issue number, with issue context
kept as "From issue-NNNN" comments in the absorbing spec. Cross-framework
suites (parameters/roundtrip, servers/strict, events/webhooks,
servers/routers/smoke) share one spec and one table-driven harness per
concern; only the per-framework server/types generation remains
duplicated, since generated servers cannot share packages.

Coverage was audited directory-by-directory against the old tree: nothing
was dropped, and three previously lost areas were restored (hand-crafted
wire-format parameter binding, operationId name-normalization,
nil-vs-empty query array serialization). Greptile review rules now
enforce the category placement so issue directories don't return.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


mromaszewicz:
@greptileai

mromaszewicz:
This code review is quite immense. It might be easier to look at my source branch to see what the layout looks like.

I think it will be a lot easier to add test cases for specific things once this is merged.

jamietanna:
@mromaszewicz would you like this change in before any other PRs? What's your priority order for these to avoid conflcits/rebasing fun?

mromaszewicz:
I will use AI to rebase this, it shines there. I'd like to get this in just as a quality of life improvement. It exposed some issues we have with matrix and label parameters as-is (so tests are disabled for now).

jamietanna:
A couple of things:

- This looks good, and a lot neater!
- Something I've considered for a while is it'd be nice to have a utility script to run that'll bootstrap one of these `issuesXXXX` directories with everything you need - maybe not as needed now, but may still be worthwhile?
- I'd quite like to make sure we keep a reference to issue numbers where there are tests that have come in via one - it helps us track down "real-world" behaviour and whether there's any additional context that can be gleamed. Can be a comment in the package's documentation string or i.e. inline in the pieces of spec that are for a given issue
- In the squashing down of files, are we confident in the coverage of everything we used to have? (I.e. we're no longer testing server generation)

Otherwise LGTM, let's get this quality-of-life improvement in!

mromaszewicz:
A lot of work has happened since this commit, I will rebase it and re-do it once 2.8.0 is out. It'll be a lot of work.

mromaszewicz:
I'm revisiting this. The code has drifted hugely, so I'm basically re-doing it, and doing it a little more cleanly this time.

We won't be needing the per-issue test dirs. I really hope to avoid them, and the new top level categories should cover more bases. It's easier to add some edge case to an existing spec than to make a whole new test dir. Most of that code is wasted boilerplate.

mromaszewicz:
@greptileai - review.

PR #2263, opened 2026-02-28
 fix: use RequiresNilCheck for all params

### LINKED ISSUE #2238: Ensure all `prefer-skip-optional-pointer` cases are checked
As we've seen with https://github.com/oapi-codegen/oapi-codegen/pull/2237 and https://github.com/oapi-codegen/oapi-codegen/commit/c4ba545dc7f4c67d796d7fbde1203b775f3a3056 we should make sure we cover this for all cases

mromaszewicz:
Fixes #2238

PR #2237 fixed query parameters to use RequiresNilCheck instead of HasOptionalPointer for nil-check guards, but the same bug remained in the header and cookie parameter sections of client.tmpl. This caused optional array params (e.g. []string) to be sent even when nil, because prefer-skip-optional-pointer removes the pointer wrapper and HasOptionalPointer returns false, skipping the nil check entirely.


jamietanna (approved):
Thanks for following up on this

PR #2264, opened 2026-02-28
Update all dependencies to Go 1.24

mromaszewicz:
This change makes our minimum required Go version 1.24 (the last supported) and updates module structure to reflect that, because we had lots of nested modules which were there to test code that was 1.24 only.


.github/workflows/ci.yml
```diff
@@ -9,6 +9,7 @@ jobs:
   build:
     uses: oapi-codegen/actions/.github/workflows/ci.yml@75566d848d25021f137594c947f26171094fb511 # v0.5.0
     with:
+      excluding_versions: '["1.22", "1.23"]'
```
jamietanna, line 12: We could do this upstream in the actions repo (after this change) and then that'll mean we can roll out minimum Go 1.24 across all our repos

README.md
```diff
@@ -59,29 +57,6 @@ From there, each invocation of `oapi-codegen` would be used like so:
 //go:generate go tool oapi-codegen -config cfg.yaml ../../api.yaml
 ```
 
-### Prior to Go 1.24
```
jamietanna, line 62: IMO this pattern may still be used by folks - especially with a separate `tools/go.mod`, it can be better for the overall dependency tree - but not blocking this review

.github/workflows/ci.yml
mromaszewicz: Yeah, we have to update all the others first. I'll do that next.

PR #2265, opened 2026-03-01
test: only use placeholder version in tests

jamietanna:


mromaszewicz:
I refactored everything to use Go 1.24 and just got rid of that go.mod altogether.

jamietanna:
https://github.com/oapi-codegen/oapi-codegen/pull/2264 makes sense to get in, instead - this was something I spotted while doing something unrelated!

PR #2274, opened 2026-03-04
fix: runtime import breaks

jinuthankachan:
* fix the break in import od runtime  in the generated code after https://github.com/oapi-codegen/runtime/pull/89/changes/702ab913d08cc6a9cfe3fb00dd8722e1cca311ae
* closes https://github.com/oapi-codegen/oapi-codegen/issues/2275


jinuthankachan:
Should be solved by https://github.com/oapi-codegen/oapi-codegen/pull/2271

PR #2277, opened 2026-03-04
refactor petstore example

mromaszewicz:
Clean up the petstore-expanded example so each server variant is
self-contained: its own generated types and embedded spec, its own copy
of the trivial in-memory store (a Go map), and its own handler
implementation. The backend is simple enough that copying it into each
variant keeps every example readable on its own, with no cross-variant
imports to follow.

Add a single shared test client (common/client) that exercises any
variant's full CRUD behavior over real HTTP, replacing the old
per-variant unit tests that drove the handlers directly without going
through the network.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


jamietanna (commented):
Although I see the benefit I do wonder if we should embrace the duplication - I think the original intent of this was "look how you'd be able to implement the simple 'petstore' API, with each of these routers"

If we set up shared code, that defeats the point

Not necessarily saying 'no' here, but wondering if we should think if this is the right approach - or whether there's a simpler/different API surface that we want to introduce an example for, where we can implement it once-per-Router?

mromaszewicz:
I'm of two minds on this one. I did this because this started out as one single example with Echo, my initial HTTP router. Now, all these implementations are copy/pasted from each other, and half of them didn't actually work properly (a generated client could not communicate with a generated server).

So, yes, a single example is more convoluted now, however, it's much more obvious which parts are generic and which parts are router-specific.

As-is in this PR, the client can communicate with all the back ends and adding another backend is much easier.

I can drop this refactor and fix up the broken servers independently, but I was thinking of how to make this more thorough and better tested and I figured a shared "store" implementation and a test via the generatd client code were pretty good.

mromaszewicz:
Ok, I'm reducing the scope of this. The implementations will all be simlilarly copy/pasted with each other, and there's going to be a test client which is runnable against any of the server backends to validate its behavior. So this is more of a new test, and bringing everything in sync with each other.

mromaszewicz:
@greptileai review

PR #2287, opened 2026-03-09
fix(cli): remove unused `initialism-overrides` option

### LINKED ISSUE #2276: bug: unused option `initialism-overrides`
At the moment, the `initialism-overrides` option is not used anywhere in the code. I suggest marking it as deprecated or removing it.

The logic has changed [in this PR](https://github.com/oapi-codegen/oapi-codegen/pull/1041/changes#diff-adba78e29f6e65748fe711bfeb66b4a82a9b59226b5d2c0d8c891e2c5a7d48f1R675).

gaiaz-iusipov:
Closes #2276


mromaszewicz:
I think this option was hardly ever used. Generally, removing an option breaks those users who set it explicitly, and it's something we try to avoid, but I think that this can be documented inside the release notes, since it a small config file change.

mromaszewicz (approved):

PR #2295, opened 2026-03-16
fix(deps): update module github.com/getkin/kin-openapi to v0.134.0 (go.mod)

ilia-rassadin-rn:
In kin-openapi v0.134.0, Discriminator.Mapping changed from map[string]string to StringMap[MappingRef], where MappingRef is type MappingRef SchemaRef (a struct with a Ref string field). See https://github.com/getkin/kin-openapi/pull/1108

This caused a compile error in generateUnion where v (MappingRef) was compared directly against element.Ref (string). Fix by comparing v.Ref instead.

Bump kin-openapi dependency to v0.134.0.


corani:
This would fix https://github.com/oapi-codegen/oapi-codegen/pull/2292, which is currently blocking us from upgrading github.com/getkin/kin-openapi to v0.134.0 in our projects. A patch release for `oapi-codegen` containing this change would be highly appreciated!

jamietanna:
Please can you make sure you've updated all the references? 

Replaces https://github.com/oapi-codegen/oapi-codegen/pull/2292

jamietanna:
We likely won't do a patch release due to changes already on `main`, but it'll be possible for folks to pin to `main`'s commits until the next release

jamietanna (approved):

ilia-rassadin-rn:
Thanks!

sonasingh46:
Thank you for the fix. Just hit this issue and reached here.

corani:
> We likely won't do a patch release due to changes already on `main`, but it'll be possible for folks to pin to `main`'s commits until the next release

Do you have any idea when the next release would be?

jamietanna:
No plans - companies sponsoring us to be able to put more time towards a release schedule would be appreciated!

https://github.com/oapi-codegen/oapi-codegen?tab=readme-ov-file#pinning-to-commits

PR #2299, opened 2026-03-23
Propagate external refs into allOf

### LINKED ISSUE #2288: allOf with cross-file  does not qualify nested type references from source schema
### Description

When using `allOf` to extend a schema from another file, oapi-codegen flattens the fields correctly but does not qualify nested type references that originate from the source schema's file. This produces uncompilable Go code.

### Reproduction

Given two OpenAPI spec files with an `import-mapping` configured:

**`api/common/openapi.yaml`**
```yaml
openapi: 3.0.0
info:
  title: Common
  version: 0.1.0
paths: {}
components:
  schemas:
    Role:
      type: string
      enum: [admin, user]
    User:
      type: object
      required: [id, username]
      properties:
        id:
          type: string
        username:
          type: string
        roles:
          type: array
          items:
            $ref: '#/components/schemas/Role'
```

**`api/service/openapi.yaml`**
```yaml
openapi: 3.0.0
info:
  title: Service
  version: 0.1.0
paths: {}
components:
  schemas:
    EnrichedUser:
      allOf:
      - $ref: '../common/openapi.yaml#/components/schemas/User'
      - type: object
        properties:
          extra_field:
            type: string
```

**`api/service/types.yaml`** (codegen config)
```yaml
package: service
generate:
  models: true
output: types.gen.go
import-mapping:
  '../common/openapi.yaml': common github.com/example/api/common
```

### Expected behavior

Generated `EnrichedUser` should qualify the `Role` reference via the import mapping:

```go
type EnrichedUser struct {
    ExtraField *string        `json:"extra_field,omitempty"`
    ID         string         `json:"id"`
    Roles      *[]common.Role `json:"roles,omitempty"`
    Username   string         `json:"username"`
}
```

### Actual behavior

Generated code uses an unqualified `Role` reference:

```go
type EnrichedUser struct {
    ExtraField *string    `json:"extra_field,omitempty"`
    ID         string     `json:"id"`
    Roles      *[]Role    `json:"roles,omitempty"`  // <-- unqualified
    Username   string     `json:"username"`
}
```

This fails to compile: `undefined: Role`

### Workaround

Override the field in the `allOf` extension with a direct cross-file `$ref`:

```yaml
    EnrichedUser:
      allOf:
      - $ref: '../common/openapi.yaml#/components/schemas/User'
      - type: object
        properties:
          roles:
            type: array
            items:
              $ref: '../common/openapi.yaml#/components/schemas/Role'
          extra_field:
            type: string
```

This correctly generates `Roles *[]common.Role`.

### Version

oapi-codegen v2.6.0

mromaszewicz:
Closes: #2288

When allOf references an external schema, valueWithPropagatedRef was only rewriting direct property $refs to include the remote file path. Nested refs — array items, additionalProperties, and sub-object properties — were left as local "#/..." refs, producing unqualified Go type names that fail to compile.

Extract the propagation logic into a recursive propagateRemoteRefs function that walks Properties, Items, and AdditionalProperties, rewriting any local ref it finds. This ensures types like `*[]externalRef0.Role` are correctly qualified instead of the broken `*[]Role`.

Extend the externalref test with a User/Role/EnrichedUser scenario that exercises the array-items case from the issue report.

### DIFF
--- a/internal/test/externalref/packageA/externalref.gen.go
+++ b/internal/test/externalref/packageA/externalref.gen.go
@@ -16,18 +16,35 @@ import (
 	externalRef0 "github.com/oapi-codegen/oapi-codegen/v2/internal/test/externalref/packageB"
 )
 
+// EnrichedUser defines model for EnrichedUser.
+type EnrichedUser struct {
+	ExtraField *string              `json:"extra_field,omitempty"`
+	Id         string               `json:"id"`
+	Roles      *[]externalRef0.Role `json:"roles,omitempty"`
+	Username   string               `json:"username"`
+}
+
 // ObjectA defines model for ObjectA.
 type ObjectA struct {
 	Name    *string               `json:"name,omitempty"`
 	ObjectB *externalRef0.ObjectB `json:"object_b,omitempty"`
 }
 
+// StatusPostPayload defines model for StatusPostPayload.
+type StatusPostPayload struct {
+	Reason *string                 `json:"reason,omitempty"`
+	Status externalRef0.StatusEnum `json:"status"`
+}
+
 // Base64 encoded, gzipped, json marshaled Swagger object
 var swaggerSpec = []string{
 
-	"H4sIAAAAAAAC/4yOMcrDMAxG7/L9/xjI7q25QI8QHKMkbhNZ2OpQgu5e7BS6dOikBw896UBIuyQm1gJ3",
-	"oISVdt/wOt0o6KWi5CSUNVIT7HeqU59CcCiaIy+wDqltjFOV/5lmOPz1n37/jvfiw90vNIxFKIznnQFm",
-	"1uG7+vUFa43Ic4Ljx7Z1SELsJcIBNa5rOY29AgAA//84dUj5+QAAAA==",
+	"H4sIAAAAAAAC/5xTzW7CMAx+F2/HSIhrbkPiDNo0LghVJnEhW5pkSToNVXn3KaEIKjqp7GbX9tfvp+1A",
+	"2MZZQyYG4B0EcaQGS7k0XokjyfdAPveo9aoGvu3g2VMNHJ5m19tZfzhzKD7xQIsqOBJVuU2sA+etIx8V",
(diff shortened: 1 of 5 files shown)


mromaszewicz:
@greptileai

cdevienne:
I tested your patch with the following files, which is an extract of a bigger api in which I encountered the same issue.

cfg-api.yaml:
```yaml
package: api
output: api.gen.go
generate:
  models: true
output-options:
  # to make sure that all types are generated
  skip-prune: true
import-mapping:
  ./types.yaml: "types"
```

api.yaml:
```yaml
components:
  schemas:
    CustomerStatusPostPayloadV1:
      allOf:
        - $ref: './types.yaml#/components/schemas/CustomerStatusV1'
        - type: object
          properties:
            name:
              type: string
```

types.yaml
```yaml
components:
  schemas:
    CustomerStatusEnum:
      type: string
      enum:
        - Approved
        - Disputed
        - Suspended

    CustomerStatusV1:
      allOf:
        - type: object
          properties:
            status:
              $ref: '#/components/schemas/CustomerStatusEnum'
```

The generated files still doesn't reference the enum properly:


```go
// Package api provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/oapi-codegen/oapi-codegen/v2 version v2.6.0 DO NOT EDIT.
package api

// CustomerStatusPostPayloadV1 defines model for CustomerStatusPostPayloadV1.
type CustomerStatusPostPayloadV1 struct {
	Name   *string             `json:"name,omitempty"`
	Status *CustomerStatusEnum `json:"status,omitempty"`
}
```

mromaszewicz:
@cdevienne , thanks for this example. I'm very minimalist with this code because this aggregate schema traversal is so fragile. Could you see if this latest PR works for you? I've reproduced your test case in our unit tests, and it failed before my changes, and passes after.

cdevienne:
@mromaszewicz Tested it, it works!

I have another similar issue though, will report it separately.

Thanks a lot

cdevienne:
Just in case, the similar issue I got is: #2308

PR #2300, opened 2026-03-23
chore(examples): update jwx to v3

lestrrat:
github.com/lestrrat-go/jwx (v0 or v1) is no longer supported. If you can, please use github.com/lestrrat-go/jwx/v3


mromaszewicz:
Awesome, thank you! It's an honor to get a PR from you. I've been using your JWT code since forever.

PR #2301, opened 2026-03-23
support optional/nullable response headers

### LINKED ISSUE #2267: Support nullable response headers
The `strict-interface` template currently treats all response headers as required. However, it should be possible to omit optional headers.

Possible template changes could be:

```diff
            type {{$opid}}{{$statusCode}}ResponseHeaders struct {
                {{range .Headers -}}
+                   {{if .Schema.OAPISchema.Nullable -}}
+                       {{.GoName}} nullable.Nullable[{{.Schema.TypeDecl}}]
+                   {{else -}}
                        {{.GoName}} {{.Schema.TypeDecl}}
+                   {{end -}}
                {{end -}}
            }
```

and:

```diff
                {{range $headers -}}
+                   {{if .Schema.OAPISchema.Nullable -}}
+                       if val, err := response.Headers.{{.GoName}}.Get(); err == nil {
+                           w.Header().Set("{{.Name}}", fmt.Sprint(val))
+                       }
+                   {{else -}}
                        w.Header().Set("{{.Name}}", fmt.Sprint(response.Headers.{{.GoName}}))
+                   {{end -}}
                {{end -}}
```

Maybe it might also make sense respecting the pointers vs. nullable setting and using pointers accordingly.

mromaszewicz:
Closes: #2267

Response headers in strict server code were always generated as direct value types regardless of whether they were required. This adds support for optional and nullable response headers in the strict-interface and strict-responses templates.

Add Required and Nullable fields to ResponseHeaderDefinition, populated from the OpenAPI header object. Add GoTypeDef(), IsOptional(), and IsNullable() methods that mirror the existing Property logic for determining whether a field should be a pointer, nullable.Nullable[T], or a direct value.

Update strict-interface.tmpl and strict-responses.tmpl to use GoTypeDef for struct field types, and to conditionally guard w.Header().Set() calls with nil/IsSpecified checks for optional/nullable headers.

Note: this is a breaking change for specs where response headers lack explicit `required: true`, since the OpenAPI default is false. Existing headers will change from direct values to pointers. This may need to be gated behind a config option.

### DIFF
--- a/configuration-schema.json
+++ b/configuration-schema.json
@@ -115,6 +115,10 @@
         "preserve-original-operation-id-casing-in-embedded-spec": {
           "type": "boolean",
           "description": "When `oapi-codegen` parses the original OpenAPI specification, it will apply the configured `output-options.name-normalizer` to each operation's `operationId` before that is used to generate code from.\nHowever, this is also applied to the copy of the `operationId`s in the `embedded-spec` generation, which means that the embedded OpenAPI specification is then out-of-sync with the input specificiation.\nTo ensure that the `operationId` in the embedded spec is preserved as-is from the input specification, set this. NOTE that this will not impact generated code.\nNOTE that if you're using `include-operation-ids` or `exclude-operation-ids` you may want to ensure that the `operationId`s used are correct."
+        },
+        "headers-implicitly-required": {
+          "type": "boolean",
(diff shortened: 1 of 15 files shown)


jamietanna (commented):
Per offline discussion, this makes sense to do - we should correctly add this behaviour to make sure that we're reflecting what is the "correct" default

As we've said, I think adding a `CompatibilityOption` for this, so folks can revert back to the old behaviour would be best

But I'm of the opinion that moving folks over by default to the correct code path (which is technically breaking their generated code) is best

Do we know if we could use `go:fix` or anything to help them migrate more easily?

mromaszewicz:
I don't know how to use go:fix to fix this. I think we'll document the change and mention it in the release.

jamietanna (approved):

PR #2305, opened 2026-03-26
Replace old hand written slice sorting with slices package, and other built-in helpers

gaiaz-iusipov:
In this PR I made small stylistic changes: using functions from standard packages (`maps.Copy`, `slices.Contains`, etc.)


mromaszewicz:
You have two unrelated changes here - using modern packages, and the error compile time check.

I would very happily merge the first part, as-is, however, since we are a code generator, we can ensure that the generated code always implements error, without having a compile time check for our own output. I don't think we need that second part.

You've also only added that to the stdhttp server, and we try to keep as much parity between generated code as possible.

gaiaz-iusipov:
Thanks. This kind of compile-time check is used in various Go projects as a guarantee that certain types implement an interface and that this behavior won’t accidentally break in the future. But I agree that it’s better if I remove those changes and keep only the set that introduces the use of the standard package.

mromaszewicz:
@greptileai

mromaszewicz (approved):

mromaszewicz:
Thank you!

PR #2314, opened 2026-04-03
Constrain Renovate to Go 1.24

mromaszewicz:
Renovate keeps suggesting dependency updates which bump our pinned Go version. This should constrain it to the 1.24 toolchain. This constraint needs to be set to match the Go version in our top level go.mod


renovate.json
```diff
@@ -4,7 +4,10 @@
 		"local>oapi-codegen/renovate-config"
 	],
 	"gomod": {
-		"ignorePaths": []
+		"ignorePaths": [],
+		"constraints": {
```
jamietanna: I'd move this up to a top-level key, rather than under `gomod` (as it applies to every update, and it may not necessarily merge in correctly - not tested it)

jamietanna (changes requested):
We'll probably also want to make sure that `GOTOOLCHAIN=local` via [`env`](https://docs.renovatebot.com/configuration-options/#env), otherwise we'll automagically pull the new toolchain on updates - we'd probably prefer a PR to fail its creation if it's trying to use a newer version?

renovate.json
mromaszewicz: done.

mromaszewicz:
> We'll probably also want to make sure that `GOTOOLCHAIN=local` via [`env`](https://docs.renovatebot.com/configuration-options/#env), otherwise we'll automagically pull the new toolchain on updates - we'd probably prefer a PR to fail its creation if it's trying to use a newer version?

We already have the toolchain pinned in go.mod, and I think this might cause problems, because the local toolchain, just hypothetically, might be 1.24.3 and this will cause a failure.

jamietanna:
Discussing in https://github.com/renovatebot/renovate/discussions/42459

jamietanna:
> because the local toolchain, just hypothetically, might be 1.24.3 and this will cause a failure.

The failure (IMO) would be preferable, as it'd make sure that it wasn't updating the `go` directive

(That being said, I think Renovate still raises a PR if branch creation fails)

jamietanna:
FYI I've got https://github.com/renovatebot/renovate/pull/42569 which will allow us to handle this a little better - working with the other maintainers to get that through hopefully this week

This will stop Renovate suggesting an update that requires an updated `go` directive - mind taking a look at the docs in the PR to see if it reads OK to you?

mromaszewicz:
I don't really care how it's done, I just want Renovate to stop recommending things we don't want :) Your proposed changes to Renovate seem fine, but i'm not super familiar with its internals, and too busy to dig in.

jamietanna:
Replaced by https://github.com/oapi-codegen/renovate-config/pull/3

Issue #1373, opened 2023-12-05
Stack Overflow error with recursive types


etabegro:
The following recursive definition of an object generates a stack overflow error. I am not sure, whether this valid according to the OpenAPI specification. However, it would at least be helpful if the error message generated by oapi-codegen was more readable than it currently is. Maybe something like "Invalid recursive object definition" would be good.

This issue is similar to #52 but not exactly the same. In that case the recursive definition was within a Reference Object. In the case of this new issue, the object directly refers to itself.

```
openapi: "3.0.0"
info:
  version: 0.1.0
  title: example
components:
  schemas:
    NewObject:
      properties:
        name:
          type: string
          description: Name of the object
      required:
        - name
     
    SecondObject:
      allOf:
        - $ref: "#/components/schemas/NewObject"
        - $ref: "#/components/schemas/SecondObject"
```

```
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc0205476b8 stack=[0xc020546000, 0xc040546000]
fatal error: stack overflow

runtime stack:
runtime.throw({0xf8cfb6?, 0x8f89bffc70?})
        C:/Program Files/Go/src/runtime/panic.go:1077 +0x65 fp=0x8f89bffc30 sp=0x8f89bffc00 pc=0x9bafe5
runtime.newstack()
        C:/Program Files/Go/src/runtime/stack.go:1107 +0x5bb fp=0x8f89bffde0 sp=0x8f89bffc30 pc=0x9d1b5b
runtime.morestack()
        C:/Program Files/Go/src/runtime/asm_amd64.s:593 +0x87 fp=0x8f89bffde8 sp=0x8f89bffde0 pc=0x9e8f07

goroutine 1 [running]:
github.com/deepmap/oapi-codegen/v2/pkg/codegen.mergeAllOf({_, _, _})
        C:/users/go/pkg/mod/github.com/deepmap/oapi-codegen/v2@v2.0.0/pkg/codegen/merge_schemas.go:73 +0x2de fp=0xc0205476c8 sp=0xc0205476c0 pc=0xe6cdbe
github.com/deepmap/oapi-codegen/v2/pkg/codegen.mergeOpenapiSchemas({0xc003a65a10, {0x0, 0x0, 0x0}, {0x0, 0x0, 0x0}, {0x0, 0x0, 0x0}, ...}, ...)
[...]

PR #2316, opened 2026-04-04
fix: allOf cycle detection oversight

mromaszewicz:
PR #1377 was merged too early. This is a small follow-up tweak to make sure that #1373 is fixed properly.

seed allOf[0] ref into cycle detection to prevent stack overflow when self-ref is first

When the self-referencing $ref appears at position 0 in allOf, its ref was never added to seenSchemaRef. If s1's own AllOf contained a back-reference to itself, mergeAllOf would not detect the cycle.

Seed allOf[0].Ref into a top-level seen set that is copied into each iteration's seenSchemaRef map, closing this gap.

### DIFF
--- a/pkg/codegen/merge_schemas.go
+++ b/pkg/codegen/merge_schemas.go
@@ -32,6 +32,13 @@ func mergeSchemas(allOf []*openapi3.SchemaRef, path []string) (Schema, error) {
 		return Schema{}, err
 	}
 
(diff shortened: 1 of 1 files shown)


jamietanna (approved):
Seems reasonable, but if we're able to get a test (or a spec) to validate the change works as expectedly that'd be ideal!

PR #2334, opened 2026-04-23
Flagify enum validator generation

mromaszewicz:
Add an `output-options.skip-enum-validate` flag that suppresses the `Valid()` method generated on enum types. The method is still emitted by default; users whose code defines its own `Valid()` on the same type can now opt out instead of seeing a compile-time conflict.

Relates to PR #2227, which introduced the `Valid()` method.


mromaszewicz:
@greptileai

jamietanna (approved):

PR #2353, opened 2026-04-30
per-operation middleware in Echo

### LINKED ISSUE #518: Using per-endpoint middleware
At the moment when `RegisterHandlersWithBaseURL` is generated there is no way to specify per-handler middleware. In our use case we need to use some middleware's on per-handler bases rather than global ones.

The `EchoRouter` interface has a 3rd param to pass on the middleware:
```
type EchoRouter interface {
	CONNECT(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
	DELETE(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
	GET(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
	HEAD(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
	OPTIONS(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
	PATCH(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
	POST(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
	PUT(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
	TRACE(path string, h echo.HandlerFunc, m ...echo.MiddlewareFunc) *echo.Route
}
```

However there's no way to let `RegisterHandlersWithBaseURL` know about it. Is there any way to do it without having a global middleware which has to check whether it can be run for the given request?

mromaszewicz:
Closes: #518

Add a RegisterHandlersWithOptions to both Echo backends, which takes a BaseUrl and a map of OperationMiddlewares, keyed by spec OperationId. Only Echo supports registering per-op middleware, so this change is only for Echo.


mromaszewicz:
> ### Greptile Summary
> 
> This PR implements per-operation middleware support for both Echo v4 and Echo v5 backends by introducing `RegisterHandlersWithOptions` (with a `RegisterHandlersOptions` struct containing `BaseURL` and `OperationMiddlewares`) and refactoring the existing `RegisterHandlers`/`RegisterHandlersWithBaseURL` helpers to delegate to it. The change is backward-compatible, correctly uses the raw spec `operationId` as the map key (with a normalized-identifier fallback when the spec omits one), and includes a solid regression test in `internal/test/issues/issue-518/`.
> 
>     * `examples/petstore-expanded/echo/api/petstore-server.gen.go` contains unrelated parameter-binding changes (`BindQueryParameter` → `BindQueryParameterWithOptions` with added `Type`/`Format` fields; same for `BindStyledParameterWithOptions`) that do not appear in the echo-v5 equivalent — this file appears to have been regenerated with a toolchain that includes additional unreleased changes, and will likely cause a CI diff failure.
> 
> 
> ### Confidence Score: 3/5
> 
> Mostly safe to merge, but the unrelated binding-change drift in the echo/v4 petstore example file needs to be resolved first.
> 
> The core template change and regression test are well-implemented, but a P1 finding — unrelated parameter-binding changes in examples/petstore-expanded/echo/api/petstore-server.gen.go absent from the echo-v5 counterpart — indicates the file was regenerated with a different toolchain and will likely cause a CI make generate diff failure.
> 
> examples/petstore-expanded/echo/api/petstore-server.gen.go — contains binding-call changes unrelated to this PR that are not present in the echo-v5 equivalent.
> ### Important Files Changed
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> ### Comments Outside Diff (1)
> 
>     1. `examples/petstore-expanded/echo/api/petstore-server.gen.go`, line 63-73 ([link](https://github.com/oapi-codegen/oapi-codegen/blob/f4a74a25e165303380af81ef55b6f54d17b8c948/examples/petstore-expanded/echo/api/petstore-server.gen.go#L63-L73))
>        [P1](#) **Unrelated drift in regenerated file**
>        `examples/petstore-expanded/echo/api/petstore-server.gen.go` contains parameter-binding changes unrelated to this PR's stated goal of per-operation middleware. Specifically, calls to `BindQueryParameter` have been replaced with `BindQueryParameterWithOptions` (gaining `Type`/`Format` fields), and `BindStyledParameterWithOptions` call sites gain the same extra fields. These changes do not appear in the otherwise-analogous echo-v5 petstore file, suggesting this file was regenerated with a local toolchain that includes additional unreleased changes. Per the project policy, CI fails when `make generate` produces a diff — if the intended generator does not yet emit these binding changes, CI will fail on this file.
> 
> 
> Reviews (1): Last reviewed commit: ["per-operation middleware in Echo"](https://github.com/oapi-codegen/oapi-codegen/commit/f4a74a25e165303380af81ef55b6f54d17b8c948) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30330787)

@greptileai - this is because for some reason, the petstore boilerplate hasn't successfully regenerated into a while. Our "make generate" target bypasses it. I will fix that in a separate commit. This is correct.

PR #2354, opened 2026-04-30
fix example codegen

mromaszewicz:
aae687ce8fe987714a5c6ba1e18a704dc4503209 broke example codegen, since the "generate" directory caused the "make generate" rule not to run, since it already existed. It needs to be .PHONY. I'll refactor examples a bit later.


jamietanna:
Oops 😅 dang makefiles!

PR #2355, opened 2026-04-30
Update Go to 1.25.9, clean up modules, and update dependencies

mromaszewicz:
Update main package Go version to 1.25.9 and remove all submodules which only existed to test on more recent versions of Go.

Update almost all dependencies to their latest releases compatible with Go 1.25.9, except those which break API or behavior. Pinned github.com/dprotaso/go-yit to its existing version (latest breaks), and pinned github.com/kataras/golog v0.1.11 / github.com/kataras/pio v0.0.13 because newer combinations are incompatible with iris@v12.2.11.

Updated fiber test to provide an explicit Host header, since fiber now requires it.

CI lint stays on 1.26; build matrix targets 1.25 and 1.26.


jamietanna (commented):
LGTM, few notes:

- We've updated our /actions project, so we should be able to revert the changes to `ci.yml`
- IMO, for Go version upgrades, I'd recommend us bumping `go` in the top-level module, and then letting Renovate suggest the maximal version updates we can get
- Looks nicer to simplify things!

PR #2361, opened 2026-05-01
Update Greptile directives

mromaszewicz:
Tell Greptile to watch out for unintended dependency changes, and explain in CONTRIBUTING.md about using Greptile.


jamietanna (approved):
One thing we can always do in the future is we bump the minor `go` directive, then we get Renovate to update all the other dependencies in line with that

PR #2366, opened 2026-05-04
Add optional type generation for inline schemas

### LINKED ISSUE #1139: Generate a type for nested struct in respond instead of anonymous struct
#### 1. What version are you using?
```bash
❯ oapi-codegen --version
v1.13.0
```
#### 2. OpenAPI 3 spec:
```yaml
  /roles/{id}:
    summary: Get role by id
    get:
      tags: [ role ]
      parameters:
        - name: id
          in: path
          description: Role ID
          required: true
          deprecated: false
          allowEmptyValue: false
          schema:
            type: integer
      responses:
        '200':
          description: Role
          content:
            application/json:
              schema:
                allOf:
                  - $ref: '#/components/schemas/Successful-Response'
                  - properties:
                      data:
                        type: object
                        properties:
                          role:
                            $ref: '#/components/schemas/Role'
                        required: [ role ]
   
components:
  schemas:
    Successful-Response:
      type: object
      properties:
        ok:
          type: boolean
          default: true
          description: Indicated whether the response is successful.
        data:
          type: object
          description: If successful, response from api
      required:
        - ok
        - data                     
```

#### 3. Cmd arguments
```bash
oapi-codegen -config oapi.yaml -package api swagger.yaml
```

#### 4. oapi-codegen config

```yaml
package: api
generate:
  models: true
  fiber-server: true
  strict-server: true
  embedded-spec: true
output: api.gen.go
```

#### 5. Expected behavior
```go
type GetRolesId200JSONResponseData struct {
	Role Role `json:"role"`
}

type GetRolesId200JSONResponse struct {
	Data GetRolesId200JSONResponseData `json:"data"`

	// Ok Indicated whether the response is successful.
	Ok bool `json:"ok"`
}
```

#### 6. Actual behavior
```go
type GetRolesId200JSONResponse struct {
	Data struct {
		Role Role `json:"role"`
	} `json:"data"`

	// Ok Indicated whether the response is successful.
	Ok bool `json:"ok"`
}
```

With actual behavior there are only two ways to set data:

1.
```go
var resp api.GetRolesId200JSONResponse
resp.Ok = true
resp.Data.Role = role
```
2.
```go
resp = api.GetRolesId200JSONResponse{
	Ok: true,
	Data: struct {
		Role api.Role `json:"role"`
	}{
		Role: dto.Role(role),
	},
}
```
But I want to do like this
```go
resp = api.GetRolesId200JSONResponse{
	Ok: true,
	Data: api.GetRolesId200JSONResponseData{
		Role: dto.Role(role),
	},
}
```

I saw issues and some fixes, but they don't seem to solve my problem.

mromaszewicz:
Adds output-options.generate-types-for-anonymous-schemas (default false). When set, every inline schema that would otherwise generate as an anonymous Go struct is instead emitted as a named type with a path- derived name (e.g. GetRolesId200JSONResponseBody_Data). Equivalent to adding x-go-type-name to every inline schema; when both are present at the same site, x-go-type-name wins.

I will need to document this, since it's still better, from a usability perspective, to put concrete types of interest in component/schemas or as a second alternative, use x-go-type-name. However, this is requested often enough, that I'd thing it's worth exposing as a feature.

Test coverage reorganized: internal/test/anonymous_inner_hoisting/ now has two subdirectories. The existing always-on hoisting test (oneOf, anyOf, additionalProperties) moves to implicit/. The new flag-gated behavior gets its own test in global/, exercising the canonical issue shape (allOf-merged response with a sibling inline `data` object that itself contains a nested `role` reference).

Closes: #1139


jamietanna:
Given https://github.com/oapi-codegen/oapi-codegen/pull/648 was going to close https://github.com/oapi-codegen/oapi-codegen/issues/626, https://github.com/oapi-codegen/oapi-codegen/issues/636. they may be relevant here

internal/test/anonymous_inner_hoisting/global/spec.yaml
```diff
@@ -0,0 +1,58 @@
+openapi: "3.0.0"
+info:
+  title: Test inline struct hoisting
+  version: "1.0"
+paths:
+  /roles/{id}:
+    summary: Get role by id
+    get:
+      operationId: GetRolesId
+      tags: [ role ]
```
jamietanna: Nit:

```suggestion
      tags:
      - role
```

Makes future changes to files less noisy, diff-wise

jamietanna (commented):
Functionally it looks good - I wonder if we can improve the generated `Description` that gets added to it?

It's generally an existing issue, but especially where we're generating this new anonymous schema, it'd be great to improve

I..e we get:

```diff
 // AddMicrosoftTeamsChatTabTaskParams defines model for add_microsoft_teams_chat_tab_task_params.
 type AddMicrosoftTeamsChatTabTaskParams struct {
-       Chat struct {
-               ID   *string `json:"id,omitempty"`
-               Name *string `json:"name,omitempty"`
-       } `json:"chat"`
+       Chat AddMicrosoftTeamsChatTabTaskParams_Chat `json:"chat"`
 
        // Link The tab link
        Link     string                                      `json:"link"`
@@ -39399,15 +39402,18 @@ type AddMicrosoftTeamsChatTabTaskParams struct {
        Title string `json:"title"`
 }
 
+// AddMicrosoftTeamsChatTabTaskParams_Chat defines model for AddMicrosoftTeamsChatTabTaskParams.Chat.
+type AddMicrosoftTeamsChatTabTaskParams_Chat struct {
+       ID   *string `json:"id,omitempty"`
+       Name *string `json:"name,omitempty"`
+}
```

I wonder if we could instead generate it as i.e. 

```go
// AddMicrosoftTeamsChatTabTaskParams_Chat defines model for the anonymous schema `chat` under path ..., or something similar?
```

internal/test/anonymous_inner_hoisting/global/cfg.yaml
```diff
@@ -0,0 +1,7 @@
+package: global
+output: client.gen.go
+generate:
+  models: true
+  client: true
+output-options:
+  generate-types-for-anonymous-schemas: true
```
jamietanna, line 7: IMO, I'd prefer this by default - most folks don't control their specs, and so they get anonymous structs used which can be awkward

I'd personally swap this with a `compatibility-options` to disable this functionality

Using an OpenAPI Overlay is one way to get around this, but managing _all_ areas this may apply will be quite hard

internal/test/anonymous_inner_hoisting/global/spec.yaml
```diff
@@ -0,0 +1,58 @@
...
+    Role:
+      type: object
+      properties:
+        id:
+          type: integer
+        name:
+          type: string
+      required:
+        - id
+        - name
```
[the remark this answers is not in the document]
mromaszewicz: @greptileai, this is always a balancing act, between testing every possible combination, or just enough to exercise the functionality. The traversal into all the code locations that you mention is already tested in many tests, so I'm not worried about reaching this new code, and I just needed one test for it, since the generator is shared, irrespective of every path that it takes.

internal/test/anonymous_inner_hoisting/global/cfg.yaml
mromaszewicz: I have the opposite view.

In specs with lots of inline schemas, this positively litters the file with new type definitions, and in many cases, you don't need to ever refer to the type, so the code, IMO, is nicer. I'm open to defaulting it to be on, however, I dislike the output. 

Whenever you have an inline schema in a request body, or a parameter, you will never need to create an instance of that type, deserialization does that for you.

The place where it is nice to have is when you have inline schemas in things you are constructing; usually response bodies, and in those cases, it's so much nicer to declare those specifically. I realize that some people don't control their specs.

mromaszewicz:
> Functionally it looks good - I wonder if we can improve the generated `Description` that gets added to it?
> 
> It's generally an existing issue, but especially where we're generating this new anonymous schema, it'd be great to improve
> 
> I..e we get:
> 
> ```diff
>  // AddMicrosoftTeamsChatTabTaskParams defines model for add_microsoft_teams_chat_tab_task_params.
>  type AddMicrosoftTeamsChatTabTaskParams struct {
> -       Chat struct {
> -               ID   *string `json:"id,omitempty"`
> -               Name *string `json:"name,omitempty"`
> -       } `json:"chat"`
> +       Chat AddMicrosoftTeamsChatTabTaskParams_Chat `json:"chat"`
>  
>         // Link The tab link
>         Link     string                                      `json:"link"`
> @@ -39399,15 +39402,18 @@ type AddMicrosoftTeamsChatTabTaskParams struct {
>         Title string `json:"title"`
>  }
>  
> +// AddMicrosoftTeamsChatTabTaskParams_Chat defines model for AddMicrosoftTeamsChatTabTaskParams.Chat.
> +type AddMicrosoftTeamsChatTabTaskParams_Chat struct {
> +       ID   *string `json:"id,omitempty"`
> +       Name *string `json:"name,omitempty"`
> +}
> ```
> 
> I wonder if we could instead generate it as i.e.
> 
> ```go
> // AddMicrosoftTeamsChatTabTaskParams_Chat defines model for the anonymous schema `chat` under path ..., or something similar?
> ```

Ok, so the code which generates those comments is kinda broken. It calls everything a "parameter". I'm going to make a minimal fix to this PR to inject the correct string with the field name, eg, `AddMicrosoftTeamsChatTabTaskParams.Chat.`, however, the rest of the string will remain wrong. I'll file an Issue for the misleading comments and deal with the high level problems separately.

Update:
After exploring a bit, the fix for this comment is quite painful. I don't want it in this PR.

mromaszewicz:
Greptile had a good finding. This refactor breaks the splitting of code generation into (types, server, strict-server), etc, because of references among the types. Operation Definitions can introduce new types (the anonymous inner types), but those are only emitted when `types` are being emitted, so if I generate `types` without a server, then say, `server: echo` into another file, then we will have missing types.

the fix is non-trivial. It's a big refactor of the codegen structure.

pkg/codegen/schema.go
```diff
@@ -667,6 +667,33 @@ func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
...
+			typeDef := TypeDefinition{
+				TypeName: typeName,
+				JsonName: strings.Join(path, "."),
+				Schema:   outSchema,
+			}
+			outSchema = Schema{
+				Description:     typeDef.Schema.Description,
+				GoType:          typeName,
+				AdditionalTypes: append(outSchema.AdditionalTypes, typeDef),
+			}
```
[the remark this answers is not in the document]
mromaszewicz: Thanks, good find. Next patch fixes it.

jamietanna:
In https://github.com/rootlyhq/rootly-go/pull/131 I've noticed that some `nullable` entries are missing - might be intentional, but I'm taking a look

PR #2371, opened 2026-05-12
feat: add json-encoding output option for controlling JSON encoder settings

CRaLFa:
## Summary

Add `output-options.json-encoding` configuration block to control JSON encoding behavior in generated code.

- **`escape-html`** (bool, default `true`): maps to `json.Encoder.SetEscapeHTML`. When set to `false`, HTML special characters (`<`, `>`, `&`) are output as-is instead of being escaped to Unicode sequences.
- **`indent-prefix`** (string, default `""`): maps to the first argument of `json.Encoder.SetIndent`.
- **`indent`** (string, default `""`): maps to the second argument of `json.Encoder.SetIndent`. When non-empty, indented output is produced.

### Example config

```yaml
output-options:
  json-encoding:
    escape-html: false
    indent: "  "
```

### Affected generated code

| Location | Change |
|---|---|
| `MarshalJSON()` in union types | `json.Marshal(v)` → encoder with configured settings |
| `MarshalJSON()` in types with `additionalProperties` | same |
| `MarshalJSON()` in union + `additionalProperties` types | same |
| JSON request body in client | `json.Marshal(body)` → encoder with configured settings |
| `Visit*Response()` in strict server | `json.NewEncoder(&buf)` → encoder with configured settings |

When no options are set (default), `json.Marshal` / `json.NewEncoder` are emitted as before — **fully backward-compatible**.

Parameter serialization (path, query, header, cookie) is intentionally excluded: applying `SetIndent` there would embed whitespace into URL/header values and break requests.

## Test plan

- [x] `TestJSONEncodingOptions` covers: default, `escape-html: false`, `indent`, both options combined, and explicit `escape-html: true`
- [x] All existing tests pass unchanged
- [x] Generated code passes `go/format` validation in all cases

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/codegen/template_helpers.go
+++ b/pkg/codegen/template_helpers.go
@@ -422,22 +422,21 @@ func jsonMarshalExpr(enc JSONEncodingOptions, varName string) string {
 	if !enc.NeedsCustomEncoding() {
 		return "json.Marshal(" + varName + ")"
 	}
-	var sb strings.Builder
-	sb.WriteString("func() ([]byte, error) {\n")
-	sb.WriteString("var __buf bytes.Buffer\n")
-	sb.WriteString("__enc := json.NewEncoder(&__buf)\n")
+	var opts string
 	if !enc.EscapeHTMLValue() {
-		sb.WriteString("__enc.SetEscapeHTML(false)\n")
+		opts += "__enc.SetEscapeHTML(false)\n"
 	}
 	if enc.Indent != "" || enc.IndentPrefix != "" {
-		fmt.Fprintf(&sb, "__enc.SetIndent(%q, %q)\n", enc.IndentPrefix, enc.Indent)
+		opts += fmt.Sprintf("__enc.SetIndent(%q, %q)\n", enc.IndentPrefix, enc.Indent)
 	}
-	sb.WriteString("if __err := __enc.Encode(" + varName + "); __err != nil {\n")
-	sb.WriteString("return nil, __err\n")
-	sb.WriteString("}\n")
-	sb.WriteString("return bytes.TrimRight(__buf.Bytes(), \"\\n\"), nil\n")
-	sb.WriteString("}()")
-	return sb.String()
+	return fmt.Sprintf(`func() ([]byte, error) {
+var __buf bytes.Buffer
+__enc := json.NewEncoder(&__buf)
+%sif __err := __enc.Encode(%s); __err != nil {
+return nil, __err
+}
+return bytes.TrimRight(__buf.Bytes(), "\n"), nil
+}()`, opts, varName)
 }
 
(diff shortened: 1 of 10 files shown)


jamietanna:
I think @mromaszewicz we should look at how we want to expose more control ie with the underlying encoder/decoder we use, especially helping with ie `encoding/json/v2`

mromaszewicz:
Yes, this is a good question. I'd love to move entirely to `encoding/json/v2` but this has some implications.

Empty arrays in v2 are always marshaled as `[]` where v` can be `null` or `[]` based on if it's unset or empty. This will be a behavior change for us. Duplicated fields will error out now (where previously it was unspecified).

As for this PR, it looks good. It's a fine extension.

pkg/codegen/template_helpers.go
```diff
@@ -413,3 +413,63 @@ var TemplateFunctions = template.FuncMap{
...
+
+// jsonMarshalExpr returns a Go expression that marshals varName to JSON,
+// respecting the JSONEncoding options. When no custom encoding is required the
+// expression is a plain json.Marshal call; otherwise it is an immediately-
+// invoked function that configures a json.Encoder appropriately.
+func jsonMarshalExpr(enc JSONEncodingOptions, varName string) string {
+	if !enc.NeedsCustomEncoding() {
+		return "json.Marshal(" + varName + ")"
+	}
+	var sb strings.Builder
```
mromaszewicz: This is really hard to read. Can we just use a multi-line string literal?

mromaszewicz:
I've been wanting a feature like this for some time actually, but I'm on the fence about how it should be done nicely.

This PR will generate a JSON marshaler inline where needed based on configuration options.

Another option would be to allow users, at the code level, to hook into encoder creation, which I think is a cleaner option.

Instead of calling `json.Marshal` or `json.NewEncoder`, we would always use the encoder, to only have a single way of doing this, and in places where json needs to be encoded, we'd call something like `createJSONEncoder`, which is defaulted.

CRaLFa:
@mromaszewicz 

Thank you for the feedback and for the alternative idea.

The `createJSONEncoder` hook approach is interesting ― always emitting a single code path and letting users override the factory at runtime would indeed be cleaner from a generated-code perspective.

Would you prefer to pivot this PR to the hook-based design, or would it make more sense to merge the current config-driven approach and track the hook idea as a separate follow-up? Happy to go either direction.

pkg/codegen/template_helpers.go
CRaLFa: Fixed in the latest commit ― replaced the `strings.Builder` approach with `fmt.Sprintf` and raw string literals. Thanks for the suggestion!

PR #2372, opened 2026-05-16
Don't generate constants for undefined scopes

### LINKED ISSUE #2367: undefined security scheme context key type when securitySchemes component is missing
After upgrading to v2.7.0, generated code fails to compile when the spec uses security at the top level or operation level but does not define the corresponding entry in components/securitySchemes.

**Minimal spec to reproduce:**

```
openapi: "3.0.0"
info:
  title: Test
  version: 1.0.0
security:
  - api_key: []
paths:
  /hello:
    get:
      operationId: getHello
      responses:
        "200":
          description: ok
```

```
oapi-codegen -generate types,gin,strict-server -o gen.go -package api spec.yaml
```

**Generated output:**

```
const (
    Api_keyScopes apiKeyContextKey = "api_key.Scopes"
)
```
```
gen.go:20:16: undefined: apiKeyContextKey
```

The constant references `apiKeyContextKey` but the type definition `type apiKeyContextKey string` is never emitted.

Adding `components/securitySchemes` fixes it -- the type is generated correctly. But the tool should not produce uncompilable code for a spec that was previously accepted, wdyt?

mromaszewicz:
A spec that lists a scheme under top-level or operation `security:` without declaring it in `components/securitySchemes` previously emitted a constant like `Api_keyScopes apiKeyContextKey = "api_key.Scopes"` while the `apiKeyContextKey` type itself was never generated. The output failed to compile with `undefined: apiKeyContextKey`, and per-framework middleware templates emitted matching dangling `context.WithValue(ctx, Api_keyScopes, ...)` references when servers were generated.

Filter operation `SecurityDefinitions` against the set of schemes actually defined in `components/securitySchemes` in `OperationDefinitions`. Both the constants block and the per-framework middleware now stay in sync with the types that get emitted, so generated code compiles whether or not the spec defines every referenced scheme.

Closes: #2367


examples/go.sum
```diff
@@ -20,42 +20,31 @@ github.com/apapsch/go-jsonmerge/v2 v2.0.0 h1:axGnT1gRIfimI7gJifB699GoE/oq+F2MU7D
 github.com/apapsch/go-jsonmerge/v2 v2.0.0/go.mod h1:lvDnEdqiQrp0O42VQGgmlKpxL1AP2+08jFMw88y4klk=
 github.com/aymerick/douceur v0.2.0 h1:Mv+mAeH1Q+n9Fr+oyamOlAkUNPWPlA8PPGR0QAaYuPk=
 github.com/aymerick/douceur v0.2.0/go.mod h1:wlT5vV2O3h55X9m7iVYN0TBM0NH/MmbLnd30/FjWUq4=
-github.com/blang/semver/v4 v4.0.0/go.mod h1:IbckMUScFkM3pff0VJDNKRiT6TG/YpiHIM2yvyW5YoQ=
```
jamietanna, line 23: Why are these being removed? 🤔

mromaszewicz: I don't know. I run "make tidy" before staging every commit, and I don't ask questions when 'go.sum' changes, because it's never wrong :)

jamietanna: It's because I broke CI with 35a6e0c76d169aafd1a1f3b8a6ef0f1b17c0ab8d :)

jamietanna: Fix incoming

jamietanna: https://github.com/oapi-codegen/oapi-codegen/pull/2365

jamietanna: (Looks like the first recent commit that didn't run `make tidy` was fbc8e0dbb991a4d4c2928fce71400f338ece7228, FYI)

PR #2379, opened 2026-05-19
Add functions to response models for schema-extraction 

### LINKED ISSUE #2191: Add functions to response models for schema-extraction
At the moment, a generated response might contain a field named `JSON200` for the JSON response body of a 200 response, for example. It would be nice if there was also a simple getter function like `GetJSON200()`. I would like to create an interface which allows a standard handling of all responses and Go does not allow interfaces to have fields, only methods.

rd-andreas-tollkoetter:
closes #2191


mromaszewicz:
I'm addressing Greptile's findings and pushing an update to your fork.

mromaszewicz:
@greptileai, review latest patch

mromaszewicz (approved):

PR #2382, opened 2026-05-19
Echo V5 now uses the generic test framework

mromaszewicz:
When I updated to Go 1.25 and un-moduled the echo-v5 code, I left behind the special-cased test for echo-v5 which existed due to module boundaries. Since those modules don't exist anymore, we can remove the special casing.


mromaszewicz:
@greptileai

PR #2390, opened 2026-06-01
fix: use raw enum values for cross-enum conflict detection

### LINKED ISSUE #2391: bug: cross-enum conflict detection is order-dependent due to GetValues() returning prefixed names
`GenerateEnums` prefixes enum constant names when two enums share a value (e.g. `"running"` in both `instance-state` and `volume-state` becomes `InstanceStateRunning` / `VolumeStateRunning`). The conflict check calls `e1.GetValues()`, which returns already-prefixed names for enums marked in an earlier iteration. Those prefixed names no longer match the raw values of later enums, so some conflicts are missed and affected enums stay unprefixed.

Whether an enum gets prefixed ends up depending on iteration order rather than actual value overlap.

**Minimal reproduction:** three string enums where A shares `"running"` with B and `"migrating"` with C. A and B are compared first and both get marked. When A is later compared with C, `A.GetValues()` returns `["AStateRunning", "AStateMigrating"]`, neither of which matches C's raw value `"Migrating"`, so C stays unprefixed.

**Fix:** compare `Schema.EnumValues` keys instead of `GetValues()` output. `EnumValues` always holds unprefixed identifiers regardless of whether the enum has been marked yet. PR: #2390

natalie-o-perret:
Fixes #2391.

Cross-enum conflict detection uses `e1.GetValues()` to check for shared values, but `GetValues()` returns prefixed names for enums already marked in an earlier loop iteration. Those prefixed names don't match the raw values of later enums, so some conflicts are silently missed, and whether an enum gets prefixed depends on iteration order rather than actual value overlap.

Fix: compare `e1.Schema.EnumValues` keys instead. These are always the unprefixed identifiers, regardless of whether the enum has been marked yet.

Regression test added (`TestEnumConflictDetectionOrderIndependent`): three string enums where A↔B and A↔C both conflict, but the A↔C conflict was previously missed once A was already prefixed.

Discovered via a downstream build failure after an upstream schema change removed a schema that had been the only direct conflict anchor for one of the enums.


natalie-o-perret:
Summoning @jamietanna for review

mromaszewicz:
CI failed because your change affects existing generated code. I pulled your PR locally and I'm investigating, but we might need to feature-flag this to avoid breaking existing users.

mromaszewicz:
I had Claude dig into this locally:


## Why the files change

This PR changes conflict detection from comparing **generated constant names** (`GetValues()`, which reflects prefixing) to comparing **raw schema values** (`Schema.EnumValues`). That fixes the order-dependence in #2391, but it also drops a *second* class of conflict the old code happened to catch: an enum whose **raw value equals another enum's prefixed constant name**.

### `internal/test/issues/issue-1189` — correct improvement ✅

Three enums all share raw values `[bar, foo]`: `TestFieldA1`, `TestFieldB`, `TestFieldC1`. All three should be prefixed. The old order-dependent code marked A1 and B, then — because `GetValues()` returned their already-prefixed names — failed to see that C1 still collided, leaving it unprefixed (`Bar`/`Foo`). The new code compares raw values, so C1 is now correctly prefixed to `TestFieldC1Bar`/`TestFieldC1Foo`. This is exactly the bug #2391 describes. Compiles fine, output is now consistent.

### `internal/test/components` — regression ❌ (won't compile)

This spec was deliberately built to test the *other* collision class. From `components.yaml`:

- `Enum1: [One, Two, Three]` — conflicts with Enum2, so it's prefixed → constant `Enum1One = "One"`
- `Enum3: [Enum1One, Foo, Bar]` — its **raw value is literally `Enum1One`** (the comment even says *"Enum values conflict with Enums above, need to be prefixed"*)

The old `GetValues()` code compared *post-prefix* names, so it saw Enum1's prefixed constant `Enum1One` collide with Enum3's value `Enum1One` and prefixed Enum3 → `Enum3Enum1One`. The new raw-value code only sees Enum3's raw values `[Enum1One, Foo, Bar]` vs Enum1's raw `[One, Two, Three]` — no overlap — so Enum3 is left unprefixed. Result:

```go
Enum1One Enum1 = "One"        // line 16
Enum1One Enum3 = "Enum1One"   // line 59  ← duplicate
```

```
components/components.gen.go:59:2: Enum1One redeclared in this block
components/components.gen.go:68:7: invalid case Enum1One in switch ... (mismatched types Enum1 and Enum3)
```

A correct fix needs to detect both classes — raw-value overlap and the case where one enum's value matches another enum's would-be prefixed constant name —
  while remaining order-independent. The current PR can't land as-is because internal/test/components no longer compiles.

mromaszewicz:
@natalie-o-perret  - This code is all extremely sensitive. 

We can make new behavior default-on, but there definitely needs to be a flag to revert to current behavior, since I don't see a way of fixing this properly without breaking some old boilerplate.

natalie-o-perret:
@mromaszewicz

> CI failed because your change affects existing generated code. I pulled your PR locally and I'm investigating, but we might need to feature-flag this to avoid breaking existing users.

Thanks for the thorough write-up. I think you may have pulled before the second force push. 

The current commit uses a 2-stage algorithm that handles both conflict classes.

> ### `internal/test/issues/issue-1189` -- correct improvement ✅

Agreed, that's exactly the bug this PR is fixing.

> ### `internal/test/components` -- regression ❌ (won't compile)

This is handled by Stage 2. After Stage 1 marks Enum1 as prefixed (it shares raw values with Enum2), Stage 2 runs an iterative `GetValues()` comparison. At that point `Enum1.GetValues()` returns `{"Enum1One": "One", ...}` while `Enum3.GetValues()` (still unprefixed) returns `{"Enum1One": "Enum1One", ...}`. The key `Enum1One` appears in both, so Enum3 gets marked for prefixing too.

`go test ./...` passes and regenerating `internal/test/components` produces `Enum3Bar`, `Enum3Enum1One`, `Enum3Foo` with no compile errors. Could you try pulling again? Happy to debug further if you're still seeing the issue.

--- 


> @natalie-o-perret - This code is all extremely sensitive.
> 
> We can make new behavior default-on, but there definitely needs to be a flag to revert to current behavior, since I don't see a way of fixing this properly without breaking some old boilerplate.

Fair point, I've added a `compatibility.old-enum-conflict-detection: true` flag (latest commit `e13b895`) that restores the pre-v2.7.1 `GetValues()`-based behavior. New behavior stays default-on; users who need to preserve existing generated output can opt out.

mromaszewicz:
@greptileai re-review.

pkg/codegen/codegen.go
```diff
@@ -1021,25 +1021,87 @@ func GenerateEnums(t *template.Template, types []TypeDefinition) (string, error)
...
-					enums[i] = e1
-					enums[j] = e2
-					break
+	if globalState.options.Compatibility.OldEnumConflictDetection {
+		// Legacy behavior (pre-v2.7.1): compare generated constant names via
+		// GetValues(). This is order-dependent because GetValues() reflects
+		// already-applied prefixes, so enums processed later may miss conflicts
+		// with enums that were prefixed in an earlier iteration. Preserved here
+		// as an escape hatch for users who need to keep existing generated output.
+		for i := range enums {
```
mromaszewicz, line 1055: If I understand this right, the next loop completely supersedes this one, so this one doesn't even need to run.

pkg/codegen/configuration.go
```diff
@@ -299,6 +299,14 @@ type CompatibilityOptions struct {
...
 	// Please see https://github.com/oapi-codegen/oapi-codegen/issues/2267
 	HeadersImplicitlyRequired bool `yaml:"headers-implicitly-required,omitempty"`
+
+	// OldEnumConflictDetection reverts enum conflict detection to the pre-v2.7.1
+	// behavior, which compared generated constant names (GetValues()) rather than
+	// raw schema values. The old approach was order-dependent: whether an enum got
+	// prefixed could vary based on map iteration order (non-deterministic in Go).
+	// Set this only if you need to preserve existing generated output while you
+	// migrate. Please see https://github.com/oapi-codegen/oapi-codegen/issues/2391
+	OldEnumConflictDetection bool `yaml:"old-enum-conflict-detection,omitempty"`
```
mromaszewicz: This is getting confusing :) There's a `old-enum-conflicts` just above.

natalie-o-perret: Sorry :woman_facepalming:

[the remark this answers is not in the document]
natalie-o-perret: It's addressed by the latest commit which now reuses `old-enum-conflicts` instead.

natalie-o-perret: Folded into `old-enum-conflicts` in 204fe57, the legacy `GetValues()`-based path now runs when that flag is set. 

Since `old-enum-conflicts` is already in `configuration-schema.json`, no schema update needed either.

pkg/codegen/codegen.go
natalie-o-perret: Not quite, Stage 2 alone misses `CState `in the `AState`/`BState`/`CState` case. 

Once Stage 2's first pass prefixes `AState` because it shares "running" with `BState`, `AState.GetValues()` returns `{AStateRunning, AStateMigrating}`. 

When it then checks `AState` vs `CState`, "migrating" in `CState` no longer matches any key in `AState`'s now-prefixed values, so `CState` is left unprefixed. 

I verified this by removing Stage 1 and running the tests, both `TestEnumConflictDetectionOrderIndependent` and `TestEnumConflictDetectionBothOrders` fail with Migrating `CState` `=` `"migrating"` in the output.

mromaszewicz:
Sorry for the delay, I am revisting this now.

We can't change the meaning of old flags if we can help it.

I will push a change to default-on your change, but allow disabling it via `disable-enum-value-conflict-resolution`

mromaszewicz (approved):

PR #2398, opened 2026-06-05
Use net/http constants in boilerplate

### LINKED ISSUE #2397: Use constants from net/http for the status codes
When generating the api.gen.go, we only see one constant from the net/http package:

```
w.WriteHeader(http.StatusNotImplemented)
```

All other places it just uses the raw status code integer value from the spec:

```
w.WriteHeader(504)
```

Would be helpful to get a readable constant name instead of the integer,
especially when you use more than the typical 200 and 404 status codes.

mromaszewicz:
Closes: #2397

Replace raw integer literals in generated strict server code with named net/http constants (e.g. http.StatusOK instead of 200). Covers WriteHeader/Status/StatusCode calls in all strict server templates and status code comparisons in generated client switch statements.

### DIFF
--- a/pkg/codegen/operations.go
+++ b/pkg/codegen/operations.go
@@ -628,6 +628,83 @@ func (r ResponseDefinition) HasFixedStatusCode() bool {
 	return err == nil
 }
 
+// HttpStatusConstant returns the net/http constant name for a fixed status code (e.g. "http.StatusOK"),
+// or the raw integer string if no named constant exists.
+func (r ResponseDefinition) HttpStatusConstant() string {
+	return httpStatusConstant(r.StatusCode)
+}
+
+func httpStatusConstant(code string) string {
+	constants := map[string]string{
+		"100": "http.StatusContinue",
+		"101": "http.StatusSwitchingProtocols",
+		"102": "http.StatusProcessing",
+		"103": "http.StatusEarlyHints",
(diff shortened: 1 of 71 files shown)


mromaszewicz:
@greptileai, please look again, comments addressed.

pkg/codegen/operations.go
```diff
@@ -628,6 +628,83 @@ func (r ResponseDefinition) HasFixedStatusCode() bool {
 	return err == nil
 }
 
+// HttpStatusConstant returns the net/http constant name for a fixed status code (e.g. "http.StatusOK"),
+// or the raw integer string if no named constant exists.
+func (r ResponseDefinition) HttpStatusConstant() string {
+	return httpStatusConstant(r.StatusCode)
+}
+
+func httpStatusConstant(code string) string {
```
jamietanna, line 637: Might be worth a test to help us catch when new statuses are added to the standard library, if straightforward?

jamietanna (approved):
Agreed this doesn't really provide too much additional value, but it can help consistency that a lot of folks' "production code" would do

pkg/codegen/operations.go
mromaszewicz: We'd have to grep the standard library source. I think being lazy here is ok. The fallback to not using the constant will work, and we can address as-needed.

PR #2400, opened 2026-06-06
test: use Petstore spec from GitHub

jamietanna:
We're seeing intermittent issues from the Petstore project such as:

    error loading swagger spec in spec.yaml
    : failed to load OpenAPI specification: error resolving reference
    "https://petstore3.swagger.io/api/v3/openapi.json#/components/schemas/Pet":
    error loading "https://petstore3.swagger.io/api/v3/openapi.json":
    request returned status code 403

We can try and switch to the raw spec in GitHub to do this lookup, to
avoid hitting the Petstore API.

It's unclear if GitHub are going to be more performant, but it may help.

### DIFF
--- a/internal/test/externalref/externalref.cfg.yaml
+++ b/internal/test/externalref/externalref.cfg.yaml
@@ -6,7 +6,7 @@ generate:
 import-mapping:
   ./packageA/spec.yaml: github.com/oapi-codegen/oapi-codegen/v2/internal/test/externalref/packageA
   ./packageB/spec.yaml: package_b github.com/oapi-codegen/oapi-codegen/v2/internal/test/externalref/packageB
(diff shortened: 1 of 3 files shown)


mromaszewicz:
I don't think that unit tests should be fetching stuff over the network at all. We can do that entirely with local files, but change looks good as-is.

This isn't a performance issue, HTTP/403 means we're tripping over some kind of auth or rate limiting

mromaszewicz (approved):

jamietanna:
Yep, agreed that we shouldn't be hitting external services as part of our own builds (local or otherwise) and instead vendoring files

Relatedly: https://github.com/oapi-codegen/oapi-codegen/issues/1564

Re performance, I mean more that these calls to GitHub may be more likely to be slow and/or hit issues given their recent availability

PR #2407, opened 2026-06-14
feat(client): improve generated Godoc

jamietanna:
I've noted this as "breaking" as it changes the way that an existing templating function works

### DIFF
--- a/pkg/codegen/templates/chi/chi-interface.tmpl
+++ b/pkg/codegen/templates/chi/chi-interface.tmpl
@@ -1,6 +1,6 @@
 // ServerInterface represents all server handlers.
 type ServerInterface interface {
-{{range .}}{{if not .IsAlias}}{{.SummaryAsComment }}
+{{range .}}{{if not .IsAlias}}{{.SummaryAsComment .OperationId }}
 // ({{.Method}} {{.Path}})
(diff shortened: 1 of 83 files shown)


jamietanna:
I've split https://github.com/oapi-codegen/oapi-codegen/issues/2406 out of this one

mromaszewicz:
I think it's a nice looking improvement in terms of comments. The templates get a tiny bit more complex, but I think that's ok.

mromaszewicz:
@greptileai please review this draft.

pkg/codegen/templates/chi/chi-interface.tmpl
```diff
@@ -1,6 +1,6 @@
 // ServerInterface represents all server handlers.
 type ServerInterface interface {
-{{range .}}{{if not .IsAlias}}{{.SummaryAsComment }}
+{{range .}}{{if not .IsAlias}}{{.SummaryAsComment .OperationId }}
```
jamietanna, line 3: Could we use `toGoComment`?

jamietanna:
@greptileai review

jamietanna:
@greptileai review

mromaszewicz:
I've removed "breaking change" since we've generally not bothered about flagging template changes. Pretty much everything is a breaking change if we follow that convention.

PR #2411, opened 2026-06-18
fix: correctly generate consistent media type for type aliases

jamietanna:
Without this change, we'll generate i.e.

    ApplicationvndApiJSON422 *CreateItem422ApplicationvndApiJSONResponseBody

Instead of

    ApplicationvndApiJSON422 *CreateItem422ApplicationVndAPIPlusJSONResponseBody

As the type doesn't exist, this introduces a compilation error.

We can instead make sure we're consistently generating the media type
name.

Co-authored-by: Claude Sonnet 4.6 <jamie.tanna+claude-code@rootly.com>


jamietanna:
A follow-up may be needed here

Issue #1189, opened 2023-08-08
Not generating all model types


beeblebrox:
Hi there, I'm having what looks like a very similar issue not generating all the mode types as fixed in #1177 and discussed in #1168. I'm using version **v1.13.4**.  

Not all the types are being generated for https://github.com/openai/openai-openapi/blob/master/openapi.yaml resulting in compilation errors.   Sorry, I don't understand the issue enough to minimize the use case(s) that is failing.  I am using `oapi-codegen -package api -o openapi.gen.go openapi.yaml ` to generate the code.   I apologize if there is some gotchya I'm not aware of that I missed.

There are many const types being affected:
```go
// Defines values for CreateChatCompletionRequestFunctionCall0.
const (
	Auto CreateChatCompletionRequestFunctionCall0 = "auto"
	None CreateChatCompletionRequestFunctionCall0 = "none"
)
```
Auto and None are not defined anywhere.
Another interesting path to note is:
```go
const (
	CodeDavinci002 CreateCompletionRequestModel1 = "code-davinci-002"
	TextAda001     CreateCompletionRequestModel1 = "text-ada-001"
	TextBabbage001 CreateCompletionRequestModel1 = "text-babbage-001"
	TextCurie001   CreateCompletionRequestModel1 = "text-curie-001"
	TextDavinci001 CreateCompletionRequestModel1 = "text-davinci-001"
	TextDavinci002 CreateCompletionRequestModel1 = "text-davinci-002"
	TextDavinci003 CreateCompletionRequestModel1 = "text-davinci-003"
)
```
where `CreateCompletionRequestModel1` is defined but is a type alias of a non-defined type:
```go
// CreateCompletionRequestModel1 defines model for .
type CreateCompletionRequestModel1 CreateCompletionRequestModel

```
I do see the type define under componets:schema.  I tried removing some more advanced schema usages like the linked references and aliases on the type but it didn't seem to make a difference.

Unlike the other issue where if the types were in the response, most of these types are used in request bodies.  I don't know if that is relevant or not to the issue.

I haven't tried an older version yet, but that's my next step.

beeblebrox:
I just tried with **v1.13.2** and a 1.12 version, same results so this isn't a regression related to the last few changes.

ShouheiNishi:
We try to create simplified test case for this issue.

```
openapi: 3.0.0
paths:
  /test:
    get:
      operationId: Test
      responses:
        "200":
          content:
            application/json:
              schema:
[...]

PR #2418, opened 2026-06-20
Detect partial enum idiom

### LINKED ISSUE #2412: Allow `anyOf` with an `enum` (to indicate "a value in this enum or a plain string") to still render the enum
i.e.

```yaml
      anyOf:
        - type: string
        - type: string
          enum:
            - rootly
            - manual
            - api
            - heartbeat
            # ...
```

Will resolve as only being a `string`.

Ideally we would keep this being a `string`, but still generate enums

mromaszewicz:
Closes: #2412

An anyOf/oneOf of a bare primitive and that same primitive constrained by an `enum` ("a value from this enum, or any other value of the same type") previously rendered as a JSON union -- or, for a top-level component, as nothing at all -- discarding the enum's known values.

Detect this idiom and instead resolve the schema to the bare primitive (kept as a type alias, so any value stays assignable without a conversion) while still emitting the enum's known values as untyped constants. No distinct enum type and no Valid() method are generated, since any value of the primitive type is legal.

Detection is deliberately strict: it fires only on a oneOf/anyOf with exactly two inline primitive members of the same type and format, where exactly one carries an enum. Anything else -- a $ref member, three or more members, mismatched type/format, or allOf -- falls through to the existing union/merge behavior unchanged.

Implemented in GenerateGoSchema via untypedEnumUnion(); enum-value construction is factored into buildEnumValues() and shared with the existing enum branch. EnumDefinition gains ConstType and SkipValidate so constants.tmpl can emit primitive-typed constants and skip Valid() for this case. Adds the issue-2412 regression fixture.

What's broken: this is the same idiom #1189 fixed. At the property level, current releases generate a union wrapper (As*/From*/Merge* accessors); collapsing to a plain string removes that API, which is a breaking change for existing users. It is additive only for top-level components (which generate nothing today). For that reason the behavior is not yet gated -- it likely needs an opt-in output-options flag (default off) to preserve the #1189 union representation, to be revisited. See the discussion on #2412. The issue-1189 golden file is regenerated to reflect the new output.


mromaszewicz:
@greptileai review

PR #2425, opened 2026-06-22
Add Greptile logo to local logo cache

mromaszewicz:
I will be thanking Greptile for allowing their tool use in the next release notes.


jamietanna (approved):
Sounds good - I think as we'd chatted, let's prioritise the companies sponsoring, then Greptile - but very appreciative of the offering!

PR #2438, opened 2026-07-06
Use `any` instead of `interface`

mromaszewicz:
The `any` keyword has now been around for years, standardize on its use.

### DIFF
--- a/examples/import-mapping/samepackage/server.gen.go
+++ b/examples/import-mapping/samepackage/server.gen.go
@@ -262,7 +262,7 @@ func (sh *strictHandler) GetUserById(w http.ResponseWriter, r *http.Request, id 
 
 	request.Id = id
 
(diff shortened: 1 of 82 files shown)


pkg/codegen/templates/webhook-initiator.tmpl
```diff
@@ -137,7 +137,7 @@ func New{{$opid}}WebhookRequest{{.Suffix}}(targetURL string{{if $hasParams}}, pa
         }
         bodyReader = strings.NewReader(bodyStr.Encode())
     {{else if eq .NameTag "Text" -}}
-        if stringer, ok := interface{}(body).(fmt.Stringer); ok {
+        if stringer, ok := any(body).(fmt.Stringer); ok {
```
jamietanna: This doesn't look right 🤔 But I guess it is?

PR #2442, opened 2026-07-09
Update echo V5 petstore example

mromaszewicz:
Now that we have echo V5 middleware, use it, rather than recreating it locally.


mromaszewicz:
The socket security warning is a false positive. I'm trying to find how to squelch that in the future in their settings.

mromaszewicz:
@SocketSecurity ignore golang/golang.org/x/tools@v0.47.0

PR #2451, opened 2026-07-12
fix: render Godoc correctly for renamed + union types

jamietanna:
As otherwise we see `defines model for .`, which isn't correct.


jamietanna:
IMO, we should review this Godoc and improve it - I'm not quite sure if this is as effective/useful as it could be

mromaszewicz (approved):

mromaszewicz:
I agree. It's not very useful. It's very, very old code. It would be nice if it referred to the exact place in the spec that it came from, for example.

PR #2465, opened 2026-07-13
Register handlers in spec order

### LINKED ISSUE #1887: Re-order generated routes
Hello,

Here is my issue : I'm working with Fiber framework and I'm use to generate server side logic (fiber-server parameter). 
I try to know if there is a solution to order my routes because parameters can't be handled by the fiber router :

Generated :
```go
router.Get(options.BaseURL+"/api/v1/templates/privates/:id", wrapper.PrivateTemplateGet)
router.Get(options.BaseURL+"/api/v1/templates/:template_visibility/shortcuts", wrapper.TemplateShortcutGetAll)
```

should be : 
```go
router.Get(options.BaseURL+"/api/v1/templates/:template_visibility/shortcuts", wrapper.TemplateShortcutGetAll)
router.Get(options.BaseURL+"/api/v1/templates/privates/:id", wrapper.PrivateTemplateGet)
```
Router mess with parameter :template_visibility and
I've looked for some docs about order generation but I've found nothing. I also tried to change some values as name or re-order route declaration, that change nothing. 

Thank you

mromaszewicz:
Closes: #1887

Route registration was emitted in sorted path order, so reordering paths in the spec had no effect. On routers that match in registration order (Fiber, Gorilla/mux) this left users no way to disambiguate overlapping paths — e.g. a request for `/templates/privates/shortcuts` always went to `/templates/privates/{id}` rather than `/templates/{visibility}/shortcuts`.

kin-openapi stores paths in a map, losing declaration order at parse time, so enable its origin tracking (`Loader.IncludeOrigin`) to recover each path's source line onto the OperationDefinition (SpecOrder). Each GenerateXServer now renders its registration template with operations sorted by SpecOrder, via a new GenerateTemplatesIntoBuffer that composes passes into one buffer (GenerateTemplates wraps it).

Model and interface generation are intentionally left in the old sorted order to preserve any existing ordering behavior — in particular the order-dependent fallback in name-collision resolution. Only handler registration follows spec order.


jamietanna:
IMO, I think this may be an option we want a compatibility config option for @mromaszewicz, especially if it changes intent for everyone's specs

mromaszewicz:
This only affects fiber and gorilla, and only in the case where a parameterized shallow path comes before a deeper path.

For example, if you have:
- `/api/something/:param`
- `/api/something/anotherthing/foo`

The first is going to shadow the second.

In every other router, path registration is order independent, so it's a no-op change, just a bit of generated code churn.

I think I want this default-on, but I can add an opt out flag (even though spec editing can accomplish the same thing)

PR #2482, opened 2026-07-17
chore: use go mod tidy instead of tidied

alexandear:
The `tidied` is not needed, since we have `go mod tidy -diff`. From https://gitlab.com/jamietanna/tidied:

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,6 @@ readme-toc-check: $(GOBIN)/mdtoc
 
 tidy-ci:
 	# for the root module, explicitly run the step, to prevent recursive calls
-	tidied -verbose
+	go mod tidy -diff
 	# then, for all child modules, use a module-managed `Makefile`
 	git ls-files '**/*go.mod' -z | xargs -0 -I{} bash -xc 'cd $$(dirname {}) && make tidy-ci'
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -57,8 +57,4 @@ readme-toc: $(GOBIN)/mdtoc
 readme-toc-check: $(GOBIN)/mdtoc
 	$(GOBIN)/mdtoc --inplace --dryrun README.md
 
-tidy-ci:
-	# for the root module, explicitly run the step, to prevent recursive calls
-	tidied -verbose
-	# then, for all child modules, use a module-managed `Makefile`
-	git ls-files '**/*go.mod' -z | xargs -0 -I{} bash -xc 'cd $$(dirname {}) && make tidy-ci'
+tidy-ci: tidy
```
jamietanna: Still needs to run against all submodules

jamietanna:
Per Greptile - does this run `go mod tidy` then exit if its out of sync, or only checks if it's out of sync?

Edit: rereviewing should be OK as `tidy-ci` only affected

jamietanna:
We'll also need to remove https://github.com/oapi-codegen/actions/blob/0b10ad68a53148bd29896be82ad92027b35507a3/.github/workflows/ci.yml#L107-L108

jamietanna (approved):

PR #2492, opened 2026-07-21
fix(stdhttp): reject mixed path-parameter segments at codegen

### LINKED ISSUE #2488: stdhttp: mixed path parameter and literal suffix generates invalid ServeMux pattern
## Version

- `oapi-codegen`: `v2.8.0`
- Go: `go1.26.5`
- Generator: `std-http-server`

## Minimal reproduction

Configuration:

```yaml
package: api
output: server.gen.go
generate:
  std-http-server: true
```

OpenAPI specification:

```yaml
openapi: 3.0.3
info:
  title: Custom action reproduction
  version: 1.0.0
paths:
  /resources/{resourceId}:apply:
    post:
      operationId: applyResource
      parameters:
        - name: resourceId
          in: path
          required: true
          schema:
            type: string
      responses:
        "204":
          description: Applied
```

The generated `std-http-server` registration is equivalent to:

```go
m.HandleFunc(
	http.MethodPost+" "+options.BaseURL+"/resources/{resourceId}:apply",
	wrapper.ApplyResource,
)
```

Registering that pattern with `http.ServeMux` panics:

```text
panic: parsing "POST /resources/{resourceId}:apply": at offset 16: bad wildcard segment (must end with '}')
```

This can also be reproduced without generated code:

```go
mux := http.NewServeMux()
mux.HandleFunc(
	"POST /resources/{resourceId}:apply",
	func(http.ResponseWriter, *http.Request) {},
)
```

Although Go 1.22+ supports whole-segment wildcards, a mixed wildcard and literal
segment is not valid. Registering `/resources/{resourceId}` instead accepts the
request, but `PathValue("resourceId")` returns `abc:apply` for
`/resources/abc:apply`; `ServeMux` does not strip the action suffix.

## Expected behavior

For `std-http-server`, either:

1. generate a compatible route that matches the whole segment, validates and
   strips the literal suffix, and passes `abc` as `resourceId`; or
2. reject this path shape during generation with a clear unsupported-route
   error instead of generating code that panics during server startup.

Literal suffixes on templated path segments are useful for custom action routes
such as `{resourceId}:apply`.

## Additional context

The v2.8.0 trailing-slash and handler-registration-order fixes do not cover this
path shape. A local workaround is to register
`/resources/{resourceIdAction}`, dispatch based on the `:apply` suffix, and set
the original `resourceId` path value before invoking the generated wrapper.

The minimal reproduction snippets in this issue may be used under Apache-2.0.

Solaris-star:
## Summary
`std-http-server` previously generated ServeMux patterns for mixed wildcard+literal segments such as:

```
/resources/{resourceId}:apply
```

Registering that with `net/http.ServeMux` panics:

```
bad wildcard segment (must end with '}')
```

Maintainer guidance on #2488 prefers a **codegen-time error** over silent rewriting.

## Fix
- Add `ValidateStdHTTPPath` / `ValidateStdHTTPPaths`
- Run them from `Generate` when `StdHTTPServer` is enabled
- Reject any path segment that mixes a `{param}` with surrounding literals

## Test plan
- [x] `go test ./pkg/codegen/ -run 'TestValidateStdHTTPPath|TestGenerateRejectsMixedServeMuxPathParam|TestSwaggerUriToStdHttpUriUri'`

Fixes #2488

### DIFF
--- a/pkg/codegen/codegen.go
+++ b/pkg/codegen/codegen.go
@@ -178,6 +178,11 @@ func Generate(spec *openapi3.T, opts Configuration) (string, error) {
 	if err := ValidateSpec(spec); err != nil {
 		return "", err
 	}
+	if opts.Generate.StdHTTPServer {
+		if err := ValidateStdHTTPPaths(spec); err != nil {
+			return "", err
+		}
+	}
 
 	// if we are provided an override for the response type suffix update it
 	if opts.OutputOptions.ResponseTypeSuffix != "" {
--- a/pkg/codegen/utils.go
+++ b/pkg/codegen/utils.go
@@ -15,6 +15,7 @@ package codegen
 
 import (
 	"bytes"
+	"errors"
 	"cmp"
 	"fmt"
 	"go/token"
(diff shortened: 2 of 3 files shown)


mromaszewicz (approved):

mromaszewicz:
Thanks for doing this. I addressed the code review comments in your PR and I'll merge it.

PR #2520, opened 2026-08-14
feat(codegen): fold OpenAPI 3.1 const-based oneOf without outer type into enums

sloudel:
## Summary

OpenAPI 3.1 expresses string/integer enums as a `oneOf` whose branches each carry a scalar `const`. When the outer schema *omits* `type` (the common idiom, since the consts imply the family), `detectEnumViaOneOf` previously required an explicit outer scalar `type` and fell through to the union generator, emitting:

```go
type PetStatus struct {
    union json.RawMessage
}
type PetStatus0 = interface{}
```

losing the enum values and names entirely.

This change lets the scalar family be inferred from the per-branch `const` values when the outer `type` is absent:

```yaml
PetStatus:
  oneOf:
    - const: available
      title: Available
    - const: pending
      title: Pending
    - const: sold
      title: Sold
```

now produces a typed enum:

```go
type PetStatus string
const (
    Available PetStatus = "available"
    Pending   PetStatus = "pending"
    Sold      PetStatus = "sold"
)
```

The integer variant (`Port` with whole-number consts) likewise emits `Port int` + numeric constants.

## Behaviour

- **Outer `type` present** — unchanged (must be `string`/`integer` to be treated as an enum).
- **Outer `type` absent** — family inferred from the `const` values; every branch must agree and each const must be a plain scalar (string, or whole-number integer). Mixed/fractional/bool/object/array → falls through to the standard union path.
- The detected type drives the Go base-type mapping via a **shallow copy** of the schema (the caller's schema is never mutated), so `oapiSchemaToGoType` yields `string`/`int` instead of falling into the generic `interface{}` branch.
- `SkipEnumViaOneOf` escape hatch and the negative paths (missing `title`/`const`, object/mixed `oneOf`) still fall through to the union generator.

## Tests

- `TestEnumViaOneOfConstType` — unit coverage of the new inference helper (string / whole-number / fractional / bool / object / array / mixed).
- Integration: `PetStatus` (string, no type) and `Port` (integer, no type) added to `internal/test/openapi31/spec.yaml`, with named-constant + JSON round-trip assertions.
- Confirmed the existing with-type enums (`Severity`, `Color`) and the negative `MixedOneOf` case still pass.

## Verification

```
go build ./...
go vet ./...
go test ./pkg/codegen/ -run TestEnumViaOneOfConstType
go test ./internal/test/openapi31/   # (run from internal/test module)
```

all green.


sloudel:
Hi maintainers — could someone approve running the CI workflow on this PR? It's currently stuck at the fork-PR approval gate (the `CI` run shows `action_required` before any jobs are dispatched).

For context: the automated review note about *integer inference permitting overflowing constants* has already been addressed in code — commit `bcf47f0d` ("fix(codegen): reject enum-via-oneOf consts outside the int64 range") adds a `toInt64` guard, and the `TestEnumViaOneOfOverflowFallsThrough` unit test covers the overflow case.

Local verification passes: `go build ./...`, `go vet ./pkg/codegen/`, `gofmt` on the touched files, the full `pkg/codegen` suite, and the `openapi31` integration tests are all green.

Summary of the change: OpenAPI 3.1 expresses scalar enums as `oneOf` of `const` values. Without an outer `type`, the model generator previously fell back to a generic union / `interface{}`, so enum constants were never emitted. This PR folds that shape into a proper typed enum (respecting the reflection that a `oneOf` whose variants are all scalar `const`s is an enum), emits the `EnumValues`/`Valid()` helpers, and guards against emitting integer constants that overflow `int64`.

Thanks for taking a look.

mromaszewicz:
I've actually been playing with your PR this morning on my own.  I found a few edge cases (which is why my code that you fixed has a type declaration, and doesn't infer type). I'll comment in full later, busy right now. This will get merged, we just need to make some small fixes.

mromaszewicz:
I'm resolving the conflicts with main and pushing an update to your branch.

mromaszewicz:
kin-openapi parses all those values in the const using float64, so it's pointless to deal with uints. I'm going to use just `int` and `int64` for scalars, where appropriate.

mromaszewicz (approved):

Issue #2430, opened 2026-06-28
"null" type isn't handled for OpenAPI 3.1


atye:
```
type: object
properties:
  id:
    type: string
  url:
    type: string
    format: uri
  challenger:
    type: "null"
```
```
error generating code: error collecting component types: error generating Go types for component schemas: error converting Schema ChallengeOpenJson to Go type: error generating Go schema for property 'challenger': error resolving primitive type: unhandled Schema type: &[null]
```

mromaszewicz:
This is interesting. I'm not sure what `type: "null"` should map to in terms of a Go type. This seems to imply that `challenger` should always be set to `null`.

Issue #1977, opened 2025-05-10
Improve error message from `unhandled Schema type`


jamietanna:
As found with #1976, this doesn't seem to be the most helpful error - what can we do / how can we make it more actionable?

Issue #1328, opened 2023-10-18
Empty schema in response


ShouheiNishi:
When schema in content for response is empty, generated strict server codes are broken.

```
$
$ cat issue.yaml
openapi: "3.0.1"
paths:
  /test:
    get:
      operationId: test
      responses:
        200:
          description: good
          content:
            application/json:
              schema:
$ go run cmd/oapi-codegen/oapi-codegen.go -generate types,server,strict-server -package issue -o issue.go issue.yaml && go test issue.go
# command-line-arguments
./issue.go:78:7: invalid receiver type Test200JSONResponse (pointer or interface type)
FAIL
$
```

renom:
Is there a workaround?

jpeach:
FWIW I worked around this with
```
      responses:
        "201":
          content:
            application/json:
              schema:
                x-go-type: std.EmptyResponse
                x-go-type-import:
                  path: example.com/std
```

Then EmptyResponse looks like:
```
type EmptyResponse struct{}

func (e EmptyResponse) MarshalJSON() ([]byte, error) {
	return []byte{}, nil
}
```

jpeach:
> FWIW I worked around this with

Serves me right for posting before testing; the JSON marshaller returns an error if you return empty bytes; so this isn't actually a workaround :(

jpeach:
This seems to generate the right code:
```openapi: "3.0.1"
paths:
  /test:
    get:
      operationId: test
      responses:
        200: {}
```

PR #2522, opened 2026-08-14
Map OpenAPI 3.1 multi-type unions to `any`

### LINKED ISSUE #2521: Multi-type unions aren't handled for OpenAPI 3.1
OpenAPI 3.1 allows `type` to be a list, so a value may be any one of several types. Generation fails on any schema that uses one.

Spec:

```yaml
openapi: 3.1.0
info:
  title: multi-type union
  version: "1.0.0"
paths: {}
components:
  schemas:
    Event:
      type: object
      additionalProperties:
        type: [string, number, boolean]
```

Config:

```yaml
package: example
output: example.gen.go
generate:
  models: true
output-options:
  skip-prune: true
```

Error:

```
error generating code: error collecting component types: error generating Go types for component schemas: error converting Schema Event to Go type: error generating type for additional properties: error resolving primitive type: unhandled Schema type: &[string number boolean]
```

The position doesn't matter. `additionalProperties` is just where it happens to land here. A union in a property, parameter, body, or array items fails the same way, because the primitive-type dispatch in `pkg/codegen/schema.go` uses `Types.Is(...)`, which only matches a single-element `type` list.

Expected: `type Event map[string]any`. Go has no type meaning "one of these", so `any` is the same permissive mapping a bare `type: "null"` got in #2430.

This comes up in real specs. Honeycomb's published 3.1 spec has two of these for genuinely polymorphic event and query-result values, and the only workaround today is an overlay that deletes the `type` keyword before generation.

`oapi-codegen` version: current `main` (6768b63)
Go version: 1.26.5

I have a fix ready and will open a PR shortly.

bendrucker:
OpenAPI 3.1 allows `type` to be a list, so a value may be any one of several types. The primitive-type dispatch in `pkg/codegen/schema.go` uses `Types.Is(...)`, which only matches a single-element list, so every multi-type union fell through to `unhandled Schema type` and failed generation. Closes #2521.

Go has no type meaning "one of these", so a union maps to `any`. That follows the bare `type: "null"` branch directly above it, added for #2430, on the same reasoning: prefer the permissive mapping over rejecting an otherwise valid spec.

The mapping is deliberately narrow so the existing error keeps catching malformed input instead of quietly widening it. #1977 asks for a *better* error here, and swallowing more cases would be the wrong direction. A list-valued `type` is 3.1-only syntax, so a 3.0 document carrying one still fails. Every entry must name a JSON Schema type, so `type: [strng, number]` still fails on the typo.

One interaction needed handling. A union carrying an `enum` reached the enum branch, which emits `const X T = ...` against a type that is now `any`. That is not a valid Go constant, so generation succeeded but the output did not compile. Unions are now excluded from enum codegen for the reason `type: array` already is, and generate the plain `any` instead.

Nullable objects are unaffected. `schemaPrimaryType` strips `"null"` before the dispatch, so `type: [object, "null"]` still generates its struct. `type: [object, string]` does collapse to `any` and drops its declared properties. That is inherent to the mapping, since no Go type is either a struct or a string.

Generation is exercised across property, `additionalProperties`, parameter, request body, response body, array items, and `allOf` positions. One rough edge worth naming: a response body typed `any` under `strict-server` emits a method on a named interface type, which does not compile. That reproduces on `main` today for both `schema: {}` and `type: "null"`, so this PR routes one more spec shape into it rather than causing it. Happy to fix that separately if you'd like it tracked.

For motivation, Honeycomb's published 3.1 spec carries two of these for genuinely polymorphic event and query-result values. The only workaround today is an overlay that strips the `type` keyword before generation.


jamietanna (commented):
@mromaszewicz I'd expect that we'd treat this as a `oneOf` in this case?

mromaszewicz:
> @mromaszewicz I'd expect that we'd treat this as a `oneOf` in this case?

You know, I'm not sure. This isn't like `oneOf` because that can handle any schema definition, however, this is much more limited, because it's just top level types. I do think `any`, like in this PR, is a better fit. If someone wants something that is `[string, array, number]`, it's on them to disentangle what this is. `any` will have the json marshaler parse this as it sees fit, while if we make it something like `json.RawMessage`, then the user can decide how to parse it, so to me, it's a choice between `any` and `json.RawMessage`. I'm leaning towards `any`.

mromaszewicz:
Thanks for the PR — the core mapping is well-guarded and the writeup is unusually thorough. We ran a deep analysis with Claude on this change (probing generated output for the interaction cases against a compiled runtime), and it surfaced a few issues that make the change incomplete as-is. Findings below, roughly in order of severity.

### 1. Parameter positions generate code that always fails at request time

A multi-type union in a parameter now generates successfully, but the endpoint is permanently broken:

```yaml
parameters:
  - name: id
    in: path
    required: true
    schema:
      type: [string, integer]
```

The generated handler binds into an `any` destination, and `runtime.BindStyledParameterWithOptions` / `BindQueryParameter` unconditionally reject that:

```
error binding string parameter: can not bind to destination of type: interface
```

So a path param typed this way 400s on every request, and a query param 400s whenever it's supplied. Before this PR the spec failed loudly at generation time; after it, the failure moves to runtime. Unlike `type: "null"`, a string-or-integer ID is a plausible real spec, and this is the one position where `any` isn't merely permissive but non-functional. We'd want parameter schemas to either keep erroring at generation, or bind the raw string — not silently generate a dead endpoint.

Relatedly, the PR body says generation is exercised across "property, additionalProperties, parameter, request body, response body, array items, and allOf positions", but the committed suite only covers property, additionalProperties, a named component, and the enum interaction. If you tested the other positions locally, could you commit those tests? The parameter case above is exactly the kind of thing that coverage would have caught.

### 2. `type` list + `oneOf`/`anyOf` siblings: the composition is silently discarded

```yaml
UnionWithOneOf:
  type: [string, number]
  oneOf:
    - type: string
    - type: number
```

generates `type UnionWithOneOf = any` with no union accessors, because `generateUnion` only runs inside the object/typeless branch of `GenerateGoSchema`. Not a regression (this errored before), but it's a legal 3.1 conjunction that now silently loses the `As…`/`From…` machinery a bare `oneOf` gets. At minimum this deserves a line in the README note.

### 3. allOf ordering now swallows unions silently

```yaml
AllOfUnionMember:
  allOf:
    - properties:
        name:
          type: string
    - type: [string, number]
```

generates a plain struct — the union member's type vanishes, because `MergeSchemas` does `result.Type = s1.Type` and discards s2's type when s1 has none. Reverse the member order and you get `any` with the properties dropped instead. The order-dependency is a pre-existing `MergeSchemas` bug (single types behave the same on `main`), so it's not this PR's fault — but the PR converts some of these from hard errors into silent, order-dependent output. Worth either a guard or a tracking issue.

### Confirmed as fine

- The strict-server `any`-response compile failure reproduces on `main` with `schema: {}`, exactly as the PR body says — pre-existing, not introduced here.
- Nullability handling is correct throughout: `[string, number, "null"]` properties render bare `any` without `omitempty`, and `nullable.Nullable[any]` under the nullable-type option.
- The enum guard is correctly placed, the 3.0/typo rejection works as described, and `x-go-type` still wins as an escape hatch since extensions are handled before type dispatch.
- A union as a `oneOf` *branch* works (accessor typed `any`), with the caveat that `As…` can never fail, so branch discrimination degrades — same as typeless branches today.

Overall: the schema-position mapping looks merge-worthy, but we'd like the parameter story resolved (and the position tests committed) before this lands.

mromaszewicz:
Update on the review findings above: the runtime side of finding 1 has shipped. [runtime v1.7.0](https://github.com/oapi-codegen/runtime/releases/tag/v1.7.0) adds a `Types []string` field to `BindStyledParameterOptions` / `BindQueryParameterOptions`, consulted only when the destination is an `any`: the value binds to the first union member that parses, in specificity order (boolean, integer, number, string) with JSON-grammar numeric detection, so union-typed parameters can now bind correctly instead of 400ing on every request.

What that leaves on the codegen side to complete the parameter story here:

1. Emit `Types: []string{...}` in the bind-option literals for multi-type union parameters — the bind call sites span the stdlib/echo/gin/fiber/iris templates.
2. Fix `ParameterDefinition.SchemaType()`, which emits the first entry of the type list (`Type: "string"` for `[string, integer]`, and `"null"` for `["null", "string"]`). Harmless while `Type` is unread, but wrong metadata to keep generating.
3. Bump the test modules' runtime dependency to v1.7.0 and add parameter positions to the `openapi31` conformance suite with bind round-trips (the client side needs no runtime change — serialization reflects on the concrete value — but a round-trip test would pin both directions).

We're happy to either take these as a follow-up PR on our side, or you're welcome to fold them into this one — whichever you prefer. If you'd rather keep this PR at its current scope, we'll handle the plumbing after merge so both land in the same release.

On the remaining findings: having dug further, finding 2 (`type` list + `oneOf`/`anyOf` siblings) turns out to be mostly theoretical — in any coherent spec the sibling `type` list is redundant with the branch types (validators intersect; linters flag it), disagreement between them makes members unsatisfiable, and every such spec fails generation on `main` today anyway, so there's nobody to regress. We're treating it as a non-issue. Finding 3 (the `MergeSchemas` order dependency) is pre-existing and we'll take care of it, along with the strict-server `any`-response compile issue, on our side — none of that needs to block this PR.

mromaszewicz:
Ok, @bendrucker , what do you think of this change. I added a commit which makes it much larger, since I had to change every template to pass through the OpenAPI 3.1 types array to the runtime package.

mromaszewicz (approved):

bendrucker:
Just getting back to this, definitely didn't anticipate those edge cases, thanks for finishing this one for me! Will take a look back through and see if there are any follow ups, might tackle #2525 as well.

bendrucker:
Revising my earlier comment: #2525 looks like the same bug as #1328, and #2514 already fixes it with the struct wrapper suggested in the issue. I reviewed that branch merged onto current main and verified it handles every shape from #2525, including the 3.1 unions from this PR. Details on the issue.

PR #2537, opened 2026-08-20
feat(codegen): migrate embedded spec compression to `compress/lzw`

### LINKED ISSUE #2536: Go 1.27 produces different answer for the embedded spec
See https://github.com/oapi-codegen/oapi-codegen/pull/2534 - when we run in Go 1.27, we see i.e.

```diff
diff --git examples/go.mod examples/go.mod
index 6f0d0f8a..176a3ddb 100644
--- examples/go.mod
+++ examples/go.mod
@@ -2,6 +2,8 @@ module github.com/oapi-codegen/oapi-codegen/v2/examples
 
 go 1.25.0
 
+toolchain go1.27.0
+
 replace github.com/oapi-codegen/oapi-codegen/v2 => ../
 
 require (
diff --git examples/overlay/api/ping.gen.go examples/overlay/api/ping.gen.go
index 8c2cfb4a..f2caa365 100644
--- examples/overlay/api/ping.gen.go
+++ examples/overlay/api/ping.gen.go
@@ -177,13 +177,13 @@ func HandlerWithOptions(si ServerInterface, options GorillaServerOptions) http.H
 // const string: with thousands of chunks the chained `+` fold is several
 // times slower for the Go compiler than parsing a slice literal.
 var swaggerSpec = []string{
-       "XFLBbtQwEP0Va+AAUna97YqLbxVCaMVhV4Jb1YPrTGKXxGPsSWlV+d/ROLsUepp48uz33sx7AUdzooiR",
-       "C5gXKM7jbNvnieIoNWVKmDlg66awdvHJzmlCMJAE1wE/JzkVzoKotYOMv5aQsQdzu167+4ui+wd0DB08",
-       "bUbaRDtL8/iIOYe+x9ioq7wR4kBCx4Eb2ZeVVjGpEPvgLKPy9FvOS0HFHtUxYbw5HZQ8N9lnVRK6MAg0",
-       "UFQfPHMqRusxsF/ut45mfbw56DN68/1f9Efo4BFzCRTBwNV2t901zXQG25SmIAbF9uJ4ydhffkLtgBJG",
-       "mwIY2J/vJsu+zVHbfg5R24WpODutU60d6MuAR2QpPRaXQ+JVwmeP7qdib7lZFZuhqLzEGOKojt+gceYm",
-       "/iC6viKfQttPxpIolnWL17udFEeRMTaeZmV1rR+KkF3CIF/vMw5g4J1+TYs+R0W/Lut/rSfMA+VZWSWW",
-       "lKQBC7dkFMwyVjC3bw3+8Kh6HOwysVpRW+hgyRMYkM0ZrSdydvJU2Ow/XV3vod7VWuufAAAA//8=",
+       "XFJBb1MxDP4rkeEA0mvTreKS24QQqji0EtymHbI8t8l4Lw6J39g05b8jJx2FnZw4X/x9n+0XcDQnihi5",
+       "gHmB4jzOth0PFE8SU6aEmQO2bAo9i092ThOCgSS4Afg5ya1wFkStA2T8tYSMI5jb/u3uL4ruH9AxDPC0",
+       "OtEq2lmS+0fMOYwjxkZdpUaIRxI6DtzIvnRaxaRCHIOzjMrTb7kvBRV7VPuE8eawU1Juss+qJHThKNBA",
+       "UX3wzKkYrU+B/XK/djTr/c1On9Gr7/+iP8IAj5hLoAgGrtab9aZppjPYpjQFHLvtxfGScXx9hDoAJYw2",
+       "BTCwPf9Nln3ro7bjHKK2C1NxdupdrQPo1wafkCWMWFwOibuEzx7dT8XecrMqNkNReYkxxJPaf4PGmZv4",
+       "nej6inwIbT4ZS6JY+hSvNxsJjiJjbDzNSnetHwrFyzLI6X3GIxh4py/bovtr0Zdh/a/1gPlIeVZWiSUl",
+       "24CF22YUzNJWMLdvDf7wqEY82mVi1VFrGGDJExiQyRmtJ3J28lTYbD9dXW+h3tVa658BAA==",
 }
```

At least for our CI, this should be made stable across versions - possibly with a Compatibility Option.

For users, it could be there's a valid reason this is happening, and they may want what looks like potentially better compression

jamietanna:
- refactor(codegen): extract `compressSpec` helper from `GenerateInlinedSpec`

  When we embed the OpenAPI spec as an "embedded spec" in the generated
  code, we compress its JSON form (with `compress/flate`) and then
  Base64-encode the output.

  While upgrading to Go 1.27, #2536 noticed that there is now a diff in
  the generated code, due to a slight compression improvement in
  `compress/flate`, which means that any Go 1.27 users will now see a
  diff.

  To provide a more stable output across Go versions, we can change our
  compression algorithm.

  Before we do this, we can extract a `compressSpec` helper function and
  add a test to cover the existing compression, which will fail on Go
  1.27.

- feat(codegen): migrate embedded spec compression to `compress/lzw`

  When we embed the OpenAPI spec as an "embedded spec" in the generated
  code, we compress its JSON form (with `compress/flate`) and then
  Base64-encode the output.

  While upgrading to Go 1.27, #2536 noticed that there is now a diff in
  the generated code, due to a slight compression improvement in
  `compress/flate`, which means that any Go 1.27 users will now see a
  diff.

  To provide a more stable output across Go versions, we can change our
  compression algorithm.

  Through testing, `compress/lzw` appears to be stable across Go 1.26 and
  Go 1.27, so we can switch this over.

  Closes #2536.

### DIFF
--- a/examples/authenticated-api/echo/api/api.gen.go
+++ b/examples/authenticated-api/echo/api/api.gen.go
@@ -5,7 +5,7 @@ package api
 
 import (
 	"bytes"
-	"compress/flate"
+	"compress/lzw"
 	"context"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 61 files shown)


jamietanna:
I could see this as a `feat` or a `fix` depending on how we want to do it - I don't think too many folks are directly depending on the internals of how it's compressed, and should instead use the generated helper functions to decompress it

jamietanna:
Replaced by https://github.com/oapi-codegen/oapi-codegen/pull/2534

PR #2546, opened 2026-08-28
docs(sponsors): re-add Elastic as a sponsor

jamietanna:
As they're back to sponsoring us, thanks folks!


jamietanna:
Thanks @tobio for chasing this 💜

PR #2550, opened 2026-09-03
fix: resolve Go field name collisions when JSON keys differ only by separator (#2495)

### LINKED ISSUE #2495: oapi-codegen v2 produces duplicate Go field names when JSON keys differ only by separator (hyphen vs underscore)
### Issue
When two JSON properties in the same object differ only by their separator — e.g. "host-fqdn" and "host_fqdn" — both are converted to the same PascalCase Go field name HostFqdn, causing a redeclared in this block compile error. Since - is an invalid Go identifier character, stripping it is unavoidable. But _ is a valid Go identifier character and should be treated differently (e.g. preserved, or collision-detected and suffixed), rather than being normalized identically to -.

### Version:
oapi-codegen Version: v2.7.2

Spec source: 
https://developer.tenable.com/openapi/vulnerability-management.json
Section: 
GET /scans/{scan_uuid}/hosts/{host_id}/plugins/{plugin_id} response — info.host-fqdn and info.host_fqdn coexist in the same object.

### Spec snippet — Tenable Vulnerability Management API
```
"info": {
    "type": "object",
    "properties": {
        "host-fqdn": {
            "type": "string",
            "description": "The host's fully qualified domain name; optional."
        },
        "host_fqdn": {
            "type": "string",
            "description": "The FQDN of the host. Normally, this is populated with the value used to scan the host in the target list of the last scan that was ran where it was seen; always present."
        },
        "host-ip": {
            "type": "string",
            "description": "The host's IPv4 address; optional."
        }
    }
}
```
### Generated output:
```
// Both JSON properties produce the identical Go field name:
HostFqdn *string `json:"host-fqdn,omitempty"`   // from "host-fqdn"  ← line 39495
...
HostFqdn *string `json:"host_fqdn,omitempty"`   // from "host_fqdn" ← line 39507  DUPLICATE
```

### Expected behaviour
The generator should detect that two JSON property names produce the same Go identifier after normalization and either:
Preserve the underscore to produce a distinct name (e.g. Host_fqdn), or
Emit a descriptive error/warning directing the user to resolve the conflict via x-go-name
Producing uncompilable output silently is not acceptable.

rifkir23:
## Problem

When two JSON properties in the same object differ only by their separator — e.g. `host-fqdn` and `host_fqdn` — both normalize to the same Go field name `HostFqdn`, producing a `redeclared in this block` compile error in the generated code. `-` is not a valid Go identifier character so stripping it is unavoidable, and `_` folds to the same PascalCase, so the two collide.

This is reproducible against real specs (the issue references the Tenable Vulnerability Management API, whose `info` object contains both `host-fqdn` and `host_fqdn`).

Fixes #2495

## Fix

Resolve Go field-name collisions within a struct in `GenFieldsFromProperties`. The first property to claim a name keeps it; each later property that would collide gets an incrementing numeric suffix (`HostFqdn`, `HostFqdn2`, `HostFqdn3`, ...), re-checked so a suffixed name cannot itself collide. This matches the direction suggested in the issue ("suffix numbers on colliding fields") and is consistent with the collision-resolution the codebase already does elsewhere (enum constants, type names).

Only the Go identifier changes — the JSON struct tag still uses the original `JsonFieldName`, so the wire format is unchanged and round-trips exactly as before. Explicit `x-go-name` overrides continue to work and take precedence.

## Tests

Added `TestGenFieldsFromProperties_ResolvesGoFieldNameCollisions`: two properties (`host-fqdn`, `host_fqdn`) now generate `HostFqdn` and `HostFqdn2` with their original JSON tags preserved. Verified the test FAILS without the fix (both fields render as `HostFqdn`, matching the reported compile error) and PASSES with it. Full `pkg/codegen` suite passes; `gofmt` and `go vet` are clean.

### DIFF
--- a/pkg/codegen/schema.go
+++ b/pkg/codegen/schema.go
@@ -140,9 +140,19 @@ type Property struct {
 	NeedsFormTag  bool
 	Extensions    map[string]any
 	Deprecated    bool
+
+	// resolvedGoFieldName, when non-empty, overrides the computed Go field
+	// name. It is set by ResolvePropertyGoFieldNameCollisions so that when two
+	// properties would normalize to the same Go identifier, both the struct
+	// declaration and the generated marshalling boilerplate agree on the same
+	// (disambiguated) name. See issue #2495.
+	resolvedGoFieldName string
 }
 
 func (p Property) GoFieldName() string {
+	if p.resolvedGoFieldName != "" {
+		return p.resolvedGoFieldName
+	}
 	goFieldName := p.JsonFieldName
 	if extension, ok := p.Extensions[extGoName]; ok {
 		if extGoFieldName, err := extParseGoFieldName(extension); err == nil {
@@ -1317,6 +1327,10 @@ func GenerateGoSchema(sref *openapi3.SchemaRef, path []string) (Schema, error) {
 			// primitive's Go type and clears the struct-shaped fields;
 			// rebuilding `struct {}` here would clobber that.
 			if len(outSchema.Properties) > 0 || outSchema.HasAdditionalProperties || len(outSchema.UnionElements) > 0 {
+				// Disambiguate any Go field-name collisions before rendering, so
+				// the struct declaration and the generated marshalling
+				// boilerplate agree on the same names. See issue #2495.
+				ResolvePropertyGoFieldNameCollisions(&outSchema)
 				outSchema.GoType = GenStructFromSchema(outSchema)
 			}
 		}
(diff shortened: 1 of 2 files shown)


rifkir23:
thanks for the careful review, both points were valid. i reworked the fix so it is a single source of truth instead of only touching the struct declaration:

- collision resolution now runs once per schema (`ResolvePropertyGoFieldNameCollisions`) and stores the disambiguated name on the property. `GoFieldName()` returns it, so the struct declaration and the generated marshal/unmarshal boilerplate (union and additional-properties templates, which all read `GoFieldName`) now use the same name. this fixes the "boilerplate reads the wrong field" case.
- the resolver reserves the synthetic `AdditionalProperties` field name first, so a real property named `additionalProperties` (or one using `x-go-name: AdditionalProperties`) is suffixed instead of producing a duplicate declaration.

added a test for the additional-properties reservation case as well. full `pkg/codegen` suite and the rest of `./...` pass, gofmt and vet clean.

PR #2559, opened 2026-09-17
Generate with a toolchain whose flate output matches the fixtures

### LINKED ISSUE #2556: Committed swaggerSpec blobs depend on the Go toolchain; pin GOTOOLCHAIN for make generate
## Summary

`make generate` rewrites the embedded `swaggerSpec` block in most committed `.gen.go` files depending on which Go toolchain runs it, because the spec is deflate-compressed at generation time and `compress/flate` output changed in Go 1.27. The fixtures currently committed on `main` were produced by Go 1.27, while `go.mod` declares `go 1.25.0`. Anyone regenerating with the declared toolchain gets ~58 files of base64-only diffs and either commits noise or hand-reverts it.

## Evidence

Measured on `internal/test/servers/strict/stdhttp/server.gen.go` at `2679cec6` (current `main`), regenerating the fixture and hashing the `swaggerSpec` block:

| Toolchain | Blob matches committed file |
|---|---|
| go1.25.0 (the `go.mod` directive) | no |
| go1.25.9 | no |
| go1.26.0 | no |
| go1.26.2 | no |
| go1.27.0 | **yes, byte-identical** |

All 1.25.x and 1.26.x toolchains produce the same blob as each other. Decoding both blobs (raw deflate) yields identical spec content, so the difference is only the compressor.

The committed blobs date from `89284708` (2026-08-27, "chore(deps): update oapi-codegen/actions action to v0.9.2"), which touched 58 generated files with base64-only changes while `go.mod` already said `go 1.25.0`. That commit's regeneration ran on a 1.27 toolchain.

## Proposal

1. Regenerate everything once with the toolchain pinned to the `go.mod` directive and commit it as a standalone chore, so feature PRs stop carrying base64 churn:
   ```sh
   GOTOOLCHAIN=go1.25.0 make generate
   ```
2. Make the pin the convention: have the `generate` Makefile target set `GOTOOLCHAIN` from the `go` directive (or document it in `CONTRIBUTING.md`), so contributors on newer local toolchains reproduce the committed bytes.
3. Check whether anything in CI (the shared `oapi-codegen/actions` workflow or a renovate post-upgrade step) regenerates and commits fixtures with a floating toolchain. If so, pin it the same way, otherwise the blobs will flip back on the next automated regen.

## Notes

- Any PR opened before the chore lands will show the base64 churn if its author regenerated with 1.25/1.26. Reviewers can ignore hunks that are only `swaggerSpec` lines, but it's easy to miss a real change hidden among them.
- Alternative: make the embedded spec toolchain-independent (e.g. store it uncompressed, or compress with a fixed, vendored encoder). That's a larger change to `pkg/codegen/templates/inline.tmpl` and the runtime decode path, so the pin seems like the pragmatic first step.

mromaszewicz:
Fixes #2556.

Two commits: the Makefile change is 20 lines, the regeneration is 57 files of base64. Reviewing them separately is much easier than reviewing the squash.

### The problem

Generated code embeds the OpenAPI spec deflate-compressed, so its base64 depends on the toolchain's `compress/flate` — whose output changed in Go 1.27. Regenerating on a newer toolchain rewrites the blob in ~57 committed files, and a real change is easy to miss among that noise.

The blobs currently on `main` were produced by **Go 1.27**, which is newer than the `go` directive allows and newer than either version CI tests, so nobody on a supported toolchain could reproduce them.

### The fix

Every toolchain from the `go` directive through 1.26.x produces identical bytes:

| toolchain | blob |
|---|---|
| go1.25.0, go1.25.9, go1.26.0, go1.26.2 | `708092753472` (all identical) |
| go1.27.1 | `0c76b0f6ff16` |

So `generate` keeps using whatever Go is installed, and only steps down a toolchain at or past the change — to the version `go.mod` declares, derived from the `go` directive so there is nothing to keep in sync by hand.

Pinning unconditionally was the obvious move and is the wrong one: it would send everyone on a supported version to download an exact patch release they do not need, and a Makefile `export` overrides a `GOTOOLCHAIN=local` set by corporate policy, turning a policy setting into a build failure.

The second commit regenerates the fixtures to the range's bytes, so contributors on a supported Go reproduce them with their own toolchain — no pin, no download.

### Verification

On a real Go 1.27.1 install:

- `make generate` leaves all 58 files byte-identical.
- The same run under `GOTOOLCHAIN=local` produces `0c76b0f6ff16` — matching what is committed on `main` today, which independently confirms those blobs came from 1.27.
- Regenerating again restores the correct bytes, so the fallback is self-healing.

`make test` and `make lint` pass on 1.27.1 against fixtures generated at 1.25/1.26 — the decode path does not care which encoder wrote them. Every changed line in the second commit is a `swaggerSpec` base64 line; the decoded spec content is unchanged.

Also confirmed while investigating step 3 of the issue: **nothing automated regenerates.** No workflow runs `make generate`, and neither `renovate.json` nor the shared `oapi-codegen/renovate-config` has a `postUpgradeTasks` that does. The churn in `89284708` came from a maintainer regenerating by hand on a renovate branch. So no CI change is needed for this to hold.

### Notes

- There is a `TODO` on the fallback: it can be deleted once the `go` directive reaches 1.27, at which point every supported toolchain agrees again.
- This should land promptly — any open PR whose author regenerated on 1.27 will conflict across 57 files.
- It does not help downstream users, who get the same churn in their own generated code when they upgrade Go. Making the embedded spec encoding toolchain-independent would, but storing it uncompressed measures 5.0x larger (70 KiB → 353 KiB across these fixtures), so that deserves its own discussion rather than being folded in here.

---

<sub>🤖 Investigated and drafted with [Claude Code](https://claude.com/claude-code).</sub>


jamietanna:
What's the issue we're trying to fix here? We looked at this with https://github.com/oapi-codegen/oapi-codegen/issues/2536 and so use 1.27 in CI

If we're trying to make local changes easier - maybe we add a warning if we detect that we're not using Go 1.27?

mromaszewicz:
> What's the issue we're trying to fix here? We looked at this with #2536 and so use 1.27 in CI
> 
> If we're trying to make local changes easier - maybe we add a warning if we detect that we're not using Go 1.27?

Accidental generated file churn. I don't need this to go in, it's just an idea and I'm on the fence about it. I'm happy to just use 1.27 myself. I just noticed that some regenerated files had flate churn when I was going over PR's.

